Issue #105, opened 2017-02-09
Implement binary colvars trajectory file


jhenin:
Using 32-bit floats, would reduce the size by a factor of 5, and would be faster to read and process for large-scale analysis programs.

giacomofiorin:
I have already some code (not finished yet) that allows for binary I/O through overloading of the formatted I/O stream operators.  This would allow using the existing I/O functions.

Aside from that basic infrastructure, I see three requirements to satisfy this:
1) Handling incomplete writes or reads, essential particularly for trajectory files that get truncated when a simulation runs out of time.
2) Allowing the option to tune the precision: for example, if one tries to calculate the accumulated work of a bias in post-processing of a long trajectory, 64-bits may be useful.  The XTC routines of GROMACS (based on the XDR API) work fairly well and could be taken as example.
3) Allowing binary I/O through the scripting interface, so that a file format can be used by a linked shared object, similarly to what `gettimestep` or `rawtimestep` can be used for in VMD.  (Accordingly, I just closed #37, which has now become part of this.)  Do you object to start by changing the prototype `cv` Tcl proc to one that processes `Tcl_Obj` arguments instead of C strings?

giacomofiorin:
Never mind about XDR, I notice now that there is no established open-source Windows port.

akohlmey:
What about the XDR routines in LAMMPS or VMD? They work just fine on windows and have been in use for many years.
But you really don't have to go that far. You can basically do native binary output from C++ (or C) without a problem, when you follow two simple rules:
1. Have a fixed size, canonical header where you have a "magic number", a version tag, and flags for endianness as well as the size of all data types used.
2. Prefix each block of output with a small section containing a (string) label (the key) and the length (in bytes) and the number items and type of the following data (thus allowing for padded data blocks)

giacomofiorin:
Good point, I didn't check the LAMMPS version.  Anyway, I'll push soon what I have so far.

jhenin:
I see no problem at all with having the cv proc have a Tcl_obj as its argument.

jhenin:
No need to be reminded of this in the short term. This will happen eventually as a consequence of other improvements.

PR #111, opened 2017-03-13
Allow colvarscript to process object pointers

giacomofiorin:
This is partly to address potential implications of this commit (which implements part of #105), and partly to just test the Reviewer workflow of GitHub...

I changed the prototypes of the scripting API functions (and their calls from VMD and NAMD) so that they now handle pointers to byte arrays rather than C-style strings.  Immediate advantage: we don't need to const-cast the string argument `argv` any more when calling `colvarscript::run()`.

The `colvarscript` class does not return object pointers yet (only strings), but that could be allowed in the future (at least for a few uses, e.g. saving a pointer to a `colvar` object and calling `proc_colvar()` directly for that object).

Disadvantage: unless we begin tracking how the object arguments are converted to/from C strings (i.e., how the scripting API is called), this could introduce a dependency on the scripting language (at least for Tcl, where declarations of `Tcl_Obj` and related functions are needed).  Currently, I hid this dependency inside the usual "dumpster" class, `colvarproxy`: this is the simplest solution, but maybe not the most robust.

### DIFF
--- a/namd/src/ScriptTcl.C
+++ b/namd/src/ScriptTcl.C
@@ -1380,29 +1380,32 @@ int ScriptTcl::Tcl_colvarfreq(ClientData clientData,
 }
 
 int ScriptTcl::Tcl_colvars(ClientData clientData,
-        Tcl_Interp *interp, int argc, char *argv[]) {
-  ScriptTcl *script = (ScriptTcl *)clientData;
+                           Tcl_Interp *interp,
+                           int objc,
+                           Tcl_Obj *const objv[])
+{
+  ScriptTcl *script = (ScriptTcl *) clientData;
   script->initcheck();
   colvarmodule *colvars = Node::Object()->colvars;
   if ( ! colvars ) {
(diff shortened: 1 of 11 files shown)


jhenin:
I think eventually (that is, to properly integrate python) we'll need the kind of tracking you mention, that is, passing a generic pointer together with a flag that says what kind of binary object it refers to. That doesn't seem like a major drawback to me.

giacomofiorin:
Python, thanks to ctypes, doesn't need any additions to the C++ code to parse most binary objects (i.e., the example file in the `python` folder of this repo already implements a scripting API).  But to access externally-defined objects from the C++ code, their definition must be available at compile time, obviously.  It may be convenient to use `objv[0]`, the first element of the list of arguments to `run()`, to detect who's the caller and not change the prototypes in question again.

giacomofiorin:
I merged the PR in to avoid increasing divergence of the `proxy` branch. FYI, I also enabled protection against force-push into `master` to avoid accidents.

PR #122, opened 2017-06-26
Clean up the colvarproxy API

giacomofiorin:
The `colvarproxy` class has grown too big already some time ago.  However, its ease of use and maintainability have become even more important recently, due to the addition of new interfaces to scripting languages, functions that are only provided by a certain code (e.g. atom group COMs), and support for specific hacks of each program (e.g. `ofstream_namd`), etc.

This PR implements changes aimed at making more readable the API for the `colvarproxy` class by separating it into individual classes and moving most of its functions to a `.cpp` file. 

- [x] Split up colvarproxy members in multiple classes
- [x] Move all but inline functions from colvarproxy.h to colvarproxy.cpp
- [x] Remove unused methods (`select_closest_image()`, `select_closest_images()`, `exit()`).
- [ ] Replace all pure virtual functions with stubs (to be discussed)
- [ ] Document which functions define the minimal API (e.g. `colvarproxy_system` and `colvarproxy_atoms`) 

After this PR is merged, other changes that could be useful are:
- [ ] Merge Tcl interface code into `colvarproxy_script` (remove duplicated code between VMD and NAMD)
- [ ] Merge PDB reading code into `colvarproxy_atoms` (remove duplicated code between VMD and NAMD)
- [ ] Write a thin C-linkage API in `colvarproxy.h` to simplify calling its methods from other codes or with `ctypes`

### DIFF
--- a/doc/cv_version.tex
+++ b/doc/cv_version.tex
@@ -1,1 +1,1 @@
-\newcommand{\cvversion}{2017-06-23_proxy-streamline}
+\newcommand{\cvversion}{2017-06-26_proxy-streamline}
--- a/doc/cv_version.tex
+++ b/doc/cv_version.tex
@@ -1,1 +1,1 @@
-\newcommand{\cvversion}{2017-06-23}
+\newcommand{\cvversion}{2017-06-23_proxy-streamline}
--- a/doc/cv_version.tex
+++ b/doc/cv_version.tex
@@ -1,1 +1,1 @@
-\newcommand{\cvversion}{2017-06-26_proxy-streamline}
+\newcommand{\cvversion}{2017-06-26_proxy-cleanup}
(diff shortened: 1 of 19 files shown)


jhenin:
Looks good to me!

I think the following two questions you raised are linked:
> Replace all pure virtual functions with stubs (to be discussed)
> Document which functions define the minimal API

The current state of the code seems good to me in this respect, that is, it is consistent with the interfaces that we currently have. I can't comment on adding stubs in the abstract, but will be happy to do it with specific new interfaces in mind (OpenMM?).

giacomofiorin:
Sounds good, I'll finish checking that the change works for all implementations (it should) and merge into master.

I don't have a specific implementation in mind regarding the questions on whether functions need to be stubs or remain pure virtual.

As a matter of fact, for OpenMM specifically I'd like to use Python interface functions as much as possible, so that much of the work would be shared with @MDAnalysis. Since both codes keep the entire system in memory, there isn't much to do for requesting and collecting atomic data.

jhenin:
To build the other implementations you'll need the fix I just committed to master (forgot to make Lepton stuff optional).

giacomofiorin:
I know.  I had only executed the test suite in NAMD as usual.

PR #124, opened 2017-07-15
Umbrella integration-based estimator for eABF

fhh2626:
Dear Colvars developers:

in this code, I implemented umbrella integration-based estimator of eabf method. Like czar estimator, no additional option is needed for end users. The code outputs xxx.UI.grad, xxx.UI.count and xxx.UI.pmf(1D only) as classical abf does.

Restarting a simulation of for this estimator can only be achieved now using "inputprefix" but not "colvarsinput", because I didn't touch the format of the .colvar.state file.

I suggest to set a option for users to select estimators (naive, czar, UI-based), but this is the right for you colvars authors.

Cheers,
Haohao


fhh2626:
For multidimensional eABF simulation (n>=3), the UI-based estimator may be computitionally expensive because of high memory cost of this method.

giacomofiorin:
Hi Haohao, I'll let Jérôme coordinate with you on deciding a clear option for selecting different eABF estimators.  

I have a few comments on the code:
- [ ] The indentation is done with Tab characters.  While we are not always strict with that in Colvars, on this we try to be: please change the Tab characters to 2 or 4 spaces.
- [ ] Because of issues with Lustre, NAMD has stopped using `std::ofstream` and replaced it with their own class.  In Colvars, we use this class implicitly through the wrappers `cvm::proxy->output_stream(name, mode)` and `cvm::proxy->close_output_stream(name)`.
- [ ] There are several hard-coded constants, including BOLTZMANN: can you use `cvm::boltmann()` to initialize it?

Thank you!

fhh2626:
Hi Giacomo,

I revised the code as you suggested. As for documentation, I would like to do it after discussing with Jérôme about the new option.

Cheers,
Haohao

giacomofiorin:
Thanks Haohao, the documentation looks like the only remaining item to me.

If you have a small test that can run on deca-alanine, consider adding it following the template in the `tests` folder, where the tests for each code are compiled from.  This will ensure that future changes will preserve your functionality.

jhenin:
Thanks for this contribution, Haohao. I am not able to review it right away, but I will whenever possible.

src/colvar_UIestimator.h
```diff
@@ -0,0 +1,831 @@
...
+            const std::vector<double> & upperboundary,   // upperboundary of 
+            const std::vector<double> & width,           // width of x
+            const int y_size)           // size of y, for example, ysize=7, then when x=1, the distribution of y in [-2,4] is considered
+        {
+            this->lowerboundary = lowerboundary;
+            this->upperboundary = upperboundary;
+            this->width = width;
+            this->dimension = lowerboundary.size();
+            this->y_size = y_size;     // keep in mind the internal (spare) matrix is stored in diagonal form
+            this->y_total_size = int(pow(y_size, dimension) + 0.000001);
```
jhenin: These "magic small numbers" are not very clean. To get a rounded rather than truncated value, use 0.5 (see the colvargrid class).

src/colvar_UIestimator.h
```diff
@@ -0,0 +1,831 @@
...
+            {
+                matrix.push_back(std::vector<int>(y_total_size, 0));
+            }
+
+            temp.resize(dimension);
+        }
+
+        int inline get_value(const std::vector<double> & x, const std::vector<double> & y)
+        {
+            //if (matrix[convert_x(x)][convert_y(x, y)]!=0)
```
jhenin: These temporary statements can be removed before merging.

src/colvar_UIestimator.h
```diff
@@ -0,0 +1,831 @@
...
+                input_grad = n_vector<std::vector<double> >(lowerboundary, upperboundary, width, 1, std::vector<double>(dimension, 0.0));
+                input_count = n_vector<int>(lowerboundary, upperboundary, width, 1, 0);
+
+                // initialize input_Grad and input_count
+                std::vector<double> loop_flag(dimension, 0);
+                for (int i = 0; i < dimension; i++)
+                {
+                    loop_flag[i] = lowerboundary[i];
+                }
+                while (true)
```
jhenin: These loops look like they could be written without gotos, just replacing both `while(true) `with `while(i >= 0)` -- or am I missing something?

Same remarks for the similar loops below.

src/colvar_UIestimator.h
```diff
@@ -0,0 +1,831 @@
...
+            }
+        }
+
+        ~UIestimator() {}
+
+        // called from MD engine every step
+        bool update(const int step, std::vector<double> x, std::vector<double> y)
+        {
+
+            //std::cout<<"weeeee: :"<<std::endl;
```
jhenin: Same here, please cleanup temporary lines.

src/colvar_UIestimator.h
```diff
@@ -0,0 +1,831 @@
...
+        }
+    };
+
+    class UIestimator      // the implemension of UI estimator
+    {
+    public:
+        UIestimator() {}
+
+        //called when (re)start an eabf simulation
+        UIestimator(const std::vector<double> lowerboundary,
```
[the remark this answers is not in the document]
giacomofiorin: With C++ 11 this may not be an issue any more.

fhh2626:
Hi Jerome, Thanks for your suggestions. I've made corresponding revisions.

src/colvar_UIestimator.h
fhh2626: This is a trick to avoid something like int(4.999999999) == 4. I'm not sure if I should use 0.5 here.

giacomofiorin: @fhh2626 What about 4.999999998?  Since you haven't looked at the colvarscript class, here is the snippet for your convenience:
```
int nbins_round = (int)(nbins+0.5);
```
This is due to the integer cast working by truncation.

fhh2626: @giacomofiorin Thanks for your reply. In this sentence, a value of 0.5 also works because 
`pow(y_size, dimension)`
should be an integer.
But in
`loop_flag_x[i] > upperboundary[i] - width[i] + 0.00001`
I have to use something like 0.00001. Or should I define a constant like
`const double EPSILON = 0.00001`

src/colvar_UIestimator.h
```diff
@@ -0,0 +1,831 @@
...
+            if (restart == true)
+            {
+                input_grad = n_vector<std::vector<double> >(lowerboundary, upperboundary, width, 1, std::vector<double>(dimension, 0.0));
+                input_count = n_vector<int>(lowerboundary, upperboundary, width, 1, 0);
+
+                // initialize input_Grad and input_count
+                std::vector<double> loop_flag(dimension, 0);
+                for (int i = 0; i < dimension; i++)
+                {
+                    loop_flag[i] = lowerboundary[i];
```
jhenin: Is there a better name for variable loop_flag? It feels strange that a flag should be of type double. As it is, I have no idea what this variable does.

giacomofiorin:
Hi @jhenin and @fhh2626 can you agree on the documentation changes needed?

I feel that there are many issues still left, but they could also be addressed after merging into master.  As it is, this code doesn't seem to break any other features or generate new leaks.  On the other side, working with the open pull request will add more merge conflicts.

Alternatively, @fhh2626 can you enable push access for us into your repository? It may be possible to do it just as a flag for the pull request, i.e. the permissions expire once the PR is closed or merged.

giacomofiorin:
@jhenin Normally, it is the assignee that should merge the branch.

Who physically clicks the button is not a huge deal, but here one of the missing items is the documentation, which you and @fhh2626 should agree on.

Moreover, I was planning to squash several commits together, given your preference to work with a (mostly) linear history.  The current tree won't conform to it.  I am reverting the merge.

giacomofiorin:
I just moved Haohao's commits to `fhh2626-master`, and am giving him permissions to push there (as a code contributor, this is reasonable to have anyway). Please write the documentation, and when you are done I will squash-merge, while preserving the commit messages.

src/colvar_UIestimator.h
fhh2626: This "Magic Number" issue is fixed.

fhh2626:
@giacomofiorin @jhenin 
I addressed the "magic number 0.00001" issue, but I was greatly hampered by the Great Firewall these days. I cannot see if I successfully pushed the file even though a vpn. Could you help me to have a look? Thanks!

giacomofiorin:
Hi Haohao, I'm sorry to read that!

I see that you pushed to your repository, but that's OK.  I will
cherry-pick your commit and apply it to the new work branch.

Giacomo

giacomofiorin:
I just merged all changes into the master branch, via a single squashed commit

fhh2626:
Thanks a lot for your help, Giacomo

PR #157, opened 2018-05-10
Pairlists for coordnum

jvermaas:
In using coordnum quite a bit, the performance has always left something to be desired if I go overboard in terms of the number of atoms involved, even if only a few are interacting. One of the ways that PLUMED handles this is to create a pairlist at a user-specified point occasionally, reducing the performance impact of computing a large number of pairs. I've put together a prototype (and completely untested!) implementation of this same sort of idea, which would maintain a pairlist occasionally. However, before I can test this, there are some practical issues that I don't know the answer to:

1.) Do you have recommendations for how to easily test this? Compiling a full VMD/NAMD with this patch applied gets tedious really quick!

2.) I think in tracing the function calls that "calc_value" for individual collective variables is computed at every frame, which is why I put the pairlist computation there. Is this a safe assumption generally?

3.) I see that there is a "step_relative()" function that basically does what I want to get the number of steps since the simulation started. Are there any problems using this instead of maintaining a separate count?


giacomofiorin:
Hi @jvermaas, thanks a lot for taking a stab at it!  @fabsugar and I had discussed the issue recently, but we have both been focused on other things, and neither did anything.

Regarding your questions:
1. The best way to test is reproducing the output of the `coordNum` and `selfCoordNum` regression tests: if you choose a fair tolerance criterion, there should be no difference within machine precision.  You could just compile NAMD with this patch using the update script at the top level dir, and run the test in `namd/tests/library/000_coordnum_harmonic-fixed` (after adapting the input to use the pairlist).  I see no need to test it in other codes, because the functionality is all internal to the library: the PBC distance functions are used the same way with or without pairlist.  (*Note:* if you choose to add a new regression test, put the Colvars config file in the `tests` folder under the top dir, and add it to the `build_tests.sh` script to create the NAMD inputs.)
2. I fully agree with your choice of putting the pairlist build code in `calc_value()`, which is always called first. But I would wrap this code in a function of the `colvar::cvc` base class, which takes pointers to the two groups (or to the same group, in the case of `selfCoordNum`).  Then you can just call it from within `calc_value()` for any derived class.  The problem may be translating the `coordNum` tolerance to a distance cutoff (because the base class is not aware of the `coordNum` specifics), but this conversion can be done inside `coordNum`.
3. Yes, please use `step_relative()`, as it is incremented consistently with the MD engine; if you increment your own counter, it will be out of sync with NAMD or LAMMPS when you use "run 0".

Lastly, I suggest removing some duplicated code by passing the pairlist flag as an argument of `switching_function()`, so that it returns immediately if the flag is "false". Since the function is inlined, there is no overhead associated to calling it every time and let it do the check.

Again, thanks a lot for starting the ball on this, and let me know if there is any of the above that I can help with.

jvermaas:
Thanks @giacomofiorin. The reason I originally thought about using a tolerance instead of a distance cutoff is because anisotropic rvecs are allowed. This means that there isn't one specific cutoff *distance* that is actually always relevant, and so in my mind it seems easier to implement by tying it closely within coordination number rather than calling a new function in cvc-land that needs to know about distances. Is there another colvar that needs a pairlist?

giacomofiorin:
Hi @jvermaas that's a fair point. I can see a couple of anisotropic use cases, even though the most common is the isotropic one by far. So for, now, let's start with `coordNum`: can you transfer the pairlist generation to its own function in the `coordnum` class?

I can look at porting it to the base class later: aside from `selfCoordNum` and in general the variants of `coordNum`, `distanceInv` and `distancePairs` also have O(N^2) cost.

jvermaas:
It is definitely possible to do so, it is just less efficient. Each of the individual components that I would want to compare with already is evaluated within the loop. I'd be looping a second time if I split the pairlist computation apart from the value computation. We'll see. `distancePairs` probably can't be pairlisted, since the value at every step depends on knowing **ALL** of the pairs. For `distanceInv`, the natural use case is one with small groups of atoms within a single colvar, so your N^2 doesn't blow up to the huge proportion it could for tumbling and interacting polymers. Basically my plan is to test what I've got now after cleaning things up and go from there.

jvermaas:
Tests added and docs updated. I think its ready for you now @giacomofiorin.

giacomofiorin:
@jvermaas Thank you for taking the time to make these changes, test and update the doc. A good portion of this code is almost 13 years old and needed a refresh.

To avoid checking in the inner loop for the presence of a pairlist and whether it should be updated, I moved the flags `use_pairlist` and `rebuild_pairlist` to template arguments/flags. The commit is 735d678b3ef4eff36d9722629f0c7c5b3e39feec. Can you check whether it works in the intended case? The regtests currently included all pass.

One advantage of moving the pairlist build/update to the template is being able to reuse it in other variables: `distanceInv` is not just for NOE restraints, and `distancePairs` is more of a primitive for a user-defined function than a function itself. With a suitable pairlist, even a Tcl-scripted function may be not too bad.

Note that I rebased your changes on top of `master` to avoid the large merge commit in the history, and plan on closing this PR: is that OK for you? For the future, I'm adding you to the list of people with write access so that next time you can fork a branch from `master` and push there.

@jhenin A natural extension of the new template scheme would be emptying `calc_gradients()` and moving its contents to `calc_value()`, making the corresponding calls conditional upon the state of the `f_cvc_gradient` feature. Any objections? This might save an additional factor of 2. 

On a general note, we may not be too far from having a single compute function for the CVCs, controlled by the feature flags.

jvermaas:
Yeah, that's fine. I'll try not to make too big of a mess in the git commits in the future. :) I'm also all in on combining the function and gradient evaluation. Most MD packages combine force and energy evaluation in the same way, so it is a fairly natural extension.

jhenin:
This is all looking great. I'm wondering one thing though - instead of building a const int flags and replicating the templated function call depending on flag values, wouldn't it be simpler to define each flag as a bool template parameter?

I completely agree about merging calc_value and calc_gradient. That would just be formalizing a connection that existing code has relied on forever.

giacomofiorin:
> This is all looking great. I'm wondering one thing though - instead of building a const int flags and replicating the templated function call depending on flag values, wouldn't it be simpler to define each flag as a bool template parameter?

I was tempted to go with that for simplicity, but each new flag added later would need to be added to the each template instantiation. Worse yet, templates need to be instantiated with compile-time constants, and each time the number of template parameters increases to **N** the number of instantiations goes up to almost **2^N**. It seems unmaintainable to me.

> I completely agree about merging calc_value and calc_gradient. That would just be formalizing a connection that existing code has relied on forever.

That really depends on the type of variable. `coordNum` and other atom- or pairwise-additive do resemble the core loops of a MD force field; other variables instead require a separate loop for the gradients. In addition there are the total force-related quantities, which we did not have runtime flags to check for until relatively recently (thanks for taking care of that).

jhenin:
> I was tempted to go with that for simplicity, but later each new flag would need to be added to each template instantiation.
I see what you mean, it's a change of interface. A small one though, and it wouldn't happen often. And conversely, adding a flag to the current version means adding a layer of nested ifs to the callers, where applicable. To me using bool template parameters is letting the compiler write the nested ifs.

> Worse yet, templates need to be instantiated with compile-time constants, and each time the number of template parameters increases to N, the number of instantiations goes up to almost 2^N. It seems unmaintainable to me.
For sure, it's only efficient if the number of flags is kept low - 3 seems ok. I don't foresee a huge number of new flags for switching_function, but maybe there are many possibilities that I don't have in mind right now.

giacomofiorin:
This PR is effectively merged by 325e1b3 (rebased onto `master`).

jhenin:
If a bool parameter is known at run-time, I'd expect the compiler to instantiate both cases and insert a conditional jump in the code... what else can it do?

giacomofiorin:
Can you test it?  All the doc I know of says that template parameters must be constant expressions:
http://en.cppreference.com/w/cpp/language/template_parameters#Template_non-type_arguments

It'd be reasonable to expect the compiler to expand all combinations of bool parameters, because they would be only 2^N, but for template parameters of any other type the size of the code would bloat.

jhenin:
Eh, you're right, compilers only do that in my dreams, not in reality. GCC rejects a non-const bool.

PR #164, opened 2018-07-25
Check for matching braces before beginning to parse

giacomofiorin:
Should take care of complicated errors due to unmatching braces.

Also a good opportunity to test the merge workflow.


giacomofiorin:
Just wanted a name for the function that is a verb, and to distinguish it from the previous version since I changed its return type (we started introducing the int return code after porting Colvars to VMD, but that function is circa 2005...).

PR #167, opened 2018-07-30
Jerome misc

jhenin:
Doc & regtest updates


doc/colvars-refman-main.tex
```diff
@@ -2688,14 +2688,20 @@ \subsection{LAMMPS keywords}
...
+The following keywords can be used within periodic components, and within
+custom \cvscriptonly{or scripted} colvars (\ref{sec:colvar_custom_function}
+\cvscriptonly{, \ref{sec::colvar_scripted}}).
 
 \begin{itemize}
 \item %
   \keydef
     {period}{%
-    \texttt{distanceZ}}{%
+    \texttt{distanceZ}, custom colvars}{%
```
giacomofiorin: The keyword period is currently parsed regardless of type: should we disallow it for cases that are not distanceZ or custom/scripted variables?

jhenin: Well, internally they are separate keywords, one at the cvc level and the other at the colvar level. The one at the colvar level is only parsed for scripted/custom colvars.

At the cvc-level, sure we could check the CVC type. Is that known to cvc::init()? This seems like it could be more trouble that it's worth... at worst, those parameters are ignored.

giacomofiorin: Unfortunately not yet, because most derived CVCs still do all initialization in their constructor, which is called after the base class constructor. In the future, they will have their own `init()` functions, and `function_type` (which is set by the derived constructor) at that point should be visible to the base-class `init()`. This is not a huge change, but definitely outside the scope of this PR. Will move ahead.

giacomofiorin:
The PR is merged, GitHub is suggesting to delete the branch `jerome-misc`, which I'm gonna follow since the commits in it have now been rebased. You can create a new branch with the same name at any time, though.

PR #174, opened 2018-09-07
Warn only, but do not try to fix zero-mass problems

giacomofiorin:
The release of NAMD 2.13b1 is being announced in advance to the external developers, via the usual formal notice sent by UIUC:
LMFO

Given the time constraints and the likely impossibility to introduce an error condition for zero-mass atoms, I propose that we warn users, but do not introduce inconsistencies with the NAMD COM code until we have offered a mechanism to control the treatment of zero-masses via the scripting interface.


jhenin:
Agreed.

PR #182, opened 2018-10-09
Linear bias vector

### LINKED ISSUE #181: LAMMPS bias vector
Hello, we are working on developing our own vector collective variable component, which we would use to bias simulations. We would hope to use a vector for speed. As a test, we were looking at the distancePairs collective variable. Using LAMMPS, we were hoping to use the linear bias function to bias a distance pair. Yet, we get the following error:
Trying to assign a colvar value with type "scalar number" to one with type "n-dimensional vector".

Here is our input:
colvarsTrajFrequency    200
colvarsRestartFrequency 200
indexFile               groups.ndx

colvar {
name    pr1
distancePairs {
group1  { atomNumbers 1 2 3 }
group2  { atomNumbers 20 21 22 }
    }
}

linear {
name              max_ent
colvars           pr1
forceConstant    1
#centers       0 0 0 0 0 0 0 0 0
centers  (0, 0, 0, 0, 0, 0, 0, 0, 0)
outputEnergy      no
}

giacomofiorin:
Allow linear bias to deal with non-scalar variables.

@alatham13 I could not add you directly as reviewer as well, but I would like to.  If you accept the invitation to join the "Reviewers" team, we should be able to do it.

Fixes #181


giacomofiorin:
Assuming this is good from e-mail feedback.

PR #185, opened 2018-10-29
Documentation improvements

giacomofiorin:
Includes better explanations of `width` and `hillWidth`, using wording that will be compatible with future keywords.  Also enables the use of hyperlinks in the NAMD UG chapter.


jhenin:
+1

PR #189, opened 2018-11-02
Minimal changes to limit width of HTML doc page

giacomofiorin:
Not urgent, but needs a look.  I pushed already the accompanying changes in the website repo.


giacomofiorin:
FYI I pushed the output of this change to the webpage repository, see e.g. https://colvars.github.io/colvars-refman-vmd/colvars-refman-vmd.html.  If it works, I'm inclined to just merge this in.

jhenin:
Not sure what guidelines to follow, but that seems a little on the narrow side to me...

Side note, the graphic in fig 1 is missing.

giacomofiorin:
@jhenin I added a simple conditional style for screen size, which makes it look OK on my phone.

As for the high-res screens showing the column to narrow, a hotkey like `Ctrl-+` or similar is supported by nearly all browsers.

jhenin:
New layout looks good to me.

PR #196, opened 2018-12-04
Parser improvements

### LINKED ISSUE #68: Parser improvements
There is a need for a new mode `parse_required`: the value of the keyword must be provided, and the default value should never be used (some initializers, and all restart reads). The new mode should remove the need for the numerous calls of the type:
`cvm::error("Error: xxyyzz is missing from the restart.\n")`
Values for new entries in the Parse_Mode enum can be given as bitwise expressions, allowing bitwise ORs in calls to `get_keyval()`.

Also, by default `get_keyval()` should record the first time it has been called within that object, and avoid overriding the current value of a variable with the default value (allow multiple iterations of `init()`).

giacomofiorin:
Closes #68 

This PR makes `Parse_Mode` a set of feature flags using bitwise operators.  A small number of hard-coded tests (e.g. missing parameters from the restart of restraint biases) make use of new flags `parse_required` and `parse_restart`.

Much duplicated code has also been removed from the `colvarparse` class.  Conversely, `cvm::to_str()` is now a set of regular overloaded functions, instantiating the original template: this is a cherry-pick of 49f2873 (`compiler_fixes` branch).  This turned out to be more important to incorporate when using the method in the parsing functions.

The `parse_quiet` flag that I mentioned in #68 seemed a bit too simplified in the end, and I opted for `logLevel` instead.  The keyword is so far *not* documented due to its limited support (see 12056cc), and should be fully implemented in another PR.

The same applies to integration with the dependency handler discussed in #68.

### DIFF
--- a/src/colvarmodule.cpp
+++ b/src/colvarmodule.cpp
@@ -1943,105 +1943,106 @@ template<typename T> std::string _to_str_vector(std::vector<T> const &x,
 }
 
 
-std::string cvm::to_str(int const &x,
-                        size_t width, size_t prec)
+
+std::string colvarmodule::to_str(int const &x,
+                                 size_t width, size_t prec)
 {
   return _to_str<int>(x, width, prec);
 }
 
-std::string cvm::to_str(size_t const &x,
-                        size_t width, size_t prec)
+std::string colvarmodule::to_str(size_t const &x,
+                                 size_t width, size_t prec)
 {
   return _to_str<size_t>(x, width, prec);
 }
 
-std::string cvm::to_str(long int const &x,
-                        size_t width, size_t prec)
+std::string colvarmodule::to_str(long int const &x,
+                                 size_t width, size_t prec)
 {
   return _to_str<long int>(x, width, prec);
 }
 
-std::string cvm::to_str(cvm::real const &x,
-                        size_t width, size_t prec)
+std::string colvarmodule::to_str(cvm::real const &x,
+                                 size_t width, size_t prec)
 {
   return _to_str<cvm::real>(x, width, prec);
 }
 
-std::string cvm::to_str(cvm::rvector const &x,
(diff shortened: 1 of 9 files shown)


src/colvarmodule.cpp
```diff
@@ -254,6 +254,7 @@ int colvarmodule::parse_global_params(std::string const &conf)
     std::string index_file_name;
     size_t pos = 0;
     while (parse->key_lookup(conf, "indexFile", &index_file_name, &pos)) {
+      cvm::log("# indexFile = \""+index_file_name+"\"\n");
```
jhenin: This feels like it could be made more natural with the improved parser - that is, allowing repeated calls to get_keyval ?

src/colvarmodule.cpp
```diff
@@ -1943,105 +1943,106 @@ template<typename T> std::string _to_str_vector(std::vector<T> const &x,
 }
 
 
-std::string cvm::to_str(int const &x,
-                        size_t width, size_t prec)
+
+std::string colvarmodule::to_str(int const &x,
```
jhenin: The difference between cvm:: and colvarmodule:: is unclear to me. In this commit a number of instances are converted to the second version, but in other files many cases of cvm:: were added. What is at stake here, besides brevity?

src/colvarmodule.cpp
giacomofiorin: Definitely yes.  I plan on implementing that separately.

src/colvarmodule.cpp
giacomofiorin: Generally, I preferred to use the full `colvarmodule::` for the implementation of the class, and `cvm::` for the calls to its functions.  The difference is 100% cosmetic right now, but it would simplify moving to the use of a namespace if we need or want to in the future.

PR #41, opened 2015-05-05
Dipole angle  implementation

alejob:
New collective variable added, dipole_angle. This new implementation is similar to colvar "angle", but "calc_gradient" and vector "r21" have been modified. A new function "total_charge", similar to "total_mass", has been implemented. Also, a new function for atom group, "dipole", has been implemented. The parser for LAMMPS and NAMD has been modified to extract charges to be used in dipole calculation. The colvars-ref-manual has been modified.

Issue #79, opened 2016-09-28
Combine "colvars" and "colvar components" chapters 


giacomofiorin:
With the introduction of `scriptedFunction`, which blurs the distinction between colvars and their components, and the (soon to be done) move of walls, correlation functions and running averages into biasing objects, there won't be a need any more to separate the documentation of colvars and CVCs.

The combined chapter should be positioned before the one on atom groups definition.

This new arrangement should benefit readers who have so far been discouraged by the complexity of colvars with more than one component (i.e. d1-d2).

PR #198, opened 2018-12-12
Added dipoleMagnitude colvar. Docs updated also.

alejob:
Hi, I've implemented the dipole magnitude collective variable. It is similar to dipole angle collective variable, details in #41 

Please let me know if you need anything else.


alejob:
I must solve a conflict in colvars-refman-main.tex

giacomofiorin:
@alejob Thanks for including this new type of variable.  Unfortunately the version that you forked from precedes a reorganization of the reference manual (see https://github.com/Colvars/colvars/issues/79) and made the merge more difficult.

One question: in the doc, the variable `dipoleMagnitude` is listed by you among those with non-scalar values, but the function implemented returns a scalar, and the name suggests that you are using `$|\vec{d}|$` as a variable, not `$\vec{d}$` (3-dimensional vector).  Can you confirm this?

alejob:
You're right, Giacomo, `dipoleMagnitude` should be in the scalar section. I'll update the docs to fix this mistake.

alejob:
I think that now everything is ok.

src/colvarcomp_distances.cpp
```diff
@@ -740,6 +740,63 @@ void colvar::distance_pairs::apply_force(colvarvalue const &force)
...
+    ai->grad = (ai->charge - aux1*ai->mass) * dipVunit;
+  }
+}
+
+void colvar::dipole_magnitude::apply_force (colvarvalue const &force)
+{
+  if (!atoms->noforce) {
+    atoms->apply_colvar_force (force.real_value);
+    for (cvm::atom_iter ai = atoms->begin(); ai != atoms->end(); ai++) {
+      cvm::atom_pos forceV=force.real_value*ai->grad;
```
giacomofiorin: @alejob Can you clarify the purpose of this line?

alejob: The lines 789 and 790 were to debugging, they built a force vector that was used to check if the total sum over the molecule was zero (don't move the center of mass). I'll check it again, and I'll do a new commit removing this lines. Thanks for the review.

giacomofiorin: No worries, I just wanted to make sure that the gradients are implemented as you intend them.

Issue #184, opened 2018-10-12
Reuse fittingGroup (requires handling of multiple sets of fit gradients)


joeyelk:
If fittingGroup could be used without refPositionsFile to reference another group of atoms in the simulation, this would allow, for example, constraining the relative tilt of two groups.

jhenin:
Hi @joeyelk,

that's something we've had in mind for a while. There are two problems with that - one is that the fitting group as implemented now stores gradients that are specific of a colvar component. That could be changed, however there is a second, more complex issue: having one group use a previously defined fitting group introduces a dependency between two atom groups. That dependency breaks the logic we currently have for parallel processing of different variables and components in SMP architectures.

By the way, the bias you are describing sounds like it can be achieved already. You can either restrain the tilt of one group with respect to the other, or restrain the difference between two tilt components.

joeyelk:
Hi @jhenin and @giacomofiorin,
Thank you for your response and suggestions.  The system I am considering is two identical groups A and B.  I would like to have collective variables that are the equivalent of the spinAngle and tilt of B referenced to A.  
Currently, I use the same reference structure for both spinAngle and tilt calculations and omit the axis keyword to use the default z axis.   
 I calculate cos (tilt angle) of A and B and restrain both to 1.0.  This allows for rotation only about the z axis, which is measured by spinAngle.  If I instead constrain the difference in cos (tilt angle) of the two groups to be zero, this would permit, for example, one molecule to rotate 90 degrees about the x axis, and the other to rotate 90 degrees about the y axis.  
The spinAngles of A and B are then subtracted and then remapped between -180 and 180 using a custom function.  I ultimately want to do metadynamics and umbrella sampling calculations with the relative spinAngle.  Does this require flagging this custom function as periodic, and if so is there a way to do this?  
As an alternative to spinAngle, I could divide A and B into subgroups and measure a dihedral.  
Best,
Joe

giacomofiorin:
Hi @joeyelk, apologies if I'm missing something.  Is it possible for you to use A as `fittingGroup` for B, and apply both the restraint on `tilt` and the metadynamics or harmonic bias in the rotated frame.  Any forces applied on B in the rotated frame are also projected onto A, for as long as `enableFitGradients` is on (default).

giacomofiorin:
[...]

PR #199, opened 2018-12-14
Allow dependency resolution when base's and derivative's features conflict

giacomofiorin:
This is done by overriding conflicting features in a base CVC's constructor in the derivative CVC's constructor, but calling `init_cvc_features()` only after the constructor chain is invoked.

Fixes the problem observed by @joeyelk and discussed in the comment of issue #184.


giacomofiorin:
I accidentally submitted the PR while editing the description, but it's updated now.

PR #203, opened 2018-12-21
Compiler fixes

giacomofiorin:
Massive set of small changes to appease compilers on a variety of platforms.

### DIFF
--- a/src/replace_stl_math.sh
+++ b/src/replace_stl_math.sh
@@ -1,0 +1,14 @@
+for f in *.{cpp,h} ; do
+
+    ## The following is not the full list under std::, but should catch all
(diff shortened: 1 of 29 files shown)


giacomofiorin:
Compiles (after @jhenin fixed my merge error) and passes tests.

src/replace_stl_math.sh
```diff
@@ -0,0 +1,14 @@
+for f in *.{cpp,h} ; do
```
jhenin: Did you mean to add this file?

giacomofiorin: Yes, to catch STL instances in branches unmerged so far.  We can delete it in the future.

PR #204, opened 2018-12-21
If regtests fail, try lower precision

jhenin:
Upon failure of a regtest result, try comparing outputs using a given range of lower precisions iteratively.
Makes it easier to run a demanding regtest and see at a glance if the failures are due to large or small divergences.

### DIFF
--- a/lammps/tests/library/run_tests.sh
+++ b/lammps/tests/library/run_tests.sh
@@ -21,13 +21,6 @@ if ! { echo ${DIRLIST} | grep -q 0 ; } then
   DIRLIST=`eval ls -d [0-9][0-9][0-9]_*`
 fi
 
-# Precision requested to pass (negative powers of ten)
-DIFF_PREC=6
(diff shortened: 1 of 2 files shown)


giacomofiorin:
Have to dust up the LAMMPS updater script, I'll try this there asap.

jhenin:
For now I kept the values that were in place for each code. If we set the passing precision to the same for both, I would favor going to 1e-6, because from experience that is the threshold for NAMD-CPU that will catch bugs, but not meaningless variations. At 1e-7 I'm afraid I'll never see the "All tests suceeded" message ever again.

giacomofiorin:
Fair enough, I forgot you changed the threshold for LAMMPS tests.

giacomofiorin:
I just switched both to 1.0e-6.  I do observe all passes with NAMD and GCC 4.8, but once they are done with the integrator refactoring that'll go as well.

Issue #168, opened 2018-08-10
Enhancements to scripting interface


jhenin:
These would be specifically useful for the VMD GUI "Colvars dashboard"

- [x] get config strings for colvars including comments (save before stripping)
- [x] get config string for the whole module
- [x] access atom groups from colvars->cvcs
- [x] clear index groups to allow reloading file

More generally, some script access to config parsing functions?

giacomofiorin:
The PR #172 contains solutions to items 2 and 4.

For item 1, I'm not sure how to go about it without changes to the lower-level parser: because comment characters are taken into account when finding opening and closing braces, we would need to change how braces are handled.

For item 3, can you specify what you had in mind?  A list of atom IDs seems reasonable, but let's make sure first that that's sufficient.

jhenin:
PR #172  looks great already!

I see the problem for item 1... From the perspective of the GUI, the alternative would be to remember the config strings that are passed to Colvars for parsing, but this runs the risk of getting out of sync if the config is changed independently from the GUI. The nice thing about querying the config in real time is that it's very robust to external changes.

For item 3, a list of lists of atom IDs would be perfect. Alternately, a way to query each group individually avoiding nested lists makes things simpler...

jhenin:
(in-person conversation GF & JH) Clarifying the requirement for item 1: that is assuming the object is created based on a single config string. Once it can be modified by parsing partial config strings, one cannot rely on such a string to recreate it. Usage by the Dashboard will be limited to the classic workflow: 1 object = 1 monolithic config string.

giacomofiorin:
The main issue stopping from implementing the remaining item (retaining comments in the config strings of individual object, as done for the whole module already) is that the parser assumes that comments are stripped before any of the following:
- parsing of global keywords;
- counting braces
- check for unknown/misspelled keywords.

This behavior can be modified, but it would take some time and require changes to the lower-level parsing functions.

Can the equivalent be achieved just by keeping the config string in Tcl?  This would work if the main use of this feature is within the Dashboard.  I understand that this is difficult to do when importing an existing configuration file.  But it could be made a lot simpler by incorporating tags as comments when a config file is generated in the Dashboard:
```
#CV-Dashboard
colvarsTrajFrequency 1000

#CV-Dashboard
colvar {
  ...
}

#CV-Dashboard
harmonic {
  ...
}
```

[...]

PR #220, opened 2019-02-15
Dashboard option to display atoms involved in cvs

jhenin:
Partially answers #168 (item 3).

### DIFF
--- a/src/colvaratoms.cpp
+++ b/src/colvaratoms.cpp
@@ -808,24 +808,24 @@ int cvm::atom_group::create_sorted_ids()
 
   // Sort the internal IDs
   std::list<int> sorted_atoms_ids_list;
-  for (size_t i = 0; i < this->size(); i++) {
+  for (size_t i = 0; i < atoms_ids.size(); i++) {
     sorted_atoms_ids_list.push_back(atoms_ids[i]);
   }
   sorted_atoms_ids_list.sort();
   sorted_atoms_ids_list.unique();
-  if (sorted_atoms_ids_list.size() != this->size()) {
+  if (sorted_atoms_ids_list.size() != atoms_ids.size()) {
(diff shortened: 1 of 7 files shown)


src/colvarscript.cpp
```diff
@@ -324,6 +324,22 @@ int colvarscript::proc_colvar(colvar *cv, int objc, unsigned char *const objv[])
     return COLVARS_OK;
   }
 
+  if (subcmd == "getatoms") {
```
jhenin: I am using the old-style scripting interface here, mostly because of the list-of-lists structure of the data, for which I have no good generic solution. I'm happy to reimplement this in the generic form when we have a converged way to handle the data structures.

jhenin:
With this latest update, I think the GUI side is complete. The changes at the C++ level seem minimally invasive to me, but they are the reason why I submitted this as a PR. See also comment https://github.com/Colvars/colvars/pull/220/files#r258517671 on the scripting interface.

src/colvarscript.cpp
giacomofiorin: That's reasonable for now.

src/colvaratoms.cpp
```diff
@@ -808,24 +808,24 @@ int cvm::atom_group::create_sorted_ids()
 
   // Sort the internal IDs
   std::list<int> sorted_atoms_ids_list;
-  for (size_t i = 0; i < this->size(); i++) {
+  for (size_t i = 0; i < atoms_ids.size(); i++) {
```
jhenin: Note: these change are necessary to obtain atom IDs when groups are marked as scalable.

src/colvarscript.cpp
giacomofiorin: P.S. Can you change to `getatomids`?

PR #222, opened 2019-02-20
Obtain ABF gradient and free energy through Tcl interface

jeffcomer:
I've added the ability to get the ABF gradient and free energy through the Tcl interface. This is useful for replica exchange ABF. I've tested it to make sure that "cv bias abf1 freeenergy $x" gives the correct effective applied energy (including the influence of fullSamples). As of now, it works only for 1-dimensional ABF. It may or may not work for eABF. I plan on looking into that.

### DIFF
--- a/src/colvarbias.cpp
+++ b/src/colvarbias.cpp
@@ -69,7 +69,6 @@ int colvarbias::init(std::string const &conf)
       cvm::error("Error: no collective variables specified.\n", INPUT_ERROR);
       return INPUT_ERROR;
     }
-
   } else {
     cvm::log("Reinitializing bias \""+name+"\".\n");
   }
@@ -266,6 +265,16 @@ int colvarbias::current_bin()
 int colvarbias::bin_count(int bin_index)
 {
   cvm::error("Error: bin_count() not implemented.\n");
+  return COLVARS_NOT_IMPLEMENTED;
+}
+cvm::real colvarbias::bin_gradient(int bin_index)
+{
+  cvm::error("Error: bin_gradient() not implemented.\n");
+  return COLVARS_NOT_IMPLEMENTED;
+}
+cvm::real colvarbias::delta_free_energy(cvm::real x)
+{
+  cvm::error("Error: delta_free_energy() not implemented.\n");
   return COLVARS_NOT_IMPLEMENTED;
 }
 int colvarbias::replica_share()
--- a/src/colvarscript.cpp
+++ b/src/colvarscript.cpp
@@ -450,6 +450,7 @@ int colvarscript::proc_bias(colvarbias *b, int objc, unsigned char *const objv[]
     return COLVARS_OK;
   }
 
+  
   if (subcmd == "share") {
     int r = b->replica_share();
     if (r < 0) {
@@ -480,10 +481,33 @@ int colvarscript::proc_bias(colvarbias *b, int objc, unsigned char *const objv[]
     if (subcmd == "count") {
       int index;
       if (!(std::istringstream(param) >> index)) {
-        result = "bin_count: error parsing bin index";
+        result = "bias count: error parsing bin index";
         return COLVARSCRIPT_ERROR;
       }
       result = cvm::to_str(b->bin_count(index));
+      return COLVARS_OK;
+    }
+
+    // Get ABF gradient
+    if (subcmd == "gradient") {
+      int index;
+      if (!(std::istringstream(param) >> index)) {
+	result = "bias gradient: error parsing bin index";
+	return COLVARSCRIPT_ERROR;
+      }
+      result = cvm::to_str(-b->bin_gradient(index));
+      return COLVARS_OK;
+    }
+
+    // Get free energy by integrating over gradients
+    if (subcmd == "freeenergy") {
+      cvm::real x;
+      if (!(std::istringstream(param) >> x)) {
+	result = "bias freeenergy: error parsing colvar value";
+	return COLVARSCRIPT_ERROR;
+      }
+      cvm::real dg = b->delta_free_energy(x);
+      result = cvm::to_str(dg);
       return COLVARS_OK;
     }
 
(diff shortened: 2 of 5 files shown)


giacomofiorin:
Hi @jeffcomer, a couple of suggstions:
- Can you use the existing `energy` command to return the bias's energy, i.e. the negative of what you return as `freeenergy`?  This would allow replica exchange scripts to be more uniform between biases.
- If only 1D ABF is tested, I strongly recommend making this feature incompatible with multi-dimensional ABF.

src/colvarbias.cpp
```diff
@@ -268,6 +267,16 @@ int colvarbias::bin_count(int bin_index)
   cvm::error("Error: bin_count() not implemented.\n");
   return COLVARS_NOT_IMPLEMENTED;
 }
+cvm::real colvarbias::bin_gradient(int bin_index)
+{
+  cvm::error("Error: bin_gradient() not implemented.\n");
+  return COLVARS_NOT_IMPLEMENTED;
+}
+cvm::real colvarbias::delta_free_energy(cvm::real x)
```
giacomofiorin: These two functions are by design incompatible with multiple variables, or non scalar variables.  Rather than overloading functions later, which will be confusing, I'd recommend using `colvarvalue` instead of `cvm::real` and `std::vector<int>` instead of just `int`.

jhenin:
In 2D and 3D ABF, you could rely on the existing integrated gradients, especially if they are only necessary at sparse times. The fullSamples question is less pressing then, because noisy gradient estimates get smoothed out by the Poisson integration process.

src/colvarscript.cpp
```diff
@@ -480,13 +481,36 @@ int colvarscript::proc_bias(colvarbias *b, int objc, unsigned char *const objv[]
...
-        result = "bin_count: error parsing bin index";
+        result = "bias count: error parsing bin index";
         return COLVARSCRIPT_ERROR;
       }
       result = cvm::to_str(b->bin_count(index));
       return COLVARS_OK;
     }
 
+    // Get ABF gradient
+    if (subcmd == "gradient") {
```
giacomofiorin: This command should have a more specific name, e.g. `bin_gradient`, and be the gradient of the bias energy, so perhaps `bin_forces` would be even more appropriate.

More generally, it seems reasonable to me to have `energy` take as optional arguments  the values of collective variables (other than the current).  `bin_energy` could then be the corresponding function that takes bin indices instead of real values.  Likewise, `forces` and `bin_forces` would do the same in gradient space.

You should not provide a complete implementation of the above, but at least use command names that do not clash with potential use in restraints or metadynamics biases.

src/colvarscript.cpp
```diff
@@ -480,13 +481,36 @@ int colvarscript::proc_bias(colvarbias *b, int objc, unsigned char *const objv[]
...
+      int index;
+      if (!(std::istringstream(param) >> index)) {
+	result = "bias gradient: error parsing bin index";
+	return COLVARSCRIPT_ERROR;
+      }
+      result = cvm::to_str(-b->bin_gradient(index));
```
giacomofiorin: Legal syntax, but confusing to read.

src/colvarscript.cpp
jeffcomer: Sure, I can call it "bin_forces". For consistency, maybe I should make "bin_count" an alias for "count", while maintaining "count" for backwards compatibility?

I shouldn't have a problem implementing "bin_energy", "bin_forces" and "forces" once I get the other issues worked out.

src/colvarscript.cpp
jeffcomer: Is it the minus sign that you find confusing?

src/colvarbias.cpp
jeffcomer: > Can you use the existing energy command to return the bias's energy, i.e. the negative of what you return as freeenergy? This would allow replica exchange scripts to be more uniform between biases.

As you noted below, I sometimes need to be able to get the bias energy at positions other than current. It does seem reasonable that colvarbias_abf should set bias_energy like the other biases. I could then, as you suggested below, just use "cv bias abf1 energy" to get the current energy. It looks like I could just set bias_energy in colvarbias_abf::update. If I do this, will the ABF energy be included in the energies reported by NAMD?

> If only 1D ABF is tested, I strongly recommend making this feature incompatible with multi-dimensional ABF.

I'm unsure how to check if the ABF is multidimensional.

> In 2D and 3D ABF, you could rely on the existing integrated gradients, especially if they are only necessary at sparse times. The fullSamples question is less pressing then, because noisy gradient estimates get smoothed out by the Poisson integration process.

Where do I get these integrated gradients?

I'd also really like to make this work for eABF, but I don't quite understand how to do this yet. I'm a little confused by "cv[i]->actual_value()" and "cv[i]->value()".

jeffcomer: > I'd recommend using colvarvalue instead of cvm::real and std::vector<int> instead of just int

I will try this.

giacomofiorin: About `bias_energy`, that's correct: if ABF starts setting it based on the integrated gradients (1D, 2D or 3D) NAMD will report that energy as well.  Please coordinate with @jhenin on this as well as how to use the multi-dim integrators and eABF.

src/colvarscript.cpp
giacomofiorin: That sounds reasonable: `count` itself would be difficult to define for real-valued argument (it needs an interval to compute a proper count).  But `bin_count` would definitely go along well.

src/colvarscript.cpp
giacomofiorin: Yes, what about some parenthesis?  Unlike in Python, unnecessary parentheses are not considered bad style :-)

giacomofiorin:
This has been closed indirectly by removing the branch that contained the code.

@jeffcomer Since the ABF class now sets the `bias_energy` field (see #294) does that effectively superseded this PR?

Issue #224, opened 2019-02-26
colvars plot_colvars_traj.py: error plotting multidim colvars


ReneHamburger1993:
The current version of the script  [colvars/colvartools/plot_colvars_traj.py](https://github.com/Colvars/colvars/blob/master/colvartools/plot_colvars_traj.py)
      
 fails to plot multicomponent colvars such as distanceVec

both at lines 336 and 396:
```
            for ic in range(cv.num_dimensions):
                y = cv.values
                if (cv.num_dimensions > 1):
                       y = cv.values[ic]
```

fix: 
change " y = cv.values[ic]" to " y = cv.values[ : , ic]"

PR #225, opened 2019-02-26
Fix: #224 plotting multicomponent colvars

ReneHamburger1993:
cv.values returns ndarray with shape: (numberofrecords, dim of respective colvar)


giacomofiorin:
Thank you for fixing this embarrassing mistake! :100:

PR #229, opened 2019-03-03
Emphasize in the doc that keywords are only accepted on new lines

giacomofiorin:


giacomofiorin:
Force-pushed to remove files accidentally committed to doc branch.

PR #233, opened 2019-03-08
Fix #226 by unregistering named atom groups

### LINKED ISSUE #226: Update list of atom group names when deleting a variable
This bug is quite rare because you have to be using atom group names to begin with, and then tinker with the config on the fly. Nonetheless, it needs to be fixed.

jhenin:


jhenin:
Good point. Of course I have to fall back to the now static cvm::main(), but this can be changed transparently in the future.

PR #240, opened 2019-03-20
Write the head of a pmf file

fhh2626:
Write the correct head of outputs when applying abf_integrate, which can be used for post-treatments.

### DIFF
--- a/colvartools/abf_data.cpp
+++ b/colvartools/abf_data.cpp
@@ -5,6 +5,7 @@
 #include <cstring>
 #include <cstdlib>
 #include <ctime>
+#include <iomanip>
 
 /// Construct gradient field object from an ABF-saved file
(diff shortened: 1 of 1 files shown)


jhenin:
Looks good, thank you.

giacomofiorin:
Hey, please update the version strings as well after merging.

jhenin:
Done. These files have barely been touched in recent years...

giacomofiorin:
Indeed.

Would you be willing to use COLVARS_VERSION there as well, though? Users of `abf_integrate` are not aware that it's independent from the library, and may think that bugs fixed in the library after a certain version may still affect the utility.

jhenin:
Good point. Done by b3fe37a04848725b6463738d24bf9e188208e0a3

Issue #239, opened 2019-03-20
Should use long long instead of long or int to represent steps


HanatoK:
My colleague tried to restart from a ABF simulation which the `step` in the state file was larger than the maximum value of `long` data type, and she encountered a bunch of lines containing "Prepared sample and gradient buffers at step XXX" in the NAMD standard output.
I took a look at the code in colvarbias_abf.cpp:
```
  // Prepare for the first sharing.
  if (shared_last_step < 0) {
    // Copy the current gradient and count values into last.
    last_gradients->copy_grid(*gradients);
    last_samples->copy_grid(*samples);
    shared_last_step = cvm::step_absolute();
    cvm::log("Prepared sample and gradient buffers at step "+cvm::to_str(cvm::step_absolute())+".");
  }
```
The `shared_last_step` is a variable of `int` type, and when the step is larger than `2,147,483,647` then it's always less than zero. `2,147,483,647` steps with a 2.0 fs timestep approximates 4.3 microseconds.
Sometimes we need to run simulations more than 4.3 microseconds, so I think use `long long` instead of `long` or `int` as the type of `shared_last_step`, `cvm::step_absolute()` and step-related variables is better.

jhenin:
We could switch at the Colvars level, but keep in mind that NAMD itself stores the time step number as an `int`. See eg. this thread: https://www.ks.uiuc.edu/Research/namd/mailing_list/namd-l.2016-2017/0975.html
To fix that you need to reset periodically to 0 with `firstTimeStep`.

That said, Colvars keeps its own record of the time step through restart files, so moving to a longer type could work independently from the back-end. Thoughts, @giacomofiorin ?

giacomofiorin:
Yeah.  I'd have changed to long long but was waiting to see what specific
type would NAMD converge to (e.g. int64).  Some compilers may nag if they
are spelled differently.  However, even after the introduction of 64bit for
global atom indices that didn't happen yet.  Best to handle it in Colvars
at this point.

We are currently honoring firstTimeStep at setup if it is larger than zero,
so definitely we need to address the case of overflow in NAMD itself.  For
example, by warning the user if the Colvars state file has a larger number
than NAMD can represent for firstTimeStep.  Not sure if this should be an
error.

Also, the type of all step numbers in the subclasses must change as well,
so best to use a typedef here.

HanatoK:
Fix by #241

PR #241, opened 2019-03-20
Use 64-bit integers to store simulation step numbers

giacomofiorin:
Implements #239 

- [x] Define a 64-bit integer type around `int64_t`
- [x] Make global step number use this type
- [x] Convert metadynamics steps (e.g. `hill::it`)
- [x] Convert ABF steps (e.g. `shared_last_step`) 
- [x] Convert restraint steps (e.g. `target_nsteps`)
- [x] Convert `UIestimator::update()` and `colvar::prev_timestep`

### DIFF
--- a/src/colvarmodule.h
+++ b/src/colvarmodule.h
@@ -85,7 +85,7 @@ public:
   friend class colvarscript;
 
   /// Use a 64-bit integer to store the step number
-  typedef int64_t step_number;
+  typedef long long step_number;
 
   /// Defining an abstract real number allows to switch precision
   typedef  double    real;
--- a/src/colvarmodule.h
+++ b/src/colvarmodule.h
@@ -84,6 +84,9 @@ public:
   // TODO colvarscript should be unaware of colvarmodule's internals
   friend class colvarscript;
 
+  /// Use a 64-bit integer to store the step number
+  typedef int64_t step_number;
+
   /// Defining an abstract real number allows to switch precision
   typedef  double    real;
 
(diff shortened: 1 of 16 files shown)


HanatoK:
Actually I have tried to use the same `typedef int64_t` in `cvm` and then overloaded `cvm::to_str`, but the function overloading is failed because on my platform (GCC 8.3.1 on Linux) `int64_t` and `long int` is the same.

src/colvarmodule.h
```diff
@@ -516,6 +519,10 @@ class colvarmodule {
   static std::string to_str(long int const &x,
                             size_t width = 0, size_t prec = 0);
 
+  /// Convert to string for output purposes
+  static std::string to_str(step_number const &x,
```
HanatoK:
```
#include <stdint.h>

typedef int64_t bigint;

double foo(long int x) {
    return x+1;
}

double foo(bigint x) {
    return x+1;
}

int main() {
    double x = foo(static_cast<bigint>(2));
    return 0;
}
```
The code above fails to compile on my computer.

giacomofiorin:
@HanatoK Me too, with earlier versions (e.g. gcc 4.8 in RH7).  In a hurry I forgot that `long int` is actually a 64-bit integer already on most Unix platforms, hence the clash.

Was the platform where #239 occurred a Windows machine by any chance?

giacomofiorin:
Never mind, #239 was dealing with `shared_last_step`, which was inconsistent with the type of the global time step regardless of the platform.

HanatoK:
I did a grep quickly on the source code. The `prev_timestep` in colvar.h is `int` type. Changing it to `cvm::step_number` would be better.

giacomofiorin:
@HanatoK Right on both counts.

giacomofiorin:
@jhenin I'm unsure about `integrateInitialSteps` and `integrateSteps`

jhenin:
Nope, those are iterations of the conjugate gradient solver. I sure hope we don't need billions of those.

PR #243, opened 2019-03-25
Provide accessor to center-of-mass gradient of scalar variables

### LINKED ISSUE #242: Feature request: providing a public method to access the scalar_com_gradient of an atom group
I am working on a colvar constituted by any combination of current available colvar components. A problem is that I need to know the sub-CVCs gradients with respect to their atom groups when the sub-CVCs support explicit gradients, and that is stored in `scalar_com_gradient`. Currently it seems there is no public method to access it.
I think it's somewhat strange that the gradient of an `atom` is public while the gradient of an `atom_group` is not.

giacomofiorin:
Closes #242

### DIFF
--- a/src/colvaratoms.cpp
+++ b/src/colvaratoms.cpp
@@ -1117,13 +1117,12 @@ void cvm::atom_group::set_weighted_gradient(cvm::rvector const &grad)
 {
   if (b_dummy) return;
 
-  if (is_enabled(f_ag_scalable)) {
-    scalar_com_gradient = grad;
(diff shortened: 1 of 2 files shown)


giacomofiorin:
@HanatoK You're right. The field was only useful for scalable groups (where the atoms aren't directly accessible), but if used for other purposes it may as well be set regardless.

The best way to check that the COM gradient is defined would then not be `f_ag_scalable`, but `f_cvc_com_based` for the parent CVC.

PR #246, opened 2019-04-09
Fix missing output of runtime histogram for histogramRestraint

giacomofiorin:
Reported by Rong Shen (U Chicago).  The forces from `histogramRestraint` seem to be applied correctly, but the changes to the `colvarbias` base class have resulted in hiding the `write_restart()` function of `colvarbias_histogram_restraint`.  As a result, `writeHistogram` was effectively ignored and the run-time histogram could not be written.  This change moves the relevant code to `write_output_files()` consistently with other biases.


giacomofiorin:
After more tests, Rong and I agreed that forces aren't applied correctly.  This is due to a forgotten setting of `f_cvb_apply_force`, which sidesteps the setting of the relevant feature in the variable.

@jhenin How do you feel about raising an error when `colvar::add_bias_force()` is called for a variable that doesn't have `f_cv_gradient` enabled?

giacomofiorin:
Rong confirms that the calculation runs correctly. Keeping open until review is done.

src/colvar.h
```diff
@@ -655,6 +655,10 @@ inline colvarvalue const & colvar::total_force() const
 
 inline void colvar::add_bias_force(colvarvalue const &force)
 {
+  if (! is_enabled(f_cv_gradient)) {
```
jhenin: In principle, this kind of sanity checks would not be necessary at run-time in production. I wonder if it could be a good idea to have a build flag enabling these tests, and allowing us to differentiate test builds and production builds? Or it could be combined with debug mode for simplicity...

For now though, I think this is the way to go.

giacomofiorin: This issue was due to an uncommon scenario: a class written before `colvardeps` was introduced, but integrated during the transition period.  A test would have caught it, unfortunately (it will now).

This particular check won't add much overhead, but it's a good suggestion to have the ability to turn them off as needed.  Replacing it with an inline function within `colvardeps`.

giacomofiorin:
@jhenin I made the `feature()` accessor `const`, which is consistent with the dependency code as it is, but making sure.

jhenin:
Unless I'm mistaken, the `const` to the left of the function name made it const, but the canonical position is to the right. So the `const` to the left is now redundant and can be deleted.

giacomofiorin:
The `const` to the left is applied to the returned reference, but the one to the right is applied to the member function itself.

PR #247, opened 2019-04-11
Collect atomic gradients for combination cvcs

jhenin:
- delegate work of colvar::collect_cvc_gradients(), which was very intrusive, to cvc::get_gradients()
- overload virtual get_gradients() in CVCs such as dihedPC
- enables gradient display for dihedPC in VMD dashboard
- could be extended to other "special" CVCs
- in the process fix destructor bug that lead to segfault (double delete)

### DIFF
--- a/src/colvarcomp.h
+++ b/src/colvarcomp.h
@@ -165,7 +165,7 @@ public:
 
   /// \brief Calculate atomic gradients and add them to the corresponding item in gradient vector
   /// May be overridden by CVCs that do not store their gradients in the classic way, see dihedPC
-  virtual void get_gradients(std::vector<int> const &atom_ids, std::vector<cvm::rvector> &atomic_gradients);
+  virtual void collect_gradients(std::vector<int> const &atom_ids, std::vector<cvm::rvector> &atomic_gradients);
 
   /// \brief Calculate the total force from the system using the
   /// inverse atomic gradients
@@ -1152,8 +1152,8 @@ public:
   virtual  ~dihedPC();
   void calc_value();
   void calc_gradients();
-  /// Re-implementation of cvc::get_gradients() to carry over atomic gradients of sub-cvcs
-  void get_gradients(std::vector<int> const &atom_ids, std::vector<cvm::rvector> &atomic_gradients);
+  /// Re-implementation of cvc::collect_gradients() to carry over atomic gradients of sub-cvcs
+  void collect_gradients(std::vector<int> const &atom_ids, std::vector<cvm::rvector> &atomic_gradients);
   void apply_force(colvarvalue const &force);
   virtual cvm::real dist2(colvarvalue const &x1,
                           colvarvalue const &x2) const;
--- a/src/colvarcomp.h
+++ b/src/colvarcomp.h
@@ -163,6 +163,10 @@ public:
   /// \brief Calculate finite-difference gradients alongside the analytical ones, for each Cartesian component
   virtual void debug_gradients();
 
+  /// \brief Calculate atomic gradients and add them to the corresponding item in gradient vector
+  /// May be overridden by CVCs that do not store their gradients in the classic way, see dihedPC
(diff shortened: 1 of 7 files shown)


src/colvarcomp.h
```diff
@@ -1148,6 +1152,8 @@ class colvar::dihedPC
   virtual  ~dihedPC();
   void calc_value();
   void calc_gradients();
+  /// Re-implementation of cvc::get_gradients() to carry over atomic gradients of sub-cvcs
+  void get_gradients(std::vector<int> const &atom_ids, std::vector<cvm::rvector> &atomic_gradients);
```
giacomofiorin: How about `colvar::alpha_angles`?

src/colvarcomp.h
```diff
@@ -163,6 +163,10 @@ class colvar::cvc
   /// \brief Calculate finite-difference gradients alongside the analytical ones, for each Cartesian component
   virtual void debug_gradients();
 
+  /// \brief Calculate atomic gradients and add them to the corresponding item in gradient vector
+  /// May be overridden by CVCs that do not store their gradients in the classic way, see dihedPC
+  virtual void get_gradients(std::vector<int> const &atom_ids, std::vector<cvm::rvector> &atomic_gradients);
```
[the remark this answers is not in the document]
jhenin: The thing is, there is no data structure to hold the gradients at the CVC level. I think that should be the case eventually, or something similar (as part of #61 ), but in the meantime I didn't see a solution to split these.

src/colvarcomp.h
jhenin: Actually writing this raised the question of addressing this situation in a more generic way: a CVC that combines several sub-CVCs. I didn't move forward yet because I see a few obstacles to handling that in an elegant way, e.g.: (1) the precise mathematical combination will be specific to each combination CVC, and (2) the dependency system is not designed to handle this kind of "horizontal" dependency (CVC to another CVC), although that could probably be added.

src/colvarcomp.h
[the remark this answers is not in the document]
giacomofiorin: That is true, but even if the gradients aren't stored yet in the CVC object, they can still be an argument to both functions.

src/colvarcomp.h
giacomofiorin: It seems a bit of work to derive a general template function for only two exceptions. With one already covered, it's mostly a matter of who deals with the other one.

src/colvarcomp.h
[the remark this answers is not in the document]
jhenin: I mean, the results of the gradient calculation cannot be stored anywhere but in the `atomic_gradients` vector. If `collect_gradients` does that, then there is nothing left for an accessor to do. Actually, I think that is the case, so let me rename this to `collect_gradients`.

src/colvarcomp.h
jhenin: That matter is now settled :-)

src/colvarcomp_protein.cpp
```diff
@@ -414,33 +468,10 @@ void colvar::dihedPC::collect_gradients(std::vector<int> const &atom_ids, std::v
...
-      }
-      if (ag.is_enabled(f_ag_fitting_group) && ag.is_enabled(f_ag_fit_gradients)) {
-        cvm::atom_group const &fg = *(ag.fitting_group);
-        for (size_t k = 0; k < fg.size(); k++) {
-          size_t a = std::lower_bound(atom_ids.begin(), atom_ids.end(),
-                                      fg[k].id) - atom_ids.begin();
-          // fit gradients are in the unrotated (simulation) frame
-          atomic_gradients[a] += coeff * fg.fit_gradients[k];
-        }
+      for (size_t k = 0; k < ag.size(); k++) {
```
jhenin: Realized treating all the atom group rotation cases was unnecessary for these sub-CVCs!

giacomofiorin:
Forgot to merge this, thanks.  I updated the version strings in 9fab4ab

PR #248, opened 2019-04-11
Change the sign of the gradient before running abf_integrate

giacomofiorin:
Due to a different convention (`colvarbias_ti` writes the system force, `colvarbias_abf` the free energy gradient) the default behavior leads to an inconsistency (reported by Wei Chen on LAMMPS users mailing list).

This change allows to change the sign of the gradients before integrating them.

The issue should not apply to a large number of users: the main use of `colvarbias_ti` is a 1D TI PMF, and porting the Poisson integrator (up to 3D) would shrink the number of uses for `abf_integrate` outside of ABF even further.


jhenin:
My preferred solution would be for gradient files to contain homogeneous data between ABF and TI... What are the reasons not to do that?

giacomofiorin:
My reasoning was originally to be consistent with a histogram of the total force compute from the `colvars.traj` file.  Admittedly the choice of the `.grad` extension is not great: how about `.force`?  Definitely I wouldn't change the sign in the contents of the file while keeping the same file name.

jhenin:
This is much better with a different extension.

PR #259, opened 2019-06-06
Corretted bug on multiple walkers metadynamics, pmf included two time…

fabsugar:
…s the same replica,

now is fixed.


giacomofiorin:
I introduced this bug in 4d5a3db30e98267f7409842a6d5d01807de7d499, thus versions of NAMD up to 2.12 were unaffected.  The bug affects, however, 2.13b1 and later.  The PMF is written with double-counting, but the state file is consistent.

PR #264, opened 2019-07-02
Fix initialization of eigenvector::ref_pos_center

giacomofiorin:
When doing compilation tests I found that `eigenvector::ref_pos_center` was being redefined inside the constructor thus hiding the member.  It doesn't get used beyond the constructor and there are currently no errors, but this should be fixed anyway should the member be used at some point.


jhenin:
Thanks for the catch. I think the best fix is the one that minimizes the scope of this variable. It can be reinstated as a member if needed in the future, which I doubt will ever happen.

jhenin:
Good to merge, in my opinion.

PR #268, opened 2019-07-05
Remove output files from .gitignore

jhenin:
As noted by @HubLot this was added a long time ago. Now that the regtest script leaves clean directory behind it, maybe we can let git consider these files?


HubLot:
Maybe, a softer approach with inverse patterns could do the trick :

```
*.traj
!**/AutoDiff/test.*.colvars.traj
*.out
!**/AutoDiff/test.*.colvars.out
```
(Not sure about the syntax but you see the idea)

giacomofiorin:
OK.  The issue is that it's not trivial to add regtest reference files because one has to use `git add -f`, correct?

HubLot:
Indeed.
For gromacs regtest files, I just realized that all *.traj and *.out files in AutoDiff folders have not been added during this commit 609c602a9b.
(because when you add directories, the missing (ignoring) files is silent)

PR #276, opened 2019-07-31
Minimize occurrence of file-read errors in multiple-walker metadynamics

giacomofiorin:
@jcphill and Igor Vorobyov report errors on Blue Waters with MW replicas reading each other's files.  Their occurrence is likely due to the combination of high frequency of writes (around a few seconds) and a high load of the shared filesystem, making some files stale.  This PR implements Jim's suggestion of writing state files for the replicas' use as a temporary file, and renaming it in place at the end.


src/colvarbias_meta.cpp
```diff
@@ -1782,62 +1779,27 @@ int colvarbias_meta::write_replica_state_file()
...
-  //   for (std::list<hill>::const_iterator h = this->hills_off_grid.begin();
-  //        h != this->hills_off_grid.end();
-  //        h++) {
-  //     rep_state_os << *h;
-  //   }
-  // }
-  // rep_state_os << "}\n\n";
-  // rep_state_os.close();
+  cvm::proxy->close_output_stream(tmp_state_file);
+  cvm::backup_file(replica_state_file.c_str());
```
jcphill: Calling backup_file is unnecessary, as the file is not actually deleted while any client has it open. This also has the problem of creating a point in time when there is no file of the desired name. If you do want the preserved the behavior of having a backup file it would be better to use std::filesystem::create_hard_link().

src/colvarbias_meta.cpp
```diff
@@ -1782,62 +1779,27 @@ int colvarbias_meta::write_replica_state_file()
...
+  std::rename(tmp_state_file.c_str(), replica_state_file.c_str());
```
jcphill: Check return code of std::rename().

giacomofiorin: I'm looking at `NAMD_backup_file()` to try and be consistent.  Do you recommend to add the extra `remove()` call for Windows and the extra checks for `errno` values?  In that case, it may be useful to just fork `NAMD_backup_file()`.

src/colvarbias_meta.cpp
giacomofiorin: Sounds good, it is indeed not very useful with the renaming.  Deleted this call in b52827a

giacomofiorin:
@jcphill The fatal error when writing the state file was introduced in 11f10963d971dcf48fd97730a7315b2415c9758e (by combining duplicated I/O code for state files from different biases), which precedes NAMD 2.12 chronologically but was merged right after (more precisely, your NAMD Git commit 074e515f97e4f79e7c72df6a7c0318f7605173b9 in February 2017).

src/colvarbias_meta.cpp
```diff
@@ -1782,62 +1779,30 @@ int colvarbias_meta::write_replica_state_file()
   if (cvm::debug()) {
     cvm::log("Writing replica state file for bias \""+name+"\"\n");
   }
-  // write down also the restart for the other replicas
-  cvm::backup_file(replica_state_file.c_str());
-  std::ostream *rep_state_os = cvm::proxy->output_stream(replica_state_file);
+
+  std::string const tmp_state_file(replica_state_file+".tmp");
+  std::ostream *rep_state_os = cvm::proxy->output_stream(tmp_state_file);
```
jcphill: Does output_stream() overwrite an existing .tmp file or do we need to call std::remove here in place of backup_file?

giacomofiorin: No, actually it calls `backup_file()` on that file as well. You're right, maybe it'll be cleaner to just `remove()` the `.tmp` file regardless.

jcphill: As long as it's not a fatal error.

src/colvarbias_meta.cpp
jcphill: NAMD_backup_file does cover the rare edge cases we've encountered over the years, so forking it would make sense. The big one is looping on EINTR, which NAMD_write also does.

src/colvarbias_meta.cpp
```diff
@@ -1782,62 +1779,31 @@ int colvarbias_meta::write_replica_state_file()
...
-  //        h != this->hills_off_grid.end();
-  //        h++) {
-  //     rep_state_os << *h;
-  //   }
-  // }
-  // rep_state_os << "}\n\n";
-  // rep_state_os.close();
+  cvm::proxy->close_output_stream(tmp_state_file);
+
+  if (std::rename(tmp_state_file.c_str(), replica_state_file.c_str())) {
```
jcphill: On Windows only you need to remove before renaming since rename will fail if the target exists, which is the expected case.

jcphill: Basically, this is the same operation as NAMD_backup_file, except adding .tmp to the source rather than .old to the destination.

giacomofiorin: Sounds good.  At any rate, multiple-walkers on Windows?

jcphill: And also an error if source doesn't exist.

jcphill: If you ever want to teach this in a workshop people have Windows laptops, so it would be nice if it didn't just fail and it's easier to fix it now.

src/colvarbias_meta.cpp
giacomofiorin: Done. Part of this code path still doesn't handle EINTR internally, but most of it is file writes, which do use `NAMD_write()` implicitly.  If this change looks reasonable to you, I'll look later for other places where file I/O could use handling of EINTR.  Thanks.

src/colvarbias_meta.cpp
```diff
@@ -3,6 +3,7 @@
 #include <iostream>
 #include <sstream>
 #include <fstream>
+#include <cstdio>
```
jcphill: Is this still needed?

giacomofiorin: Nope.

src/colvarbias_meta.cpp
jcphill: I think this is enough fixing for now.

giacomofiorin:
I ran the NAMD regression tests and the MW metadynamics one in particular.  Ran them twice without cleaning up the output files in between to decrease the changes that I missed something.  The test runs two replicas sequentially, so it's hard to test for race conditions.  Ideally the added error checks should allow identifying problems more easily.

src/colvarbias_meta.cpp
jcphill: I realized in the shower this morning that Windows won't let you remove a file that is open by another process. Since that is likely in this case, you will want (on Windows only) to rename the state file to a .old file rather than trying to remove it. It is possible that _unlink() on Windows does not have this limitation. There is a nice discussion of the pain of deleting files on Windows at https://boostgsoc13.github.io/boost.afio/doc/html/afio/FAQ/deleting_open_files.html

giacomofiorin: OK, but then it's almost just as likely that the `.old` file is also still open, especially when files are written in short increments.  Would we need to change the backup extension, à la GROMACS?

jcphill: I'm assuming that the readers will be able finish reading and close the .old file in less time than the update interval. Nothing should be opening the state file after it is renamed to .old.

giacomofiorin: @jcphill The PR is closed so I'm not sure if there will be a notification; the commit for this is cf5d5b2

giacomofiorin: Scratch that, it's ddb312a

PR #277, opened 2019-08-02
Update NAMD Make.depends based on C++11-only dependencies

giacomofiorin:
The NAMD Makefile target for `depends` does not currently handle files that are conditionally included based on `__cplusplus`, such as `colvar_geometricpath.h`.

For now I re-generated `Make.depends` by turning on `-std=c++11` manually and plan to submit it to NAMD Gerrit as-is.

For the future, options are:
- Regenerate `Make.depends` every time the Colvars version is updated in a NAMD source tree by patching the NAMD `Makefile` to achieve this.
- Modify upstream the NAMD `Makefile` to use at least `-std=c++0x` for `depends`.

### DIFF
--- a/namd/Make.depends
+++ b/namd/Make.depends
@@ -7571,6 +7571,7 @@ obj/colvar.o: \
 	colvars/src/colvarcomp.h \
 	colvars/src/colvaratoms.h \
 	colvars/src/colvarproxy.h \
+	colvars/src/colvar_geometricpath.h \
 	colvars/src/colvarscript.h \
 	colvars/src/colvarbias.h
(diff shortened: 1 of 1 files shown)


HanatoK:
Is `-std=c++0x` the default option in NAMD `Makefile`? 
In the `depends` target of `Makefile` the loop `for i in $(OBJS) ...` has the `-std=c++0x` already, but the loop `for i in $(COLVARSLIB) ...` doesn't have this.

giacomofiorin:
@HanatoK Indeed that's what I'd like to add upstream.

PR #284, opened 2019-10-09
Add Makefile useful for compilation tests

giacomofiorin:
This is a rather simple and fairly portable `Makefile` that builds and links the static library for the purpose of performing compilation tests more quickly.  For obvious reasons, `colvarproxy` interface files with other packages are not included in the build and therefore not covered.

Compilation flags have reasonable defaults for testing (e.g. `-std=c++11 -pedantic`) but can be overridden by environment variables.

Lepton is detected by the presence of a folder called `lepton` inside the `src` folder (similar to what is done in LAMMPS).

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,0 +1,82 @@
+# -*- makefile -*-
+
+# This Makefile is useful for compilation tests only.  To use the library,
+# please use instead the build recipes of the packages that include it.
+
+COLVARS_LIB = libcolvars.a
+COLVARS_OBJ_DIR =
+
+.PHONY: default clean
+
+default: $(COLVARS_LIB)
+
+.SUFFIXES:
+.SUFFIXES: .cpp .o
+
+ifeq ($(CXX),)
+CXX := g++
+endif
+ifeq ($(CXXFLAGS),)
+CXXFLAGS := -std=c++11 -pedantic -g -O2
+ifneq ($(CXX),"CC")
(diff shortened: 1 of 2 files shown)


src/Makefile
```diff
@@ -0,0 +1,82 @@
...
+	@for src in $^ ; do \
+	  obj=`basename $$src .cpp`.o ; \
+	  $(CXX) -MM $(COLVARS_INCFLAGS) $(LEPTON_INCFLAGS) \
+	    -MT '$$(COLVARS_OBJ_DIR)'$$obj $$src >> $@ ; \
+	  done
+
+include Makefile.deps
+
+# No explicit dependencies for Lepton: we assume its object files are always
+# newer than the corresponding source files
```
jhenin: Is there a specific reason for that, or is it just because Lepton updates are expected to be very rare?

giacomofiorin: Partly because they are rare, but mostly because I imagine that the `lepton` folder would be added/removed multiple times when testing the same Colvars work tree.  In that case it is highly recommended to do `make clean` each time, making dependencies much less needed.

This is unlike the LAMMPS tree, where the `lepton` folder is permanently there and it makes more sense to track build dependencies, so there is a `Makefile.lepton.deps`.

giacomofiorin:
Forgot to mention also that to test different flags on the same worktree
(C++11 vs. C++03) make clean is actually required.

PR #285, opened 2019-10-11
Add CMake recipe for tests

giacomofiorin:
- Change default location of Lepton in the top-level folder (more reasonable than inside `src`).

- Allows turning Lepton on or off explicitly with `-DENABLE_LEPTON=ON|OFF`, or implicitly by testing if the `${LEPTON_DIR}` path exists. `LEPTON_DIR` can be provided as a CMake flag, or left at its default location.  `CMAKE_CXX_STANDARD` is left at its default value for the compiler (`98` in most cases) and is raised to `11` when Lepton is enabled.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -1,0 +1,29 @@
+cmake_minimum_required(VERSION 2.8.12)
+
+project(colvars CXX)
+get_filename_component(COLVARS_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/.. ABSOLUTE)
+
+#set(CMAKE_VERBOSE_MAKEFILE on)
+
+option(ENABLE_LEPTON "Build and link the Lepton library" OFF)
+
+file(GLOB COLVARS_SOURCES ${COLVARS_SOURCE_DIR}/src/[^.]*.cpp)
+
+if(ENABLE_LEPTON)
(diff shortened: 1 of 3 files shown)


cmake/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+cmake_minimum_required(VERSION 2.8.12)
+
+project(colvars CXX)
+get_filename_component(COLVARS_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/.. ABSOLUTE)
+
+#set(CMAKE_VERBOSE_MAKEFILE on)
+
+option(ENABLE_LEPTON "Build and link the Lepton library" OFF)
```
jhenin: Could this default to ON if Lepton sources are found?

giacomofiorin: I preferred not to have this as the standard arrangement, since one of the key selling points of CMake are out-of-source builds.  I was able, however, to find out how to handle the commandline-provided path (turned out to be trivial).

giacomofiorin:
Caveat: with the current logic it is not possible to have a `lepton` folder inside the Colvars tree while also disable Lepton with `-DENABLE_LEPTON=OFF`. I consider that an edge case, since with `CMake` the Colvars directory should be kept clean.

PR #287, opened 2019-10-14
Require C++11 for Lepton

giacomofiorin:


giacomofiorin:
For NAMD, you have already uploaded the latest Lepton, and we have seen no problems on older platforms so far.  If there are any, they should probably be addressed by changing the NAMD build system.

For LAMMPS, I'm submitting a PR there to list the requirement in the `README` for the Colvars library and adapting the CMake recipe to enforce the requirement automatically.  Note that they have just started nudging people towards C++11, so it makes sense to make this change now:
https://lammps.sandia.gov/doc/Build_settings.html#cxx11

giacomofiorin:
https://github.com/lammps/lammps/pull/1724

giacomofiorin:
All good with this?  The problem is moot for VMD because it doesn't include Lepton yet, so there is no point in checking for C++11 there.

src/colvar.cpp
```diff
@@ -439,6 +439,21 @@ int colvar::init_custom_function(std::string const &conf)
 
 int colvar::init_custom_function(std::string const &conf)
```
jhenin: Right now the only call to this function is under an #ifdef LEPTON already, so I don't think this version of the function is ever called. But the code looks good to me, if it was to be called.

giacomofiorin: This version would be called if Lepton is turned off in LAMMPS, manually via `COLVARS_LEPTON=no` or automatically by `DISABLE_CXX11_REQUIREMENT`.  It is possible, though rare, that someone builds Colvars with C++11 but with Lepton explicitly disabled.  Hence the second `#ifdef`, which is definitely redundant in most cases by far.

PR #291, opened 2019-10-22
Reduce compiler warnings

akohlmey:
This change comments out parameter names where they are not used. That will reduce compiler warnings when compiling with `-Wall -Wextra` and signals to other developers that those arguments are currently not used. It cannot be applied to arguments with default values, so a few of those warnings remain. Also, I have not looked into lepton. This was tested with the standalone compilation test version.

### DIFF
--- a/src/colvar.cpp
+++ b/src/colvar.cpp
@@ -680,7 +680,7 @@ int colvar::init_output_flags(std::string const &conf)
 // read the configuration and set up corresponding instances, for
 // each type of component implemented
 template<typename def_class_name> int colvar::init_components_type(std::string const &conf,
-                                                                   char const *def_desc,
+                                                                   char const * /* def_desc */,
                                                                    char const *def_config_key)
 {
   size_t def_count = 0;
@@ -1245,7 +1245,7 @@ int colvar::collect_cvc_data()
 }
 
 
-int colvar::check_cvc_range(int first_cvc, size_t num_cvcs)
+int colvar::check_cvc_range(int first_cvc, size_t /* num_cvcs */)
 {
   if ((first_cvc < 0) || (first_cvc >= ((int) cvcs.size()))) {
     cvm::error("Error: trying to address a component outside the "
--- a/src/colvar_geometricpath.h
+++ b/src/colvar_geometricpath.h
@@ -121,7 +121,7 @@ void GeometricPathBase<element_type, scalar_type, path_type>::initialize(size_t 
 }
 
 template <typename element_type, typename scalar_type, path_sz path_type>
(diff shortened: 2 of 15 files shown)


jhenin:
Thanks Axel! Most of these are virtual functions where parameters are used only in some implementations (when they aren't mere stubs). Is having the parameter type but no name standard practice for this case? That seems useful.

akohlmey:
> Thanks Axel! Most of these are virtual functions where parameters are used only in some implementations (when they aren't mere stubs). Is having the parameter type but no name standard practice for this case? That seems useful.

Yes, that is part of the C++ standard. For example in C++ it is allowed to do:
```
int main(int, char**) {
    return 0;
}
```
What I consider good practice (and what we adopted in LAMMPS for example), is to leave the parameter names in place, but comment them out. So the example would become:
```
int main(int /* argc */, char** /* argv */) {
    return 0;
}
```
This way you a) signal your intention to not use those arguments (even though they are required through the prototype/base class), but you also b) communicate what the names would be, in case the code is changed to use them (and thus keep a consistent naming scheme across similar functions).

The only case where you cannot do this is for functions with default arguments. In that case the default arguments would have to be replaced with overloading the function to have multiple signatures and then have the one with fewer arguments call the ones with more and the default value applied. I have not done this, since I think that defeats the purpose of improving the readability and maintainability of code while reducing warnings that have no value and may hide warnings that should be paid attention to. So having a few of those warnings left is not such a big deal. I should emphasize that the "unused parameter" warning is not enabled by `-Wall` typically, so it is really a minor thing and only really important when auditing code. On LAMMPS, for example, we have a post-processing of the warnings issues by compilers, so we can see, if modifying code will increase or decrease warnings in the Jenkins status pages.

giacomofiorin:
@akohlmey I reverted one edit that eventually would break things in other branches: it is in a function whose body is generated by a macro, and some versions of it will actually use the parameter. For now the simplest thing is to revert; I'll add later a blanket statement that uses the parameter to the macro to silence that particular instance of the warning.

akohlmey:
>
> @akohlmey <https://github.com/akohlmey> I reverted one edit that
> eventually would break things in other branches: it is in a function whose
> body is generated by a macro, and some versions of it will actually use the
> parameter. For now the simplest thing is to revert; I'll add later a
> blanket statement that uses the parameter to the macro to silence that
> particular instance of the warning.
>

no worries. take what you are happy with and don't use what not. no need to
go through any hoops to silence all those warnings. the point of the PR is
to **reduce** the warnings to the minimum, to make other, more important
warnings stand out more.
some cannot be avoided. especially when using macros. the better approach
there would be to look into replacing those macros with templates, but that
is a lot more work and thus usually not worth it.

giacomofiorin:
@akohlmey Thanks.  Being the macro used to generate a bunch of functions, silencing the warning in it would probably be worth it.  (I'd like to use a template in general, but the stringification is too convenient).

PR #293, opened 2019-11-01
Document compilation details that affect features

### LINKED ISSUE #250: Add subsection to briefly describe compiler options & unsupported features
There are several components or software environment features that affect the functionality of Colvars and are not always determined by the hosting engine alone:
- Tcl interpreter
- Lepton library
- C++11 syntax

A subsection containing just a list of links to the specific features that depends upon them should be enough.

giacomofiorin:
Fixes #250.

Note that the message printing the C++11 standard used at build time is mostly useful in VMD, where C++11 is still rare, and LAMMPS, where it is becoming the new standard but isn't there yet.

### DIFF
--- a/doc/colvars-refman-lammps.tex
+++ b/doc/colvars-refman-lammps.tex
@@ -31,9 +31,3 @@
 
 
 \input{colvars-refman.tex}
-
-
-% Emacs
-% Local Variables:
-% TeX-PDF-mode: t
-% End:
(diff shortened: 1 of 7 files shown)


giacomofiorin:
Yo @jhenin can I get a review, too?

giacomofiorin:
Yo: not funny on an election year, dude!

PR #294, opened 2019-11-02
bias_energy for ABF

jeffcomer:
For one-dimensional ABF, the energy due to the ABF bias is now calculated. As with the other biases, colvarbias::bias_energy is set, which in turn allows the ABF bias energy to appear in the log files of MD codes and to be accessible from the Tcl interface. The purpose of including the energy of the ABF bias is to allow for hybrids of replica exchange and ABF, although it could be useful for other purposes.

The code accounts for fullSamples, so that the ABF bias energy accurately reflects the forces applied. It has been tested to demonstrate that the bias energy is the negative of the ABF potential of mean force estimate. Correct calculation of the bias energy with extended ABF has not yet been tested, but I expect that it will work. 

plot_pmf_agreement1


giacomofiorin:
@jeffcomer I noticed that `calc_energy()` hides a virtual function with slightly different prototype.  This is a warning worth addressing, so I pushed commit 175ff685.  Can you please pull it into your local `master` branch?

src/colvarbias_abf.cpp
```diff
@@ -841,7 +841,7 @@ cvm::real colvarbias_abf::calc_energy()
   if (num_variables() != 1) return 0.0;
   
   // Get the home bin.
-  int home0 = gradients->current_bin_scalar(0);
+  unsigned int home0 = gradients->current_bin_scalar(0);
   if (home0 < 0) return 0.0;
```
giacomofiorin: If `home0` is unsigned, won't this return always `0.0`?

src/colvarbias_abf.cpp
```diff
@@ -832,3 +835,43 @@ int colvarbias_abf::write_output_files()
...
+{
+  if (num_variables() != 1) return 0.0;
+  
+  // Get the home bin.
+  int home0 = gradients->current_bin_scalar(0);
+  if (home0 < 0) return 0.0;
+  int home = (home0 < gradients->number_of_points(0))?home0:gradients->number_of_points(0)-1;
+ 
+  // Integrate the gradient up to the home bin.
+  double sum = 0.0;
```
giacomofiorin: It would be best to use `cvm::real`. We probably won't ever use single-precision math for production, but this is not guaranteed.

src/colvarbias_abf.h
```diff
@@ -159,7 +159,9 @@ class colvarbias_abf : public colvarbias {
   virtual std::istream& read_state_data(std::istream&);
   virtual std::ostream& write_state_data(std::ostream&);
   virtual int write_output_files();
+
+  /// Calculate the bias energy for 1D ABF
+  cvm::real calc_energy();
```
giacomofiorin: I changed this to remove the hiding warning, as stated in comments.

src/colvargrid.h
```diff
@@ -416,6 +416,20 @@ template <class T> class colvar_grid : public colvarparse {
...
   }
 
+  /// \brief Report the fraction of bin beyond current_bin_scalar()
+  inline cvm::real current_bin_scalar_fraction(int const i) const
+  {
+    return value_to_bin_scalar_fraction(use_actual_value[i] ? cv[i]->actual_value() : cv[i]->value(), i);
+  }
+
+  /// \brief Use the lower boundary and the width to report which bin
+  /// the provided value is in
```
giacomofiorin: Update comment?

jeffcomer: I'm not sure what mean here. Is the comment for "current_bin_scalar_fraction" not clear?

src/colvarbias_abf.cpp
jeffcomer: Yes, this should be an `int`. Making this change.

src/colvarbias_abf.cpp
jeffcomer: I'll change this to cvm::real.

src/colvarbias_abf.h
jeffcomer: I didn't want to shadow calc_energy(). I changed the name of my function to "calc_abf_energy" to avoid this. I just wanted to put all the logic to integrate the gradient in a separate function.

jeffcomer: Sorry, hold on. I'm that good with Github. Let me see if I can do this the way you suggested, by first pulling your commit.

jeffcomer: Okay, I think I correctly integrated your commit.

jeffcomer:
Okay, I think I made the modifications as you suggested.

src/colvargrid.h
giacomofiorin: I mean the comment of `value_to_bin_scalar_fraction()` not in sync with the one of `current_bin_scalar_fraction()`. The comment doesn't change behavior so I'll merge anyway and update later.

giacomofiorin:
@jeffcomer @jhenin I just realized after merging that this change should be documented. dee72c88 does so: I don't think it's worth a PR, but please take a quick look.

PR #295, opened 2019-11-08
Support different unit systems

jhenin:
This is a minimal implementation, in that only VMD actually supports different units, while the MD engines just check that Colvars units are identical to their own internal convention.

- VMD accepts changing unit system only when no colvars are currently
defined to avoid conflicting units between different inputs.
- the VMD Dashboard supports switching units when applicable
- NAMD only accepts the unit system "real"
- Gromacs only accepts the unit system "gromacs" (to be done in gmx branch)
- LAMMPS only accepts its currently set unit system (added test)

### DIFF
--- a/src/colvarproxy.h
+++ b/src/colvarproxy.h
@@ -48,11 +48,47 @@ public:
   /// Destructor
   virtual ~colvarproxy_system();
 
-  /// \brief Value of the unit for atomic coordinates with respect to
-  /// angstroms (used by some variables for hard-coded default values)
-  virtual cvm::real unit_angstrom() = 0;
+  // **********************************************************************
+  // TODO FIXME unsolved: calls to back-end PBC functions assume back-end length unit!
+  // We can use different unit from back-end in VMD bc using PBC functions from colvarproxy base class
+  // or test and default to internal PBC functions if we have our own units?
+  // Possible performance cost? a couple floating-point ops per coordinate
+  // **********************************************************************
 
-  /// \brief Boltzmann constant
+  // Colvars internal units are user specified, because the module exchanges info in unknown
+  // composite dimensions with user input, while it only exchanges quantities of known
+  // dimension with the back-end (length and forces)
+
+  /// \brief Name of the unit system used internally by Colvars (by default, that of the back-end).
+  /// Supported depending on the back-end: real (A, kcal/mol), metal (A, eV), electron (Bohr, Hartree), gromacs (nm, kJ/mol)
+  std::string units;
+
+  /// \brief Request to set the units used internally by Colvars
+  virtual int set_unit_system(std::string const &units, bool colvars_defined) = 0;
+
+  /// \brief Value of 1 Angstrom in the internal (front-end) Colvars unit for atomic coordinates
+  /// * This defaults to the back-end unit
+  /// * can be user-specified once, and can only be changed by resetting the module
+  /// as user-defined values in composite units must be compatible with that system
+  /// Attempts to change it will raise errors
+  cvm::real angstrom_value;
+
+  /// \brief Value of 1 Angstrom in the backend's unit for atomic coordinates
+  virtual cvm::real backend_angstrom_value() = 0;
+
+  /// \brief Value of 1 kcal/mol in the internal Colvars unit for energy
+  cvm::real kcal_mol_value;
+
+  /// \brief Value of 1 kcal/mol in the backend's unit for energy
+  //virtual cvm::real backend_kcal_mol_value() = 0;
+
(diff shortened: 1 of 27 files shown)


giacomofiorin:
I pushed a commit to #295 that also contains a warning. To avoid flooding the output, the warning is printed only upon first use of `cvm::load_coords_xyz()`. What do you think?

jhenin:
As we know, the output is verbose and rarely read, but we can hope that users who notice suspicious results will comb through it.

jhenin:
This passes regression tests and I think it implements what we need. Okay to merge?

giacomofiorin:
Reviewing this shortly.

src/colvarproxy.h
```diff
@@ -48,11 +48,53 @@ class colvarproxy_system {
   /// Destructor
   virtual ~colvarproxy_system();
 
-  /// \brief Value of the unit for atomic coordinates with respect to
-  /// angstroms (used by some variables for hard-coded default values)
-  virtual cvm::real unit_angstrom() = 0;
+  // **********************************************************************
+  // TODO FIXME unsolved: calls to back-end PBC functions assume back-end length unit!
+  // We can use different unit from back-end in VMD bc using PBC functions from colvarproxy base class
+  // or test and default to internal PBC functions if we have our own units?
```
giacomofiorin: The needed check is between two versions of the PBC functions: one using the backend's PBC handler (or in any case, its unit cell parameters) and one using internal parameters entirely.  Since the units are (or at least should be) set at init time, I think this can be done with class polymorphism.  Objects could store a pointer to a "PBC handler" object, which has several variants, including calling the backend's function or handling the PBCs internally (with internal units).  Another variant of this object could be to just use straight differences, which would eliminate a lot of the checks for `f_cvc_pbc_minimum_image`. What do you think?

src/colvarscript.h
```diff
@@ -292,6 +293,14 @@ extern "C" {
            return COLVARS_OK;
            )
 
+  CVSCRIPT(cv_units,
+           "Get the current Colvars unit system",
+           0, 0,
+           { },
+           script->set_str_result(cvm::proxy->units);
```
giacomofiorin: How about having a setter as well?

src/colvarproxy.h
```diff
@@ -48,11 +48,53 @@ class colvarproxy_system {
...
+  cvm::real angstrom_value;
+
+  /// \brief Value of 1 Angstrom in the backend's unit for atomic coordinates
+  virtual cvm::real backend_angstrom_value() = 0;
+
+  /// \brief Value of 1 kcal/mol in the internal Colvars unit for energy
+  cvm::real kcal_mol_value;
+
+  /// \brief Value of 1 kcal/mol in the backend's unit for energy
+  //virtual cvm::real backend_kcal_mol_value() = 0;
```
giacomofiorin: Why are this and the analogous function commented out?

src/colvarproxy.h
jhenin: Sorry, I introduced this note early in the process and it's outdated now. Since then the problem has been fixed in VMD by adapting our internal wrapping function, and for all other backends we use the same units internally. I'm changing this comment to reflect the current situation.

src/colvarscript.h
jhenin: I considered it, and there are arguments going both ways: because parsing a one line config string with the unit keyword does exactly the same job, and that's what the dashboard does now. So functionally it would be redundant, but it would be nice for the completeness of the scripting interface itself - which I think is important. So ok, I'm adding a setter.

src/colvarproxy.h
jhenin: Because they are not used now, but would be used in a possible implementation that does perform energy unit conversion - either in any MD engine, or in VMD if we say got forces from IMD. To unclutter the code, I'll remove them now. If needed they'll be easy to add back.

jhenin:
@giacomofiorin Thanks for all the good comments. I think I'm just missing the docs at this point.

jhenin:
Github prompted me to resolve a conflict with master but that created a nasty merge commit that I just deleted, hence the force-push.

jhenin:
@giacomofiorin Does commit e087781 give all necessary info?

giacomofiorin:
I had minor conflicts when rebasing.  I believe I resolved them correctly, but take a look.

Related to this PR (but not needed to merge it), how about saving the units keyword in the state file?  I'd have tried it already if the extra content didn't invalidate the reference for the regression tests.  We could theoretically strip the units keyword out when comparing...

jhenin:
Sounds like a good idea. A state file will certainly not be compatible with a run in different units.

Issue #252, opened 2019-04-25
Strange boundary effect in extended-Lagrangian based simulation


fhh2626:
Hi, all,

Recently, when I ran a eABF (or its variants) simulation using stratification strategy, the free energies close to window boundaries look quite weird:
![Figure_1](https://user-images.githubusercontent.com/25058290/56710208-902a0780-6757-11e9-88ca-7279146474d0.png)
(The PMF describing the hydration of methane is calculated using the files of NAMD ABF turorial. I only changed ABF to eABF)

Sometimes the little noise are not very significant (for example, in the deca-alanine case), but the problem should exist whatever the system is.
The CZAR, UI and naive estimator give similar results, so this should not be a issue of estimator.
I'm not should if this problem exists in an old version of Colvars, but it seems this is the first time I got the problem.

Thanks,
Haohao

jhenin:
Thank you for the report @fhh2626 and sorry about the delay in replying.

Do the artefacts also appear in the PMFs for individual windows? How did you merge them?

You may be aware of this already: getting a PMF from stratified eABF is a little special because the free energy estimators depend on histograms, and the histograms cannot be simply merged. Instead the gradients should be obtained for each window separately, then merged, then integrated.

fhh2626:
Thanks for your reply @jhenin .

What you said just implies that one cannot use "InputPrefix" to merge czar files, even with "writeCZARwindowFile". Am I correct? If so, then there is no way to use pABF with stratification strategy. Is it possible to have pABF in abf_integrate?

Thanks,
Haohao

jhenin:
> there is no way to use pABF with stratification strategy

You can use pABF in each window. Once you gather the window together, the CZAR scheme is essentially broken, because there is no continuous histogram. I suppose one could cheat and rescale the different zcount files to join them, WHAM-style.

> Is it possible to have pABF in abf_integrate?

I'm not sure what you mean.

fhh2626:
Thanks for your explanation. Now, if one wants to use pABF, an explicit call of NAMD must be made, which means one cannot integrate a gradient file easily using only grad and count file. It would be, if possible, very good to have something like abf_integrate, but using a pABF-like integrator.

jhenin:
Let me add something on the vocabulary here - maybe we need to improve the documentation on this.

[...]

Issue #245, opened 2019-04-01
Why does the harmonicWalls restraint apply on actual CVs instead of extended CVs?


HanatoK:
I am a little curious about the harmonic walls implementation with extended Lagrangian dynamics. Is applying the wall potential to actual CVs theoretically better than to extended CVs?

jhenin:
Very good question. The reason is that is makes it easier to use a histogram-based estimator after eABF. Such estimators (like CZAR) rely on the histogram of the actual CV. If the wall potential is applied to the extended coordinate, then it "bleeds" into the distribution of the actual CV, giving a wrong free energy gradient near the wall.

HanatoK:
@jhenin Thanks for your answer!

fhh2626:
> 
> 
> Very good question. The reason is that is makes it easier to use a histogram-based estimator after eABF. Such estimators (like CZAR) rely on the histogram of the actual CV. If the wall potential is applied to the extended coordinate, then it "bleeds" into the distribution of the actual CV, giving a wrong free energy gradient near the wall.

@jhenin What about letting CZAR support the "expandBoundaries" ? Occationally I found adding wall potential on the actual CV leads to strange performance in meta-eABF and extended-Lagrangian-based metadynamics.

giacomofiorin:
@jhenin can reply on whether it's appropriate to port `expandBoundaries` to ABF: right now, it's pretty much contained in `colvarbias_meta::update_grid_params()`.

However, I'm missing the connection with the wall potentials.

jhenin:
I think the problem with expandBoundaries is that it would make the sampling histogram decay to zero at the boundaries, which is acceptable for single-window sampling but not really for stratified sampling, where you want accurate free energy derivatives all the way to the boundaries.

PR #299, opened 2019-12-04
Option to determine where to add harmonic-wall biases

fhh2626:
Based on the disscussion of #245 and #252 , I feel that adding harmonic-wall restraints on the real CVs rather than the extended ones may cause some boundary effects in the extended-system-based algorithms (eABF, eMtD, meta-eABF, etc.). Plus, this kind of restraints feel unconfortable for me, since importance-sampling algorithms sometimes try to cancel the harmonic-wall potential.

I added an option "biasOnExtendedValue" for "harmonicWalls" to determine whether the harmonic-wall biases are added onto the extended degrees of freedom. The default value is no, which adapts the Colvars convention.

### DIFF
--- a/src/colvarbias_restraint.h
+++ b/src/colvarbias_restraint.h
@@ -280,6 +280,9 @@ protected:
   /// \brief If both walls are defined, use this k for the upper
   cvm::real upper_wall_k;
 
+  /// \brief Whether add biases on the extended DOF
+  bool b_bias_on_extended_value;
+
   virtual cvm::real colvar_distance(size_t i) const;
   virtual cvm::real restraint_potential(size_t i) const;
   virtual colvarvalue const restraint_force(size_t i) const;
--- a/src/colvarbias_restraint.cpp
+++ b/src/colvarbias_restraint.cpp
@@ -1055,6 +1055,8 @@ int colvarbias_restraint_harmonic_walls::init(std::string const &conf)
     }
   }
 
+  get_keyval(conf, "biasOnExtendedValue", b_bias_on_extended_value, false);
+
   return COLVARS_OK;
 }
 
@@ -1083,7 +1085,11 @@ void colvarbias_restraint_harmonic_walls::communicate_forces()
                variables(i)->name+"\".\n");
     }
     // Impulse-style multiple timestep
-    variables(i)->add_bias_force_actual_value(cvm::real(time_step_factor) * colvar_forces[i]);
+    if (b_bias_on_extended_value) {
+      variables(i)->add_bias_force(cvm::real(time_step_factor) * colvar_forces[i]);
+    } else {
+      variables(i)->add_bias_force_actual_value(cvm::real(time_step_factor) * colvar_forces[i]);
+    }
   }
 }
 


src/colvarbias_restraint.h
```diff
@@ -280,6 +280,9 @@ class colvarbias_restraint_harmonic_walls
   /// \brief If both walls are defined, use this k for the upper
   cvm::real upper_wall_k;
 
+  /// \brief Whether add biases on the extended DOF
+  bool b_bias_on_extended_value;
```
giacomofiorin: @fhh2626 Why not make it part of the `colvarbias` base class?

jhenin: I agree.

jhenin:
If this option is provided, we'll need to give guidelines as to when it should be used.

src/colvarbias_restraint.h
giacomofiorin: This is not on the shared repository, so I probably can't push to it.

giacomofiorin: Better yet, this should be a feature, not a boolean flag.  Please leave as is until merged.

giacomofiorin:
@jhenin @fhh2626 This sounds like it is probably ready, but some doc changes are required.

src/colvarbias_restraint.h
fhh2626: Sorry, I'm a little bit confused. Whenever a extended DOF is added into the system, it is always implied that the biases are added on the extended DOF. `colvarbias_restraint_harmonic_walls` is the an exception, isn't it? Then, what does it mean if `b_bias_on_extended_value` is part of `colvarbias`?

giacomofiorin: What you are proposing is adding a keyword that allows the user to control this behavior in a way that is general to many classes: it makes sense for consistency to have it available everywhere. The default value for this new feature could be `true` for the for most biases, and `false` for `harmonicWalls` specifically.

If a new restraint in the future needs to behave the same way as `harmonicWalls`, this can be achieved by making the default value in the same class `false` as well (as opposed to duplicating the relevant code).

Having a `feature` instead of a simple `bool` flag also enables to do internal error checks at run time; but features are defined at the base-class level (and may or may not be used by the derivative classes).

giacomofiorin:
@jhenin To clarify, since you recommend to have guidelines describing this behavior, I'm inclined to wait merging this until these guidelines are included in the doc. The PR is assigned to you so that you decide if and when to merge this.

jhenin:
In order not to mess with your master branch @fhh2626 I have created branch [`bias_ext_coord`](/Colvars/colvars/compare/bias_ext_coord).

On thing that was missing from your implementation was getting the extended value of the colvars to compute the biases.

The `colvarbias` base class cannot implement that in a generic way because there is no unified mechanism for it to access colvar values (each bias calls accessors in the `colvar` class). So for now only one type of bias implements it.

This raises a non-trivial question: what to make of `subtractAppliedForce` in both cases. The current behavior is that every bias but `harmonicWalls` adds their biasing forces to `previous_colvar_forces`, to be subtracted from the total force if requested. I think that should happen only if the force is applied to the actual colvar, otherwise it is not (directly) taken into account in the total force calculation. Even though it does affect the actual colvar through coupling with the extended DOF.

giacomofiorin:
@jhenin @fhh2626 Does #321 implement the substance of this PR?

fhh2626:
@jhenin Oh, I didn't realize that you have already done extensive work in #321. I'm really sorry for my unavailability in the past few days.
@giacomofiorin Yes, I think so.

jhenin:
@fhh2626  While documenting this option, I've decided to rename it `bypassExtendedLagrangian`. It's also implemented for histograms now, which is useful for analyzing extended Lagrangian simulations.
See commit 866b7739eb7aa8e7c09f34884d1852bf230d798a

giacomofiorin:
The new name is much clearer.

Since you have added support for it in `histogram` as well, it may be a good idea to add a pair of tests for it, one with the flag disabled and one enabled.

jhenin:
@giacomofiorin Very good point!

fhh2626:
> @fhh2626 While documenting this option, I've decided to rename it `bypassExtendedLagrangian`. It's also implemented for histograms now, which is useful for analyzing extended Lagrangian simulations.
> See commit [866b773](https://github.com/Colvars/colvars/commit/866b7739eb7aa8e7c09f34884d1852bf230d798a)

Thanks for your information.

jhenin:
@fhh2626 Please note that the implementation was incomplete as the user keyword could not disable the feature. This is fixed by ed5d88962f4fabdf95b94fe012fe42d9dc90ed1c

PR #302, opened 2019-12-05
Allow redefining existing index groups (warn when this happens)

### LINKED ISSUE #215: Reading multi-part configurations across different back-ends
A topic for discussion: in NAMD, multiple config files can be read using the scripting interface. The most interesting use case for me now is having one file with colvar definitions, written using the VMD dashboard, and one with biases.

For that to make sense, we'd need a system to read input from several files for each back-end. Any thoughts on that?

### LINKED ISSUE #255: Limitation of using multiple input files
Hi,

Following the discussion in #215, I implemented this feature in the Gromacs backend (commit 
e7d4443) but I came across some limitations about `indexFile` keyword.

Let's take 2 colvars input files:

file1.in
```
colvarsTrajFrequency 1
colvarsRestartFrequency 10
indexFile index.ndx

colvar {

    name one

    outputAppliedForce on

    width 0.05

    rmsd {
        atoms {
            indexGroup RMSD_atoms
        }
        refPositionsFile rmsd_atoms_refpos.xyz
    }
}


harmonic {
    colvars        one
    centers        0.01
    forceConstant  0.004184
}
```
file2.in
```
indexFile index.ndx

colvar {

    name two

    cartesian {
        atoms {
            indexGroup RMSD_atoms
        }
    }
}
```

When Colvars read the second file, it exit with an error about identical group:
```
colvars: Reading new configuration from file "file2.dat":
colvars: # indexFile = "index.ndx"
colvars: Error: the group name "Protein" appears in multiple index files.
```
However, if I remove the `indexFile` keyword in file2.dat, colvars complains it doesn’t know about the group:
```
colvars: Error: could not find index group RMSD_atoms among those provided by the index file.
```

It seems currently impossible to use a defined group in a 2nd config file if an index file has been already used in the 1st config file, at least when using mutiple `colvars->read_config_file()`

giacomofiorin:
Fixes #255.

The only caveat is that what was an error is now a warning; the index group will be-redefined and the user be warned that the new definition will only apply to new variables.

I think it's best to use the most recent definition when there is a clash, but please chip in if you think otherwise.

Also fixes #215 indirectly. LAMMPS was taken care of by #216 and the GROMACS code is currently in #190.

### DIFF
--- a/doc/colvars-refman-main.tex
+++ b/doc/colvars-refman-main.tex
@@ -518,7 +518,8 @@ The following keywords are available in the global context of the Colvars config
     UNIX filename}{%
     This option reads an index file (usually with a \texttt{.ndx}
     extension) as produced by the \texttt{make\_ndx} tool of GROMACS.
-    This keyword may be repeated to load multiple index files: the same group name cannot appear in multiple index files.
+    This keyword may be repeated to load multiple index files.
+    A group with the same name may appear multiple times, as long as it contains the same indices in identical order each time: an error is raised otherwise.
     \cvlammpsonly{In LAMMPS, the \texttt{group2ndx} command can be used to generate such file from existing groups.
     Note that the Colvars module reads the indices of atoms from the index file: therefore, the LAMMPS groups do not need to remain active during the simulation, and could be deleted right after issuing \texttt{group2ndx}.
 }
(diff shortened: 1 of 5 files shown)


jhenin:
I agree that using the most recent definition is better. An even better behavior might be to check the atom list of the index group being redefined, say nothing (no warning) if it is unchanged, but throw an error if it changes. 

Second question: @HubLot said in #255  that index group definitions were not remembered between calls to `colvars->read_config_file()`. I am not sure why that was the case.

giacomofiorin:
Since I ended up using pointers to vectors, it is probably easy to swap in the new vector if it's different, and do nothing otherwise.

I tried using two `indexFile` commands in a row, with only partially overlapping sets of groups, and the set of groups at the end is the union of them.

HubLot:
Great changes, thanks.

Issue #10, opened 2013-06-13
How to handle different units across MD codes?


jhenin:
At this point, we follow the NAMD conventions of Angstroms, kcal/mol, angles in degrees. Is that acceptable for other codes? How flexible can and should we make that?

giacomofiorin:
Solved by 80b8ac

giacomofiorin:
The various extended-Lagrangian options are not compatible with multiple unit systems, and while they are intuitive in the fs-based ones (NAMD, LAMMPS with units real), their values are not easy to convert to other systems to override default values.

jhenin:
Situation clarified a little by 376e173f2c1429958ca06da825f5dbcfb5f99b89. The only limitation at this point is extendedLangevinDamping required in ps-1, which I think is not too much of a constraint. We take advantage of the proxy converting internal time units to fs.

giacomofiorin:
What about extendedTimeConstant, which assumes fs?
```
\item %
  \keydef
    {extendedTimeConstant}{%
    \texttt{colvar}}{%
    Oscillation period of the fictitious particle (fs)}{%
    positive decimal}{%
    \texttt{200}}{%
```

Also, the ps-fs issue is a small thing but it is still rather NAMD-specific.  In the LAMMPS version of the Langevin thermostat:
http://lammps.sandia.gov/doc/fix_langevin.html
1/gamma is the parameter, given in time units (which is fs for the "units real" system).  There were already a couple of people using the Langevin thermostat in Colvars with LAMMPS, and those did read the manual carefully to know the difference.  But confusion is easy, unfortunately.

jhenin:
I could have the proxy expose a conversion factor for time units, and then take user input in whatever LAMMPS is setup to use. That would break current inputs that provide fs and ps-1, though. We need a good medium to warn our LAMMPS users of this change.

giacomofiorin:
For breaking with a previous convention the cleanest way would be to have a new keyword replace the old one (as we did with `outputTotalForce` vs `outputSystemForce`).  LAMMPS users generally write their own input files and can edit them if needed, especially for a seldom used feature (which you are now proposing to use more often for eABF).

[...]

PR #303, opened 2019-12-05
Check units upon restarting

giacomofiorin:
I couldn't make the check run every time because `units` is left empty unless provided by the user.

If that's OK, what is missing for #10?


giacomofiorin:
Don't worry about the conflicting history, I'll cherry pick.

jhenin:
> If that's OK, what is missing for #10?

I couldn't resist the pleasure of closing an issue from 2013...

jhenin:
> Don't worry about the conflicting history, I'll cherry pick.

Done.

PR #305, opened 2019-12-05
Get current VMD molid from [cv molid]

jhenin:
useful for Dashboard to handle many molecules.


jhenin:
This puts an end to the Dashboard looking at the top molecule just in case Colvars might be connected to it.

PR #306, opened 2019-12-05
Experimental: try to deal with biases and general config in Dashboard

jhenin:
Due to reports of user frustration to be unable to see biases at all in the Dashboard, and having to maintain separate files, I am looking at ways to keep track of them in this branch.

The main difficulty is knowing what parts of the config strings that have been parsed are still current, and what parts have since been superseded.

### DIFF
--- a/vmd/cv_dashboard/cv_dashboard.tcl
+++ b/vmd/cv_dashboard/cv_dashboard.tcl
@@ -37,6 +37,7 @@ namespace eval ::cv_dashboard {
   variable filetype "atomsFile"
   variable colvar_configs  ;# dictionary mapping names to cfg strings
   set colvar_configs [dict create]
+  variable non_colvar_config "" ;# config string for everything but colvars
 
   # Handle to keep track of interactive plot
   variable plothandle
@@ -158,7 +159,8 @@ proc ::cv_dashboard::apply_config { cfg } {
   set cvs_after [run_cv list]
 
   # Extract config for individual colvars
-  set cv_configs [extract_colvar_configs $cfg]
+  lassign [extract_colvar_configs $cfg] cv_configs non_colvar
+  append ::cv_dashboard::non_colvar_config $non_colvar
 
   # Completely update the map of colvar configs
   set new_map [dict create]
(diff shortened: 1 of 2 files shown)


jhenin:
To rephrase, this is at the intersection between restarts and config strings: how to create a config strings that will produce the current state of colvars (except the local changes in bias parameters due to scripting).

For biases, there is a relatively easy solution: add a scripting command` cv bias <name> getkeyword`, which together with `getconfig` lets us recreate a config string (unfortunately, stripped of comments).

For general parameters, I think that would require active help from the parser to remember the relevant keyword/value pairs.

giacomofiorin:
I made some small changes into this branch before forgetting about them (c74abbc). I thought that it would help usability if `update` and `reset` are no-op when the module isn't there.

jhenin:
Rebased onto master.

giacomofiorin:
@jhenin I removed my changes to proxy from this branch: they were now full of conflicts with `master`, so I just recreated them there directly (50be17e948706427303fcb37824eecd7a02a3be0), together with a small fix that raises an error when somebody chooses a negative molid instead of assuming that they just meant `top` (7421954d95f5a00067e453ff4a18e0890bf3ebd7).

Converting the PR as draft until conflicts in the Dashboard are resolved.

jhenin:
New discussion in connection with #397 

It would make sense to display a table of biases somewhere, but so far we were not really equipped to keep track of their config string, and hence offer the option to edit it of save it to a file. Now that the regex parsing is more mature, however, it is starting to seem more feasible. The main hurdle is that they are controlled by bias-specific keywords. For colvars we match the config and the colvar using the name property, however that is not mandatory for biases, and the default depends on the number of biases of a given type that exist, so guessing that might depend on a heuristic.

So there could be a tab for biases, and another tab for the general config of the Colvars Module, with the option to simply edit the text.

jhenin:
To deal with time-dependent biases, a nifty extra would be a way to map trajectory frames in VMD to the correct simulation timesteps. Not trivial.

jhenin:
Superseded by #438

PR #308, opened 2019-12-07
Dashboard: auto-updating atom selections

jhenin:
To take full advantage of the option to switch between molecules in VMD, we need to adapt atom groups when the atom numbering changes between molecules, but the groups can be identified using VMD selections (eg. comparing homologous proteins or simulations with slightly different system compositions). This case is handled by this feature: atomNumbers keywords inserted by the dashboard are accompanied by a magic comment that instructs the Dashboard to recompute the atom selection when the config is applied, ie. when changing molecules.


giacomofiorin:
@jhenin I cherry-picked `0ca89cb` (`const` fix) onto `master` and rebased this branch accordingly.

jhenin:
Ok. This is good for merging on my end.

PR #309, opened 2019-12-09
Automated installation of Dashboard files

giacomofiorin:
Uses the `update-colvars-code.sh` script to update the VMD plugin tree.  Will try to auto-detect whether the plugin tree is "installed" or in still in source form.


vmd/cv_dashboard/pkgIndex.tcl
```diff
@@ -1,4 +1,4 @@
-# Tcl package index file, version 1.1
+# Tcl package index file, version 1.2
```
jhenin: This is a trick: it's the version number of the Tcl utility that created the index file.

That said, I think it is time to bump the Dashboard version.

PR #310, opened 2019-12-10
Provide access to select parameters from the CVCs

### LINKED ISSUE #269: Changes to improve integration of DEER code
The aspects that touch on general code organization are becoming more relevant, so I'm opening an issue to not restrict the conversation to just myself and @fabsugar.  The `rad` bias and `deer` components require lots of "vertical" communication, e.g. between the bias and components.  The following are changes that I thing are needed or at least very desirable:

- [ ] Use the `f_cv_single_component` feature more transparently e.g. by having an accessor (`single_component()`?) that returns the `colvar::cvcs[0]` pointer after checking that the feature is enabled.  This arrangement is akin to `colvar::dist2()` and `f_cv_homogeneous`, which exists already.
- [ ] Provide accessors so that `cvc` member functions that biases can call them.  In the case at hand, `update_params_rad()` and `update_params_rad_chis()` are ad-hoc wrappers which could be better handled by a general method.  
- [ ] Use `cvc` internal fields to set default values for the parent `colvar` object. I already to have components set values like `lowerBoundary` (e.g. `rmsd` setting it to 0). @fabsugar pointed out that for `deer`- and `deerKernel`-based variables the default `width` of 1 makes no sense, and would like to have the `cvc` supply it.  The conditions under which this can be done are already captured by `f_cv_homogeneous` unless I am mistaken.

giacomofiorin:
When working on #269 I realized that the corresponding design would eventually lead to adding a large number of functions when features are added, and these functions are further duplicated by the need to traverse the `colvar` class.

I implemented a more generic access mechanism using string identifier for the parameters. The API allows reading the parameter's value and the gradient of the variable with respect to it. The former uses a generic `void *` pointer, and can abstract in principle any data type. The second uses `colvarvalue *`, which is suitable for the gradient of a real-valued parameter (the most common case by far where one would want to use the gradient).

For more complex properties than the gradient (e.g. the "epsilon" function used in the parameters of the `deer` class) I assume that using a string with a prefix or suffix would work. In theory I could have made the gradient visible via `d_xi/d_pname`, where `pname` is the given parameter, but having a specialized function avoids having to typecast all the time (which may get confusing with `colvarvalue`).

Fixes #269 (by means of a slightly different design).


src/colvarcomp.cpp
```diff
@@ -291,6 +297,105 @@ colvar::cvc::~cvc()
...
+  }
+  return result;
+}
+
+
+int colvar::cvc::set_param(std::string const &param_name,
+                           void const *new_value)
+{
+  if (param_map.count(param_name) > 0) {
+
```
HanatoK: Why not just iterating over the key value of `param_map` and comparing?
```
for (auto it = param_map.begin(); it != param_map.end(); ++param_map) {
  if (param_name.compare(it->first)) {
    // Does the second field of the map point to the variables sup_coeff, sup_np, period and wrap_center?
    it->second = new_value;
  }
}
```

HanatoK:
@giacomofiorin
1. I am not familiar with the DEER code. Is the code in the `rad` branch?
2. What's the meaning of the gradients of parameters?

src/colvar.cpp
```diff
@@ -1892,6 +1892,48 @@ int colvar::update_cvc_config(std::vector<std::string> const &confs)
...
+{
+  if (is_enabled(f_cv_single_component)) {
+    return cvcs[0]->param_exists(param_name);
+  }
+  return cvm::error("Error: calling colvar::cvc_param_exists() for a variable "
+                    "with more than one component.\n", COLVARS_NOT_IMPLEMENTED);
+}
+
+
+void const *colvar::get_cvc_param(std::string const &param_name)
```
HanatoK: It seems it is easy to make it compatible with multiple components. For example:
```
void const *colvar::get_cvc_param(std::string const &param_name, size_t cvc_index = 0) {
  return cvcs[cvc_index]->get_param(param_name);
}
```

src/colvarcomp.cpp
giacomofiorin: Unfortunately not all the parameters have the same type, so the assignment operator cannot be the same for all values. So several exceptions would have to crawl their way into that loop regardless. The best design is in my opinion the one you used already with `std::map<std::string, std::function>` but this part of the code cannot be conditionally compiled yet.

src/colvar.cpp
giacomofiorin: Indeed, that would be the way to extend it easily. I would like to start with a single component first, make sure this interface does what is needed for Fabrizio's code, and extend it after that.

giacomofiorin:
@HanatoK 1. Yes, the DEER code is in the `rad` branch, and it has been very poorly integrated so far because it was first forked over 2 years ago.  Since it was never well integrated in the first place, I'm opting for adding the glue code to `master` first.

2. The RAD method is implemented as a bias (as it should) but it can also modify parameters used to compute the collective variable (i.e. parameters of the corresponding CVC).  Specifically, it can use different minimization criteria to best-fit those parameters against experimental data.  Depending on the criterion used, the gradient of the variable with respect to the parameters (implemented here) or other related quantities are needed.

https://doi.org/10.1016/j.str.2018.10.013

HanatoK:
@giacomofiorin Thanks for the paper!
Actually I think if C++17 is allowed then we can use std::any, and also a value of parameter can be stored directly in the map without a member variable. For example:
https://gist.github.com/HanatoK/f489f782d6f54326e55d5d7aaa469519

giacomofiorin:
A container based on `std::any`, or in general using more dynamic typing would be very helpful, particularly for code that is less performance-sensitive (feature flags, parameters).  But C++17 is still unsupported by most commercial compilers, including several of those used in major supercomputing architectures:
https://en.cppreference.com/w/cpp/compiler_support#cpp17

src/colvarcomp.h
```diff
@@ -211,6 +209,27 @@ class colvar::cvc
...
+
+  /// Pointer to the gradients of the variable with respect to param_name
+  virtual colvarvalue const *get_param_grad(std::string const &param_name);
+
+  /// Get a copy of the names of registered parameter gradients
+  virtual std::vector<std::string> get_param_grad_names();
+
+  /// Set the named parameter to the given value
+  virtual int set_param(std::string const &param_name, void const *new_value);
+
```
jhenin: How about making `register_params` and associated functions members of `colvarparse`? That could be extremely generic, and I strongly agree that `register_params` can be bundled into `get_keyval`.

Admittedly `get_param_grad` is more CVC-specific, although it would work for a colvar, and possibly for a bias if that means the gradient of the bias energy.

giacomofiorin: `get_param_grad` will be tricky, because its return value should be `colvarvalue *` for the one associated to a variable/CVC (gradient of a general quantity), and `real` for a bias (gradient of an energy), which I didn't add yet.  So these are two entirely different functions, both `virtual`...

To simplify a bit I moved the new functions to a their own class, which is for now inherited simply as `public` by `colvarparse` at the root, but we can play around with that if needed.

giacomofiorin:
Hi, I'm planning to merge this.  Any objections?

jhenin:
No objection.

PR #312, opened 2019-12-12
Ignore a keyword if it has content to the left of it (regardless of braces)

### LINKED ISSUE #311: Parser detects component name keyword inside unrelated customFunction line
The following config:
```
colvar {
    name d
    customFunction distance + 2

    distance {
        name distance
        group1 {
            atomNumbers 1 2
        }
        group2 {
            atomNumbers 3 4
        }
    }
}
```

gives the error:
```
colvars: Error: definition for atom group "group1" not found.
colvars: Error: definition for atom group "group2" not found.
colvars: Error: keyword "+" is not supported, or not recognized in this context.
colvars: Error: in setting up component "distance".
colvars: Error: no valid components were provided for this collective variable.
```

Expected behavior: the distance keyword should only be parsed when at the beginning of a line.

giacomofiorin:
When keyword `A` is used as part of the value of another keyword `B` on the same line and without braces, the parser will try to look for the value of keyword `A` in the same line.

This change fixes #311 by checking that there is no character other than the allowed keyword delimiters between the beginning of the line and the beginning of the keyword.

### DIFF
--- a/src/colvarparse.cpp
+++ b/src/colvarparse.cpp
@@ -33,6 +33,8 @@ namespace {
 
 
 colvarparse::colvarparse()
+  : keyword_delimiters_left("\n"+std::string(white_space)+"}"),
+    keyword_delimiters_right("\n"+std::string(white_space)+"{")
 {
   colvarparse::clear();
 }
@@ -46,6 +48,8 @@ void colvarparse::clear()
 
 
 colvarparse::colvarparse(const std::string& conf)
+  : keyword_delimiters_left("\n"+std::string(white_space)+"}"),
+    keyword_delimiters_right("\n"+std::string(white_space)+"{")
 {
   colvarparse::set_string(conf);
 }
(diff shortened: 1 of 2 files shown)


jhenin:
Taking my test example:
```
colvar {
    distance {
        group1 {
            atomNumbers 1 2
        }
        group2 {
            atomNumbers 3 4
        }
    } BLAH BLAH BLAH name myColvar
}
```

The parser in this branch will not parse the `name` keyword, which is correct. But neither will it complain that `BLAH` is not a valid keyword, which I think it should.

jhenin:
I see some inconsistencies with what happens after a closing brace. The manual states that no keyword can appear after a closing brace on a line. So strictly speaking, anything after a closing brace other than whitespace or a comment should trigger an error. Right?

Currently, within a `colvar {} `block, a keyword such as `name` is accepted. After the closing brace of a colvar block, a global keyword such as `colvarsTrajFrequency` is recognized, whereas the `colvar` keyword is not.

giacomofiorin:
"Strictly" is the operative word here. This syntax is not allowed and is rather uncommon as well: whether an error is raised or the keyword ignored is a matter of choice. We last agreed to leave it.

However, noe with Dashboard templates syntax like this will become more common, because the insertion place of a template string cannot be predicted.  It may be helpful, though, if all multi-line template strings were also terminated by a newline, so that this syntax is not encouraged further.

Meanwhile I'll add more commits to this PR and change its description accordingly.

jhenin:
I am labeling this as an enhancement to indicate that this is a modest deviation from documented behavior: accepting input which should be illegal. The only possible negative impact on user experience is the inconsistency in parsing by the VMD Dashboard and Colvars. However the Dashboard relies on templates that do not contain faulty input, so this could only happen if the user actively corrupts the proposed configuration.

jhenin:
@giacomofiorin Is this a draft in the sense that it only fixes part of the issue (in which case I am tempted to merge it as an incremental improvement) or is it likely to break stuff?

giacomofiorin:
The former.  But test it again, since it's been a year.

jhenin:
This is working correctly when rebased on current master. I propose merging.

PR #313, opened 2019-12-15
[WIP] Port existing colvarscript functions to new interface

jhenin:
Ported all functions to the macro-based framework.

Freshly rebased onto master. Passes regtests, although far from all scripting functions are tested in there.

A number of members of colvarscript and colvarmodule have been made public.
(see big lines of TODO)
This should probably be changed by exposing the relevant functionality through
colvarscript methods.

### DIFF
--- a/src/colvarscript.h
+++ b/src/colvarscript.h
@@ -57,7 +57,7 @@ public:
   int run(int objc, unsigned char *const objv[]);
 
   /// Set the return value of the script command to the given string
-  inline void set_str_result(std::string const &s)
+  inline void set_result_str(std::string const &s)
   {
     result = s;
   }
@@ -259,7 +259,7 @@ extern "C" {
            "Print the help message",
            0, 0,
            {},
-           script->set_str_result(script->help_string());
+           script->set_result_str(script->help_string());
            return COLVARS_OK;
            )
 
@@ -267,7 +267,7 @@ extern "C" {
            "Get the Colvars Module version number",
            0, 0,
            {},
-           script->set_str_result(COLVARS_VERSION);
+           script->set_result_str(COLVARS_VERSION);
            return COLVARS_OK;
            )
 
@@ -299,7 +299,7 @@ extern "C" {
            "Get the module's configuration string read so far",
            0, 0,
            { },
-           script->set_str_result(cvm::main()->get_config());
+           script->set_result_str(cvm::main()->get_config());
            return COLVARS_OK;
            )
 
@@ -337,7 +337,7 @@ extern "C" {
                   ++cvi) {
                 res += (cvi == script->colvars->colvars.begin() ? "" : " ") + (*cvi)->name;
               }
-              script->set_str_result(res);
+              script->set_result_str(res);
               return COLVARS_OK;
             } else if (!strcmp(script->obj_to_str(objv[2]), "biases")) {
               for (std::vector<colvarbias *>::iterator bi = script->colvars->biases.begin();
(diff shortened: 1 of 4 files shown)


jhenin:
The main reason to push this forward was to get to a unified, consistent state before implementing major changes such as the Python interface. And the macro system seemed more flexible.

As you suggested I've isolated the significant changes from there by creating this file:https://github.com/Colvars/colvars/blob/colvarscript_generic/src/colvarscript_commands.h

That file can be pulled into this branch and adapted, and then we can get rid of this branch.

giacomofiorin:
Yup: the purpose of the code currently in `colvarscript-files` is exactly that.

I didn't fully understand how you suggest to proceed: do you want to copy changes from `colvarscript-files` into `colvarscript_generic` and merge the latter into `master` or the other way around?

jhenin:
>  do you want to copy changes from colvarscript-files into colvarscript_generic and merge the latter into master or the other way around?

Changes in colvarscript-files are much more extensive (and unfinished). I suggest grabbing the file colvarscript_commands.h from colvarscript_generic, and then finishing the work in colvarscript-files.

giacomofiorin:
OK that makes more sense. It will create some more conflicts, but hopefully it will be a one-off.

src/colvarscript.h
```diff
@@ -88,6 +98,9 @@ class colvarscript  {
     cv_frame,
     cv_units,
     cv_colvar,
+    // Should the following subcommands be listed in the enum if they are handled by the cv_colvar function?
+    // do we want a mechanism to keep help strings for subcommands, or one big help string for the
```
giacomofiorin: Just noticed this question.  In my view a separate help string for each subcommand is desirable for a more "permanent" implementation (i.e. one that is likely to stick around in a precompiled package for some time).  That would mean also breaking up `proc_colvar`, `proc_bias` and `proc_features` into individual commands.

Also, the implementation of `CVSCRIPT` that I added at this location in the branch `colvarscript-files` would also take care of any unused `enum` entries, because it generates them only if the corresponding scripting command is defined by an instance of `CVSCRIPT`.

src/colvarscript.h
```diff
@@ -254,15 +279,35 @@ extern "C" {
...
            }
-           script->set_str_result("Error parsing configuration string");
+           script->set_error_msg("Error parsing configuration string");
            return COLVARSCRIPT_ERROR;
            )
 
+  CVSCRIPT(cv_configfile,
+           "Read configuration from a file",
+           1, 1,
+           {"conf_file (str) - Path to configuration file"},
```
giacomofiorin: Note that I haven't implemented a handler yet for the fourth argument of `CVSCRIPT`, which would be ignored.

src/colvarscript.h
```diff
@@ -296,13 +437,83 @@ extern "C" {
...
+              int error = script->proxy->set_frame(strtol(script->obj_to_str(objv[2]), NULL, 10));
+              script->set_result_str(cvm::to_str(error == COLVARS_OK ? 0 : -1));
+              return COLVARS_OK;
+            }
+           )
+
+  CVSCRIPT(cv_colvar,
+           "Access colvar-specific commands: ",
+           2, 4,
+           {"name (str) - colvar name"},
```
giacomofiorin: This string should also contain documentation for the optional parameters, which could also exceed 4, e.g.:
`cv colvar xi set feature on|off`

Notice how the commands implemented so far in the `colvarscript-files` branch are simpler (one argument at most).

src/colvarscript_commands.h
```diff
@@ -0,0 +1,402 @@
+// -*- c++ -*-
+
+#ifndef COLVARSCRIPT_COMMANDS_H
```
giacomofiorin: Copying over this file from `colvarscript-files` and adding to it the new implementations of the older scripting commands may work. However, the corresponding commands would have to be deleted from `colvarscript.h` as well: at the moment they seem to be duplicated.

jhenin:
Since this is not slated for merging anymore, I am closing the PR. Some lessons can be learned from the changes in this branch, they have been or will be implemented in other branches.

PR #314, opened 2019-12-16
More flexible gradient scale in Dashboard

jhenin:
with more robust handling of changes in colvar config
and better documentation

### DIFF
--- a/doc/colvars-refman-main.tex
+++ b/doc/colvars-refman-main.tex
@@ -170,7 +170,14 @@ Atomic coordinates can be modified using VMD's ``Mouse/Move'' functions, and the
 
 A dropdown list allows for changing the current unit system if no variables are defined.
 If some variables are already defined, it is recommended to edit the configuration for all of
-them at once (eg. pressing Ctrl-a, then e), checking that all quantities are expressed in the desired set of units, and adding the \texttt{units} keyword (\ref{sec:units}).
+them at once (eg. pressing Ctrl-a, then Ctrl-e), checking that all quantities are expressed in the desired set of units, and adding the
+\texttt{units} keyword to the general parameters, outside of \texttt{colvar \{\}} blocks (\ref{sec:units}).
+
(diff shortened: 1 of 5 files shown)


jhenin:
A more robust implementation would access the width parameter of colvars through the scripting API, possibly based on #310

doc/colvars-refman-main.tex
```diff
@@ -219,6 +233,12 @@ \subsection*{Overview}
 Each atom group is shown in a different color.
 ``Show gradient'' is available for scalar variables only.
 It creates a graphical representation of the atomic gradients of the selected variables, visualizing how the value of the collective variable would vary in response to a change in atomic coordinates.
+Vectors representing the gradient are rescaled as indicated by the radio buttons \emph{Set max. vector norm} and \emph{Set scaling factor}.
+\emph{Set max. vector norm} rescales gradients so that the largest vector component of each colvar's gradient is represented by an arrow of the specified length, in \AA.
```
giacomofiorin: Using Å units here: is that intentional?

giacomofiorin:
I wouldn't want a run-time `width` parameter associated with the CVC, because its role would be combined with the coefficient anyway. The reason why I added it in #310 is to provide the colvar with a sensible *default* value other than 1. There is also a caveat that for the `deer` variable (not merged yet) the width of a vector-valued variable may not be the same for all vector components: one more reason to keep associating `width` with the colvar, and not with the CVC (1 place to test for special cases vs. 2 places to test).

Let's keep your quick hack for a little while to get the colvar's `width`, then add a dedicated scripting method after #313 is merged?

doc/colvars-refman-main.tex
jhenin: Yup, the arrow is just a graphical object in VMD, the only immediate comparison is with atomic coordinates as displayed in VMD, so Angstrom seemed the best unit. I could implement unit conversion there, but there is no issue of portability to other codes, so I didn't think it was worth it.

PR #320, opened 2020-01-30
Flexible restart-reading

### LINKED ISSUE #187: "cv load" should not active colvar value check
Hi, all,

I would like to suggest that "cv load" should not active:
if (after_restart) {
    if (cvm::proxy->simulation_running()) {
      cvm::real const jump2 = dist2(x, x_restart) / (width*width);
      if (jump2 > 0.25) {
        cvm::error("Error: the calculated value of colvar \""+name+
                   "\":\n"+cvm::to_str(x)+"\n differs greatly from the value "
                   "last read from the state file:\n"+cvm::to_str(x_restart)+
                   "\nPossible causes are changes in configuration, "
                   "wrong state file, or how PBC wrapping is handled.\n",
                   INPUT_ERROR);
        return INPUT_ERROR;
      }
}
in colvar.cpp.
Since when one wants to use "cv load", the system's configuration has usually changed.

Thanks,
Haohao

giacomofiorin:
The modified input state reading functions will load each block of the state file (after the global header) depending on the prefixing keyword (`colvar` or one of the biases), ignoring non-matching objects in either the state file or the current configuration. When a match is found, any error in reading the relevant block would trigger a global error.

In the process, I added support for deferred state file load in LAMMPS, which was missing.

Caveat 1: it wasn't as simple as I mentioned in #187 due to a design change dating back to #18 that made all restraints share the same keyword `restraint` in the state file. I think what I added here is backward-compatible, but best to make an extra check.

Caveat 2: any blocks of the state file that are ignored because they have no match in the current configuration could very well be gibberish, for as long as they have matching open-close braces.

Caveat 3: it is unlikely that a state file gets accidentally truncated right in between two blocks, but if that happens the new code won't catch the error (unlike the existing code, which requires a correct restart for everything).

Addresses part of the specific issues raised in #187.  Not addressed is the option to ignore the checks upon restarting by specifying which part of the state file to load.


jhenin:
This looks good! One last case I see that is not relaxed by this is the case where a colvar value conflicts with that in the state file. A likely cause for that is continuing a Colvars run with a new set of atomic coordinates, possibly from an exchange scheme. That may well be the case for #187, which would then not be completely addressed by this.

giacomofiorin:
Thanks for pointing it out: I amended the description of this PR accordingly.

I am not fond of a global flag `cv load true|false`, because it could easily get added to a script and left there accidentally when the script is reused.

I much prefer the ability to load the state individually for an object.  This could be done either with `cv load bias <name> <filename>` or more clearly `cv bias <name> load <filename>`.  Either way this is best done after #316 is done: adding a note there.

jhenin:
I think `cv bias <name> load <filename>` is a very good syntax to provide. That works only for NAMD though, although maybe a equivalent form of `fix modify` could be proposed in LAMMPS. For something like Gromacs, I don't know.

giacomofiorin:
I just checked that in LAMMPS you can do the following with a variable:
```
fix Colvars all colvars ...
variable CV string "fix_modify Colvars"
run 0
${CV} <command>
```
which would simplify a lot the use of `fix_modify` as a script command interface (at least for the text-based backend).

The caveat is the `run` command in between, because as @akohlmey pointed out in #216 the module and the proxy are initialized only when the run begins.  I think that can be changed, though.

As for GROMACS, I assume that its Python API will gain traction soon, and the interface used in #316 will be easier to use in Python than the current one.  The same is true for the LAMMPS Python interface, which is not used more often than it deserves due to the need for an additional step of building LAMMPS as a Python extension module (not difficult per se, but most sysadmins will just build it the executable).

Of course, there is always the option to add Tcl to those codes if you like...

giacomofiorin:
Other discussed items are already in #320. Merging this.

PR #324, opened 2020-02-16
Remove default values for lowerWall and upperWall legacy keywords

giacomofiorin:
As already anticipated in previous versions of the doc (NAMD 2.13, LAMMPS or 12Dec2018 later) these two legacy keywords do not have a default value any more and need to be explicitly provided.

To avoid excessive refactoring of existing input files, I've kept their role as legacy keywords, though.

The default choice of wall == boundary is particularly poor for metadynamics: for this reason, I added a paragraph to the documentation to guide the users towards a sensible choice.


giacomofiorin:
I rebased this onto `master`. Previously I had it for the branch of #316, but that would add unnecessary dependencies. It can be merged as soon as agreed upon.

PR #325, opened 2020-02-19
Allow setting sigma parameter directly for metadynamics

giacomofiorin:
The new option `gaussianSigmas` allows setting the sigmas of the Gaussian explicitly for each colvar in the input.

The older keyword `hillWidth` is still implemented, but it does not have a default value any more (it had been already documented as such, but not changed yet in the code, for about 2 years).


giacomofiorin:
Integrated suggestion from @fabsugar to make `writeHillsTrajectory` compatible with `sum_hills`.

HanatoK:
@giacomofiorin Does that mean we can set sigma values for each CV separately in multidimensional metadynamics?
Also, as you referred to `sum_hills`, is that the `sum_hills` tool in `PLUMED`?

giacomofiorin:
That's right.  Note that up until this change, you could already choose a different `width` for each variable, and different "sigmas" would be computed accordingly.

This change simply makes it more explicit: if you choose `gaussianSigmas` it requires you to specify, for each variable, *both* the grid spacing and the Gaussians' sigmas.  It is also more similar to PLUMED, which would help users who use both implementations.  This also means additional typing, of course.

For that reason I thought to leave open the option to keep using the `width` parameters to define the sigmas: you only need to choose one reasonable value for `hillWidth`, which you would not only share between the variables of a 2-D or 3-D metadynamics run, but also reuse between independent 1D runs.

giacomofiorin:
@fabsugar is happy enough on a quick look, and I'll consider @HanatoK's emoji as other positive feedback. @jhenin if you don't mind, I'll go ahead and merge.

jhenin:
[Thumbs up emoji]

PR #327, opened 2020-03-01
Colvars doc update and question regarding including in VMD/NAMD manual

giacomofiorin:
There is a caveat with the Colvars doc when embedded in the NAMD and VMD manuals. Conditional text (e.g. `\cvscriptonly{...}`) does not play nice with `\input`ing long sections in the LaTeX environment used at UIUC.

Acknowledging that the Colvars manual is a "section" of disproportionate size (longer than one third of either the NAMD and VMD manual), maybe it's easier to stop embedding it and just link it?  Specific versions of the GitHub Pages website for each NAMD or VMD version would be needed, but that's doable.

Alternatively, to stay with the current arrangement, is there a way to `\input` conditionally files into the NAMD or VMD LaTeX file?  

EDIT: I made an attempt to the latter solution.  Merging the first commit (i.e. the useful content?) into master so that it's in sync with the code, but waiting to merge the rest to confirm that everything works as it should with the NAMD and VMD doc build systems.

EDIT 2: Jim mentioned that the doc build host is running CentOS 6, perhaps try building there.

### DIFF
--- a/doc/colvars-refman-lammps.tex
+++ b/doc/colvars-refman-lammps.tex
@@ -29,5 +29,10 @@
 % content that applies only to programs that use atom names (i.e. not LAMMPS)
 \newcommand{\cvnamebasedonly}[1]{}
 
+% File output prefixes
+\newcommand{\outputName}{\emph{output}}
+\newcommand{\restartName}{\emph{restart}}
+\newcommand{\inputName}{\emph{input}}
+
 
 \input{colvars-refman.tex}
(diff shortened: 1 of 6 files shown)


giacomofiorin:
I was able to reproduce the issue previously seen with old versions of `latex2html`. When a file that is not supposed to be included (because it is inside a `\cvxxxxxonly{...}` block), `latex2html` accidentally sill parsed it for content, generating garbage.

I worked around it by naming the file included differently depending on whether the stand-alone manual, the NAMD or VMD user's guides are being built.

I'll merge this, and submit an exact copy of the Colvars chapter for the NAMD UG (without inlining `cvscript-tcl.tex`) at the next opportunity.

Issue #328, opened 2020-03-06
Reflecting boundary conditions for extended Lagrangian


jhenin:
Might bring a definitive solution to issues like #245 and #252

fabsugar:
A similar issue at the boundaries is present in the case of metadynamics. In that case the problem is mostly solved using hills reflected at the boundaries. An even better solution but more cumbersome is to use inversion condition at the boundaries [https://journals.aps.org/pre/pdf/10.1103/PhysRevE.81.055701], the latter adjust not only the free energy value but also its gradient; however reflection seems to be sufficient in most cases.

jhenin:
Thanks Fabio! As you see I've written the basic code. I need to test it with eABF.

jhenin:
Closing this, now redundant with the PR #329 . Discussion can take place there.

HanatoK:
@jhenin Although I remember this feature, I have to check the code for the actual option names. Could you please update the doc about the options accordingly?

jhenin:
Thanks for the reminder!
434fa73a23abe9c8e3cd2cc3fc0751ba15ac61ba

PR #329, opened 2020-03-06
Implement reflecting b.c. for ext Lagrangian

jhenin:
In (quick) response to issue #328 

TODO:
- [ ] regtests
- [ ] documentation

### DIFF
--- a/src/colvar.cpp
+++ b/src/colvar.cpp
@@ -681,6 +681,9 @@ int colvar::init_extended_Lagrangian(std::string const &conf)
       // Adjust Langevin sigma for slow time step if time_step_factor != 1
       ext_sigma = cvm::sqrt(2.0 * cvm::boltzmann() * temp * ext_gamma * ext_mass / (cvm::dt() * cvm::real(time_step_factor)));
     }
+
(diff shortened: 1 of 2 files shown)


HanatoK:
The code looks OK, but I'm not sure if the reflecting boundary can help. The paper you mentioned in the issue exemplifies it by metadynamics, which is a biasing potential method filling the wells. Does it also generate correct gradients for eABF's estimators?

jhenin:
Good question @HanatoK . I've looked at exactly that today, here are the results:

image

Incidentally this shows why I'm not a fan of applying wall potentials to the extended coordinate (the bypassExtendedLagrangian option of #299 ).

HanatoK:
@jhenin It looks good!

fhh2626:
Looks good for me. What's interesting is your results on ext. CV vs real CV. This is quite unexpected.

jhenin:
@fhh2626 Note that when using the bias on the real CV, the same kind of artefact appears on the eABF PMF. So choosing the variable to apply the wall is choosing which estimator is wrong at the boundary.

PR #341, opened 2020-04-08
Detect new grid parameters

### LINKED ISSUE #340: Wrong number of bins in histogram grid if histogramGrid provided
```
colvar {
    name d
    lowerBoundary 0
    upperBoundary 2
(...)
}
histogram {
  colvars d
  histogramGrid {
    lowerBoundaries 0
    upperBoundaries 10
    widths 0.01
  }
}
```
This example gives a grid with just 2 bins, as would be expected from the colvars's parameters.

This only seems to happen if `lowerBoundary` (colvar) == `lowerBoundaries` (histogramGrid). Otherwise the number of bins is correct.
This is due to the following test in `colvargrid::parse_params()`:
```
        if ( (old_nx[i] != nx[i]) ||
             (cvm::sqrt(cv[i]->dist2(old_lb[i],
                                     lower_boundaries[i])) > 1.0E-10) ) {
          new_params = true;
        }
```
which I don't understand.

jhenin:
Fixes #340


src/colvargrid.h
```diff
@@ -860,12 +862,15 @@ template <class T> class colvar_grid : public colvarparse {
...
     bool new_params = false;
     if (old_nx.size()) {
       for (size_t i = 0; i < nd; i++) {
-        if ( (old_nx[i] != nx[i]) ||
-             (cvm::sqrt(cv[i]->dist2(old_lb[i],
-                                     lower_boundaries[i])) > 1.0E-10) ) {
+        if (old_nx[i] != nx[i] ||
+            cvm::sqrt(cv[i]->dist2(old_lb[i], lower_boundaries[i])) > eps ||
+            cvm::sqrt(cv[i]->dist2(old_ub[i], upper_boundaries[i])) > eps ||
+            cvm::fabs(old_w[i] - widths[i]) > eps) {
```
giacomofiorin, line 873: You're right. The previous version of the test will fail in a variety of situations. Good catch.

giacomofiorin:
Let me add a test before merging.

giacomofiorin:
@jhenin There is still a small issue with `run_tests.sh`, which after your change doesn't populate the `AutoDiff` folder with output from the second run, i.e. `test.restart`.  Can you fix it?  For the time being I manually copied the output files in there.

Other smalls problems (all predating this PR, of course):
1. The OpenDX output files aren't being tested, and are written regardless.  Perhaps I'll add a test with three variables, where a .dx file is most useful :-)
2. It may be a good idea to allow disabling `outputFile` and `outputFileDX` by providing an empty value for them. This way one can have fewer files while a simulation is running.

jhenin:
> There is still a small issue with run_tests.sh,

Oops, that's fixed now.

> The OpenDX output files aren't being tested, and are written regardless. Perhaps I'll add a test with three variables, where a .dx file is most useful :-)

In ABF I don't write a dx file in dimension < 3... Maye that could be the default?

> It may be a good idea to allow disabling outputFile and outputFileDX by providing an empty value for them. This way one can have fewer files while a simulation is running.

Agreed.

jhenin:
@giacomofiorin I implemented something along the lines we discussed. Please have a look.

jhenin:
Ok, back to using empty string value as default. I'm proposing to use "none" as a file name to disable output.

PR #344, opened 2020-04-09
Support for symmetry permutations of atoms in RMSD

jhenin:
Reported value is minimal with respect to specified permutations.
Gradients are calculated accordingly.

Use case: DBC coordinate with symmetric ligand.


src/colvarcomp_distances.cpp
```diff
@@ -1045,6 +1045,44 @@ colvar::rmsd::rmsd(std::string const &conf)
...
+  n_permutations = 1;
+
+  while (key_lookup(conf, "atomPermutation", &perm_conf, &pos)) {
+    std::vector<size_t> perm;
+    if (perm_conf.size()) {
+      std::istringstream is(perm_conf);
+      size_t index;
+      while (is >> index) {
+        std::vector<int> const &ids = atoms->ids();
+        size_t const ia = std::find(ids.begin(), ids.end(), index-1) - ids.begin();
```
jhenin, line 1060: @giacomofiorin I have a slight doubt about this: are we sure that the internal atom ids are just the user-facing 1-based numbers minus 1? If not, how do we find one given the other? Should the proxy report if the back-end's internal numbers are 1-based or 0-based, or provide a mapping function? Specifically, I'm thinking of the upcoming Fortran-based back-ends.

src/colvarcomp_distances.cpp
```diff
@@ -1056,6 +1094,20 @@ void colvar::rmsd::calc_value()
...
+  for (size_t ip = 1; ip < n_permutations; ip++) {
+    cvm::real value = 0.0;
+    for (size_t ia = 0; ia < atoms->size(); ia++) {
+      value += ((*atoms)[ia].pos - ref_pos[ref_pos_index++]).norm2();
+    }
+    if (value < x.real_value) {
+      x.real_value = value;
+      best_perm_index = ip;
+    }
+  }
```
giacomofiorin, line 1110: Do you see this as potentially useful for larger selections (e.g. macromolecules instead of small molecules)? If so, it may be a good idea to split this up as different CVC instances, each with a different ordered selection and reference coordinates, and select the minimum for computing the colvar.  It probably can't be done with just `customFunction`, but definitely `scriptedFunction`.  The reason I bring it up is that different CVCs can be computed in parallel.

Nothing to gain of course in performance for small groups.

src/colvarcomp_distances.cpp
giacomofiorin: There is no guarantee, although in all backends this is true at the moment because atom IDs are just array indices (0-based). What I find tricky with this approach is that you could select atoms in a compact way (e.g. with `atomNameResidueRange`) but you would need then to use internal indices anyway to write the permutation. It would seem clearer to me to specify each permutation as its own selection and reference coordinates (see my other comment), unless I'm missing something.

src/colvarcomp_distances.cpp
```diff
@@ -1045,6 +1045,44 @@ colvar::rmsd::rmsd(std::string const &conf)
...
+      }
+      if (perm.size() != atoms->size()) {
+        cvm::error("Error: symmetry permutation in input contains " + cvm::to_str(perm.size()) +
+                  " indices, but group \"atoms\" contains " + cvm::to_str(atoms->size()) + " atoms.");
+        return;
+      }
+      cvm::log("atomPermutation = " + cvm::to_str(perm));
+      n_permutations++;
+      // Record a copy of reference positions in new order
+      for (size_t ia = 0; ia < atoms->size(); ia++) {
```
giacomofiorin, line 1081: Remember about `-fno-for-scope`. We still need to remain compatible with Windows 98, man.

src/colvarcomp_distances.cpp
jhenin: That is the main, idea behind the FRMSD code that inspired this (https://www.ncbi.nlm.nih.gov/pubmed/30828834), but ultimately I don't think that is a significant use case for Colvars, because symmetric atoms in macromolecules are anecdotal: most of them are hydrogen, and the others are end groups like carboxylates. So this is intentionally tailored for small molecules with significant symmetry, typically something like phenol.

Actually this can already be achieved with several RMSD CVCs and a `customFunction` that is their min. However that solution:

- is not available in VMD/Dashboard until Lepton gets included
- has several CVCs with the same fittingGroup doing the same alignment
- needs a long configuration string that is mostly duplicated - and DBC config strings are quite long and complex already
- requires the user to manually provide different sets of reference positions

A straightforward extension of the present implementation, however, is to provide alternate ways to define the permutations in the input. I agree that this one is potentially cumbersome. Only mentioning atoms that actually change in the permutation is a good idea.

src/colvarcomp_distances.cpp
jhenin: For the permutation you need to mention atoms individually, but I agree that it would be nice to skip atoms that are unchanged.

src/colvarcomp_distances.cpp
jhenin: Hey, you removed that flag in your patches! How am I supposed to catch that now???

Seriously though, this builds even with the flag, because the variable is not redefined after the loop. The previous instance is inside a block, not a for loop.

I've added the flag back into my build tree so I can catch the next instance of this.

src/colvarcomp_distances.cpp
jhenin: > A straightforward extension

Edit: it's not that straightforward. I will only implement it if a real user requests it.

src/colvarcomp_distances.cpp
giacomofiorin: Fair enough. However, I've been checking that VMD builds and runs correctly using the pristine version from CVS. I think that's what we all should be doing, especially seeing as 99.99% of VMD users rely on precompiled builds.

I use the patches to enable the new features in the VMD build that I actually *use* myself in production.

PR #345, opened 2020-04-10
Do not accumulate data at step 0

### LINKED ISSUE #343: Stop accumulating histograms and metadynamics at step 0?
Currently, the 0-th step is being counted twice (because it is the same as the last step of the *preceding* job). This adds a tiny error during MD simulations in NAMD or LAMMPS, and no error in a VMD analysis run. So it has been mostly harmless.

But now you have use cases that rely on many short `run` commands (e.g. constant pH in NAMD, replica exchange, etc), so the error isn't necessarily tiny any more.

ABF and TI currently skip the 0-th step of any `run` command in NAMD, although for different reasons (the total forces aren't available yet).  I need to test with LAMMPS, because there are no ABF inputs in the cross-engine `tests` folder (and not many in `namd/tests/library` anyway...).

giacomofiorin:
A new flag `zeroStepData` is introduced (default: off) to control whether data is recorded at step 0 (see #343).  For simplicity, this check is applied at the beginning of every run, regardless of whether it's a continuation of a previous one.  Calls to functions that increment binned data are now wrapped with `if (cvm::step_relative() > 0 || is_enabled(f_cvb_zero_step_data)) { ... }`.  Regtests are updated accordingly.

This PR closes #343.

@whitead For ALB it wasn't 100% clear to me whether it is appropriate to make this change. If you think so, either post a commit or simply comment here. Otherwise, the code will be left untouched.


whitead:
@giacomofiorin thanks for looking at this. This will not affect ALB. It is split into equilibration/accumulation steps. At step 0 it is in equilibration so nothing is accumulated until `update_freq / 2` . It will mean the first amount of time being equilibrated is one frame less, but that won't affect any statistics.

giacomofiorin:
This looks ready to merge for me. All test inputs leave `zeroStepData` to its default value of `no`, which is the correct choice in most cases (MD run). Should one want to use `zeroStepData yes`, there is an error thrown when ABF can't get the total forces immediately (i.e. NAMD). However, this doesn't cover the case of eABF.

@jhenin @HanatoK @fhh2626  Can you comment on this?

jhenin:
Good to me with these changes.

giacomofiorin:
@jhenin Why re-generate all the test outputs including trajectory files when only the name of the keyword is changed?

jhenin:
@giacomofiorin Good point, that was very heavy-handed. I've reduced the update to the log files.

PR #346, opened 2020-04-15
Load/save state information from/to strings for the module or for individual biases

giacomofiorin:
Four new scripting commands are added.  Formatted I/O is used: this mechanism won't be as fast as binary I/O, but it will remove the need for temporary files in a variety of situations (and is probably easier to debug in scripts).

The corresponding tests use an ABF instance as a bias with a non-trivial state, so that the output of the tests is not sensitive to the changes in #345.  However, the mechanism is valid for all types of bias (metadynamics, histograms, etc.).


jhenin:
@giacomofiorin This looks good to me!

PR #349, opened 2020-05-01
Add front-end for volumetric map-based collective variables

giacomofiorin:
This PR adds the code needed to use volumetric map-based collective variables, used in:
Fiorin G, Marinelli F, Faraldo-Gómez JD.
_“Direct derivation of free energies of membrane deformation and other solvent density variations from
enhanced sampling molecular dynamics.”_
J. Comp. Chem. **41**(5):449-459 (2020) https://doi.org/10.1002/jcc.26075

The bulk of the implementation (back-end) is in a change submitted directly onto the NAMD repository:
https://charm.cs.illinois.edu/gerrit/c/namd/+/5392
This PR is open for discussing Colvars-specific issues and will be merged at some point after the NAMD change is merged.

EDIT: I amended the test input `006_multiple_maps/gridforcemaps.tcl` so that the order of the maps is consistent with the comments in it. (I've been switching them around several times to verify the correct behavior and forgot to put them in the original order.) Sadly, this required also updating the regtest reference trajectories, because `GridForces` uses single-precision floating point numbers for the potential :-(


giacomofiorin:
Merging this based on private chat with @jhenin and the fact that the NAMD counterpart was just merged on Gerrit.

One issue brought up was the ability to patch an earlier version of NAMD (e.g. 2.13) without compile errors. This is subject to the ability to detect the specific minor version at compile time: https://charm.cs.illinois.edu/gerrit/c/namd/+/3926
If that NAMD change does not get merged in a reasonable time, I will provide an alternative preprocessor define to allow as an option building this snapshot against versions of NAMD without the matching `GlobalMaster` code.

PR #350, opened 2020-05-07
Conditionally compile volumetric map support based on NAMD version

giacomofiorin:
Uses internally the macro `NAMD_VERSION_NUMBER` defined in NAMD Gerrit change https://charm.cs.illinois.edu/gerrit/c/namd/+/3926 (but not yet merged) with fall-back to a hard-coded value corresponding to 2.14b1.

To build the current Colvars snapshot against older NAMD snapshots (e.g. 2.13) one has to set explicitly the number version at compile time. Although the numbering scheme is not too obvious, `-DNAMD_VERSION_NUMBER=0` will suffice.

Submitting as draft and passing over to @jhenin for fine tuning.

### DIFF
--- a/namd/src/colvarproxy_namd.C
+++ b/namd/src/colvarproxy_namd.C
@@ -243,11 +243,13 @@ int colvarproxy_namd::setup()
     atom_groups_new_colvar_forces[ig] = cvm::rvector(0.0, 0.0, 0.0);
   }
 
+#if NAMD_VERSION_NUMBER >= 34471681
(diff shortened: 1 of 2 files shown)


jhenin:
Current state of the patch https://charm.cs.illinois.edu/gerrit/c/namd/+/3926 is MERGE_CONFLICT... Not there yet.

giacomofiorin:
You may want to ping them... Unfortunately if they decided not to go ahead with that change while in the process of handling two major versions for the near future (2.x and 3.x) they would probably have a reason not to.

PR #357, opened 2020-06-25
Skip the zero-step also when multiple run commands are executed in sequence

giacomofiorin:
This PR adds logic to detect whether the current step is a "continuation" one, meaning the 0-th step of a `run` command is being executed and the `run` command is not the first in the sequence (a condition that would be already covered by `step_relative() == 0`).

Implements #343 properly (see comments). In #345 I was just too busy "WFH" to bother checking that the previous logic also covered the use of multiple `run` commands (intended use case). This is now covered by a dedicated test.

Due to the subtleties of ABF and TI, the change in logic is not applied to them.


giacomofiorin:
I made the test a bit more sophisticated by incorporating a metadynamics instance and making one of the `run` commands run for zero steps. @jhenin Okay to merge in your opinion?

PR #361, opened 2020-09-16
Format code examples with colored background

giacomofiorin:
There are very few examples in the Colvars guide, thus it is easy to miss them especially in the HTML version: Tex4ht ignores the `\ttfamily` directive and just formats everything the same (text and code).  This ends up being not only harmful to the copy-pasters (not an issue in my view), but also to those who are genuinely reading the doc.

This PR introduces two new LaTeX environments, `mdexampleinput` for examples of input of the hosting engine, and `cvexampleinput` for Colvars configuration examples. They are colored differently (orange and purple, respectively) to more quickly differentiate the two.

Both environments also contain explicit `<pre>` tags when the HTML version is compiled, to prevent Tex4ht from trying to be smart about the fonts and ending up doing the dumb thing. 

All the examples that I could find on a quick look (again, **we know** that there aren't many!) are now formatted with the new environments.  I purposefully left alone the scripting API examples (one section that actually contains a high density of examples), because I'm working on that section in another branch.

### DIFF
--- a/doc/colvars-refman-css.tex
+++ b/doc/colvars-refman-css.tex
@@ -6,6 +6,9 @@
 % Colvars repository at GitHub.
 
 \Css{
+.heading {
+    background-color: rgb(215, 215, 215);
+}
 @media only screen and (max-width: 959px) {
 .inner {
     max-width: 100\%;
--- a/doc/colvars-refman-main.tex
+++ b/doc/colvars-refman-main.tex
@@ -821,11 +821,11 @@ A collective variable is defined by the keyword \texttt{colvar} followed by its 
 
 \begin{cvexampleinput}
 \-colvar \{\\
-\-~~name xi\\
-\-~~$<$other options$>$\\
-\-~~function\_name \{\\
+\-~~name~xi\\
+\-~~$<$other~options$>$\\
+\-~~function\_name~\{\\
 \-~~~~$<$parameters$>$\\
-\-~~~~$<$atom selection$>$\\
+\-~~~~$<$atom~selection$>$\\
 \-~~\}\\
 \}
 \end{cvexampleinput}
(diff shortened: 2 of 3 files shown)


giacomofiorin:
Note: the HTML version builds correctly (used the LaTeX distribution from Fedora 32) and displays well on Firefox and Chromium/Chrome (which should include Edge, too). Don't have access to Safari at the moment.

I'll leave open for a bit for suggestions regarding the visual appearance and then merge if none come out.

akohlmey:
> Note: the HTML version builds correctly (used the LaTeX distribution from Fedora 32) and displays well on Firefox and Chromium/Chrome (which should include Edge, too). Don't have access to Safari at the moment.
> 
> I'll leave open for a bit for suggestions regarding the visual appearance and then merge if none come out.

FYI, we are testing browser compatibility for LAMMPS docs with https://www.browserstack.com/

jhenin (commented):
This works really well! There are a lot of shorter snippets that are not yet formatted the same way, is that intentional?

giacomofiorin:
@akohlmey Thanks for the suggestion, I signed up and it looks pretty neat! I wonder if the macOS browsers are running on actual Apple hardware given the legal requirements :-)

@jhenin Err, yes, "intentional" as in "I didn't get around to it yet". 

My idea was to merge the feature into `master` first and gradually apply it to the examples that I haven't included yet. I so far looked only for things that had non-breaking spaces (`~` in LaTeX) which we used for the verbatim-like segments so far. The other snippets will require replacing spaces with `~`.

I'm working on the scripting interface section on a branch, but any segments that you see please feel free to just adapt it and just push. I've been using `\paragraph*` to label some of them but not all need it.

akohlmey:
> @akohlmey Thanks for the suggestion, I signed up and it looks pretty neat! I wonder if the macOS browsers are running on actual Apple hardware given the legal requirements :-)

i think so. this all looks like some variant on using remote desktops, e.g. via VNC or a similar protocol with a javascript/webassembly client. they even have ios and android phones to try webpages out on... ;-)

PR #366, opened 2020-10-02
Vmd volmaps interface

jhenin:
Creating a PR to centralize discussion, instead of individual commits.

With today's update we're close to a good support. One point on which I'm undecided is how to handle selected colvars in the list, as opposed to the one under the mouse cursor when right-clicking.

On a related note I still see 0 gradients for mapTotal variables, even though the features are nominally there.


jhenin:
> When I select multiple colvars, each with a different map, the "Show volmap" button from the context menu only acts on the colvar that is being right-clicked, not all the selected ones.

I have code for this, but I'm unsure what to do if the colvar under the mouse cursor is different from the selection. Should it replace it, be appended to it?

> I don't recall quickly VMD extensions that use context menus. If this is the first, it may be a good idea to check with John about possible issues on some Tk platforms.

I've seen them in other plugins. And really they don't use any feature that isn't used in many other places. It's just a regular menu widget and a binding to a mouse button.

jhenin:
A related issue that cropped up: when saving the colvars to a file, the indexFile parameter is not saved, which can cause a failure to read the file in another Colvars session. That should probably be added, like the units line. Is there a robust way to get the relevant index file(s) at run time?

giacomofiorin:
There are a couple of things that I still want to finalize on this branch besides the Dashboard support, so I converted it to draft

These are mostly technical but user-facing. E.g. NAMD GridForces uses both a "scale" column and a "charge" column, the product of which makes up the atomic weights. I used here only a single column (it's silly to me to do otherwise), which can probably be the product of the two columns using `vecmul` in Tcl.

I haven't looked yet at the reason why gradients aren't exposed, but they are computed and stored. Most likely I forgot to call a small piece of the needed plumbing code.

giacomofiorin:
For the indexFile issue, the cleanest would be allowing the general (module-level) options to being also kept as strings while parsing. But that would take more work as previously discussed, and it is probably good to keep a list of the file names anyway. Hence ed5fab36

giacomofiorin:
I'm not fully satisfied with putting the volmaps show/hide buttons under "advanced settings", since they more operations than settings. 

I would rather have them in a togglable sub-panel (see the latest commit).  If needed in the future, this and other togglable sub-panels could still become again separate windows.  What do you think?

Side note: for people who are working on their laptop, putting a "delete" command in the contextual menu is dangerous.  I would either remove it, or look into having a confirmation pop-up window.

giacomofiorin:
@jhenin Here is why the gradients were zero in VMD: quite simply, I had not added yet the relevant code in the backend function because I thought that the gradients wouldn't be needed in VMD... :-)) Obviously that was an oversight, given the capabilities of the Dashboard.

I figured out how to compute the volmap's gradient from the VMD code that is used to compute isosurfaces, and almost completely done implementing the call. (Need to figure out how it is defined, because although the direction looks right, the sign was the opposite and I am worried about the norm of the gradient vectors).

Below is how the gradients look like for the phosphate groups of lipid bilayer leaflet. I'm finding the gradient display quite useful especially when going over the individual maps in a Multi-Map configuration. But this also means clicking multiple times on "Show atoms", "Show gradients" and "Show volmaps" (and the "Hide" counterparts), so I would much prefer to keep them close together. 

What are your thoughts about a layout where atoms, gradients, volmaps and rotations show/hide buttons (and may other displays to be added later) can be together without too much clutter?
image

giacomofiorin:
The gradients in VMD are implemented. As usual, implementing a robust testing script took ~10 times longer than implementing the feature itself... I had forgotten that `volmap` treats the "radius" as a diameter.

jhenin:
> What are your thoughts about a layout where atoms, gradients, volmaps and rotations show/hide buttons (and may other displays to be added later) can be together without too much clutter?

In more than 99% of cases there will be neither volmaps nor rotations defined. Using the same testing logic I used for the context menus, the Volmaps / Rotation panels could appear magically when relevant colvars exist, and become active when they are selected?

jhenin:
Also, nice job making the window adapt to the newly shown entries, I wanted to do that but couldn't find a way. But it breaks when I try to show both Volmaps options and advanced settings at the same time.

The advanced settings could be a different tab instead.

giacomofiorin:
> In more than 99% of cases there will be neither volmaps nor rotations defined. That's (most of the time) when I want those options to stay out of the way. Using the same testing logic I used for the context menus, the Volmaps / Rotation panels could appear magically when relevant colvars exist, and become active when they are selected?

Sold!

giacomofiorin:
> Also, nice job making the window adapt to the newly shown entries, I wanted to do that but couldn't find a way. But it breaks when I try to show both Volmaps options and advanced settings at the same time.
> 
> The advanced settings could be a different tab instead.

Adapting similar Tk code that you already put in is as far as I'm gonna go... :-)  

I'm aware of the issue (just forgot to mention it). If the show/hide volmaps buttons appear automatically, there is no need a sub-panel for the volmap stuff at all and it can be consolidated with the advanced settings. 

Also agree with advanced settings being its own thing (not sure what you mean by "tab": is that a separate window?). The choice of material could apply to all subsequent graphic primitives (atoms/gradients/volmaps/rotations). Also, I'd like to make the arrows a bit thicker, since in the figure above it is a bit hard to see which way they point.

giacomofiorin:
Also about the display of atoms and gradients: would you be open to intersecting them with another selection (that would be equal to `"(all)"` by default) to show fewer of the atoms and gradients? This way it would be possible to understand more clearly how certain pieces of the system move when there are lots of atoms involved.

jhenin:
> not sure what you mean by "tab"

Real tabs (like browser tabs) as used eg. in the Qwikmd plugin.

> would you be open to intersecting them with another selection (that would be equal to "(all)" by default) to show fewer of the atoms and gradients? 

Sounds good. That would go into the gradients section of the advanced settings.

giacomofiorin:
@jhenin I just implemented the two changes that I was suggesting, and with those I'm pretty much done with changes to the Dashboard. 

I didn't try making the show/hide volmaps buttons appear automatically, since you may want to fine-tune the buttons' layout. If not, let me know and I'll just copy and adapt your code again. I assume that the volmap stuff could be automatically shown/hidden in the Settings tab, too.

For the material, take a look and see if it would make sense to allow choosing a different material for each property (atoms, gradients, volmaps, rotations) or keep as is. I am fine either way.

I don't plan on making further changes to the C++ code: once you get to review it, we can merge in my opinion.

image

giacomofiorin:
P.S. Rebased onto `master` and edited out a "WIP" in a commit message that was finalized by a separate commit.

jhenin:
Beautiful picture btw!

With latest adjustments, I think this is almost good to merge... it's just missing a couple lines in the documentation of the Dashboard to say this feature is there.

jhenin:
@giacomofiorin I've gone ahead and split the main menu into two tabs, please tell me what you think.

jhenin:
Tested the PBC wrapping in VMD, that works well, it's particularly nice to have the gradient display when atoms straddle two periodic cells.

PR #370, opened 2020-10-21
Enforce and document XYZ file format

### LINKED ISSUE #369: Error checking in XYZ reader
A LAMMPS user recently reported a problem in a simulation where the reference positions generated a full-flip rotation, leading to gimbal lock for `spinAngle`. The problem originated in `refPositionsFile`, the contents of which were:
```
x1 y1 z1
...
xN yN zN
```
rather than:
```
N
Comment line
C x1 y1 z1
...
C xN yN zN
```
The first version can pass for the second, because numbers are read as strings into the atom names, but never used. The last reference positions then remain uninitialized.

The XYZ reader itself could probably just check if there are more atoms than entries available from the file, and the documentation describe explicitly what is the expected format.

Additionally, we could also enforce that the number of atoms in the XYZ file must match either the size of the group, or the size of the system. But that may break some people's inputs. Related issue #161

giacomofiorin:
Raise error when number of positions in XYZ file doesn't match number of atoms.  Especially useful for LAMMPS users, who have no option for reading PDB files.

Document what is the expected XYZ file format (in case people are not using VMD to prepare it).
    
Fixes #369

### DIFF
--- a/doc/colvars-refman-main.tex
+++ b/doc/colvars-refman-main.tex
@@ -878,7 +878,84 @@ The values of the atomic coordinates and other fields, such as occupancy or temp
 PDB files may be used either as one of the available mechanisms to select atoms (see the \refkey{atomsFile}{atom-group|atomsFile} keyword), or more frequently to read reference coordinates for least-squares fit alignment (see the \refkey{atomsFile}{atom-group|refPositionsFile} keyword).
 }
 
-% TODO Move multicolumn format here?
+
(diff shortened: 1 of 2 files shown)


jhenin:
I like the doc section about file formats!

giacomofiorin:
> I like the doc section about file formats!

Great, what do you think about the multicol file section? The text that @fabsugar added to document the EBMeta feature can probably fit in there as well. There is also the "OpenDX" format, although it is not used in input so far.

PR #371, opened 2020-10-21
GROMACS documentation fixes

giacomofiorin:
`\cvnamebasedonly` selects features where atoms are selected by name and residue, or PDB files are loaded.  Neither is supported in GROMACS at the moment.

I simply disabled the macro, but for more flexible behavior it may be appropriate to break up these two features, and set them by code, e.g.:
```
\cvnamdonly{
\def\cvreadpdb{1}
\def\cvatomnamesel{1}
}
\cvgromacsonly{
\def\cvreadpdb{0}
\def\cvatomnamesel{0}
}
```
and perhaps leave only those `\cv*only{...}` macros that identify the particular backend.

### DIFF
--- a/doc/colvars-refman-gromacs.tex
+++ b/doc/colvars-refman-gromacs.tex
@@ -31,7 +31,7 @@
 \newcommand{\cvleptononly}[1]{#1}
 
 % content that applies only to programs that use atom names (i.e. not LAMMPS)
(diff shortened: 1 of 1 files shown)


jhenin (approved):
I agree with the idea of splitting up this condition, but that can wait until it becomes necessary, whenever a back-end supports one and not the other. Let's not fix it until it's broken, there is enough on our plate as it is.

PR #374, opened 2020-10-27
Write NDX index files within the Dashboard

giacomofiorin:
Allows for more readable Colvars config files.

So far implemented is the Tcl function, passing to @jhenin for thoughts about whether it would be appropriate to call it from the GUI.

### DIFF
--- a/src/colvarscript_commands.h
+++ b/src/colvarscript_commands.h
@@ -263,6 +263,20 @@ CVSCRIPT(cv_listindexfiles,
          return COLVARS_OK;
          )
 
+CVSCRIPT(cv_listindexgroups,
+         "Get a list of the names of index groups loaded in this session",
+         0, 0,
+         "",
+         int const n_groups = script->module()->index_group_names.size();
+         std::string result;
+         for (int i = 0; i < n_files; i++) {
+           if (i > 0) result.append(1, ' ');
+           result.append(script->module()->index_group_names[i]);
+         }
+         script->set_result_str(result);
+         return COLVARS_OK;
+         )
+
 CVSCRIPT(cv_load,
          "Load data from a state file into all matching colvars and biases",
          1, 1,
--- a/vmd/cv_dashboard/cv_dashboard_editor.tcl
+++ b/vmd/cv_dashboard/cv_dashboard_editor.tcl
@@ -432,3 +432,35 @@ proc ::cv_dashboard::editor_help {} {
   help_window $w.editor "Help on colvars config editor" "Colvars Dashboard: the editor window" \
 {Help text}
 }
+
+
+proc ::cv_dashboard::write_index_group { ndxfile sel name } {
+    if { [catch [list tell "${ndxfile}"]] } {
+        # ${ndxfile} is the filename
+        set output [open "${ndxfile}" "a"]
+    } else {
+        # ${ndxfile} is the channel
+        set output ${ndxfile}
+    }
+
+    puts ${output} [format "\[ %s \]" ${name}]
+    set line_buf 0
+    set atom_count 0
+    foreach num [${sel} get serial] {
+        incr atom_count
+        puts -nonewline ${output} [format " %9d" ${num}]
+        set line_buf [expr ${line_buf} + 10]
+        if { ${line_buf} > 70 } {
+            set line_buf 0
+            puts -nonewline ${output} "\n"
+        }
+    }
+    if { ${line_buf} > 0 } {
+        puts -nonewline ${output} "\n"
+    }
+    puts -nonewline ${output} "\n"
+
+    if { "${output}" != "${ndxfile}" } {
+        close ${output}
+    }
+}


jhenin:
I see how that could help with large selections. I see two possible ways to integrate it into the UI:

1. in the editor, selecting an `atomNumbers` line and providing a button "make index file"
2. in the main window, offering to turn all instances of `atomNumbers` to index groups and write a single index file for the whole config.

A limitation I can see is when there is already an index file defined, then you may want to extend it instead of creating a new one. That would require parsing the file, or requesting the names and members of existing index groups from Colvars through the scripting API, so they can be added to the new file. In any case, there is a non-trivial logic there, so we'd have to think about the workflows that real users will adopt.

giacomofiorin:
> I see how that could help with large selections. I see two possible ways to integrate it into the UI:
> 
>     1. in the editor, selecting an `atomNumbers` line and providing a button "make index file"
> 
>     2. in the main window, offering to turn all instances of `atomNumbers` to index groups and write a single index file for the whole config.

I'm not too sure about the second, because the selections should ideally have names that are easy to remember and reuse. It would take a lot of work for an automatic conversion to detect duplicates.

> A limitation I can see is when there is already an index file defined, then you may want to extend it instead of creating a new one. That would require parsing the file, or requesting the names and members of existing index groups from Colvars through the scripting API, so they can be added to the new file. In any case, there is a non-trivial logic there, so we'd have to think about the workflows that real users will adopt.

The Tcl function that I added is already configured to append to an existing index file, and I just added a script function to list the groups' names, complementing the existing one that lists the index files (8e0cec91).

The groups' names must be unique, irrespective of having one index file or multiple ones: so it probably doesn't matter where the newly created groups would be saved. The first index file on the list of those loaded (prompting to create one if there isn't any) should be good enough.

So in the config editor, you could probably have a drop-down menu labeled e.g. "Atoms from index group" that lists them and allows you to select one. If the list changes while the config editor window is open, would Tk allow you to refresh the contents of the menu?

As for creating index groups from VMD selections, that could probably be a button in the main window, since this is a task that shouldn't require creating or modifying a variable. I'd go as far as saying that this would be a useful feature on its own, without any variables or biases ever being created. (Thus probably useful enough to warrant the extra work?)

After a new index group is created, I think it should be fine to write it to an index file, call `resetindexgroups` and then re-read all the index files. If calling `cv config "indexFile $f"` for this last step is not appropriate, I am fine with adding a script function that calls `cvm::read_index_file()` directly.

jhenin:
> The Tcl function that I added is already configured to append to an existing index file, and I just added a script function to list the groups' names, complementing the existing one that lists the index files ([8e0cec9](https://github.com/Colvars/colvars/commit/8e0cec9172bda48d7d10f5a5836a4f7745efccf3)).
> 
> The groups' names must be unique, irrespective of having one index file or multiple ones: so it probably doesn't matter where the newly created groups would be saved. The first index file on the list of those loaded (prompting to create one if there isn't any) should be good enough.
> 
> So in the config editor, you could probably have a drop-down menu labeled e.g. "Atoms from index group" that lists them and allows you to select one. If the list changes while the config editor window is open, would Tk allow you to refresh the contents of the menu?

Only by polling the Module at regular intervals, which seems a little overkill. Is the list likely to change from outside the Dashboard? Because any change that happens within the Dashboard could easily trigger a UI update.

> As for creating index groups from VMD selections, that could probably be a button in the main window, since this is a task that shouldn't require creating or modifying a variable. I'd go as far as saying that this would be a useful feature on its own, without any variables or biases ever being created. (Thus probably useful enough to warrant the extra work?)

It doesn't require creating or modifying a variable, but the context where it's useful is that, defining variables. I think it's not integrated into the workflow of the main window. So the best place in my view is still the editor window.

> After a new index group is created, I think it should be fine to write it to an index file, call `resetindexgroups` and then re-read all the index files. If calling `cv config "indexFile $f"` for this last step is not appropriate,

I think that will work fine. If we refresh the loaded index files, then we should also refresh all the colvar definitions accordingly, so we need to delete every variable and parse the whole config again anyway.

giacomofiorin:
> > So in the config editor, you could probably have a drop-down menu labeled e.g. "Atoms from index group" that lists them and allows you to select one. If the list changes while the config editor window is open, would Tk allow you to refresh the contents of the menu?
> 
> Only by polling the Module at regular intervals, which seems a little overkill. Is the list likely to change from outside the Dashboard? Because any change that happens within the Dashboard could easily trigger a UI update.

Fine with me. The worst that could happen is loading a new file from a script command, outside the Dashboard. But in this case, having to do a manual refresh is not a big deal. I myself would definitely keep an editor window open while also loading new configuration elsewhere. My concern was mostly about the possibility to update the UI at all, which you confirm is there.
 
> It doesn't require creating or modifying a variable, but the context where it's useful is that, defining variables. I think it's not integrated into the workflow of the main window. So the best place in my view is still the editor window.

If there is no requirement to define a dummy variable just to be able to define the group, the config editor is good. My 

> I think that will work fine. If we refresh the loaded index files, then we should also refresh all the colvar definitions accordingly, so we need to delete every variable and parse the whole config again anyway.

Sounds good!

jhenin:
>  My concern was mostly about the possibility to update the UI at all, which you confirm is there.

There is a good example in `::cv_dashboard::update_mol_list`: by tracing a variable graciously provided by VMD it detects any molecule being created or deleted and updates the drop-down accordingly.

We don't have this kind of callback mechanism for changes in index group definitions though,  and I don't think it's worth creating one, although the general idea of supporting callbacks in the Colvars scripting interface is interesting.

jhenin:
Made redundant by #428

PR #377, opened 2020-11-10
Fix bug when updating gromacs colvars proxy version

HubLot:
A quick fix for gromacs update part


HubLot:
@jhenin  realized they were still related version of `gromacs_proxy_version.h` in gromacs-XXX folders.
I removed those since now, we use the one in `gromacs/src/`.

PR #379, opened 2020-11-13
Prevent reading 'vanilla' .cpt files by a Gromacs patched colvars

### LINKED ISSUE #378: [Gromacs Integration] Reading a "vanilla" .cpt from a Gromacs patched with colvars crash
Hi,

When trying to read a .cpt file created with a standard Gromacs, it crashes when we used a Gromacs patched with colvars:

```
$ /gromacs/patched/with/colvars/gmx dump -cp vanilla.cpt
[...]
Command line:
  gmx_d dump -cp vanilla.cpt

GROMACS version = 2020.3
GROMACS build time UNUSED =
GROMACS build user UNUSED =
GROMACS build host UNUSED =

[...]

checkpoint file version = 21

[...]

state flags = 478
ekin data flags = 0
energy history flags = 255
df history flags = 0
ED data sets = 0
swap = 0
AWH history flags = 0
pull history flags = 0
colvars atoms = 9

-------------------------------------------------------
Program:     gmx dump, version 2020.4-MODIFIED
Source file: src/gromacs/fileio/checkpoint.cpp (line 703)

Fatal error:
Count mismatch for state entry box, code count is 9, file count is 1

For more information and tips for troubleshooting, please check the GROMACS
website at http://www.gromacs.org/Documentation/Errors
-------------------------------------------------------
```
Here, the Gromacs patched version try to read the "colvars atoms" line in the cpt. But this line doesn't exist in this .cpt because it has been created with a standard Gromacs version.

The bug was spotted by @flblanc.

HubLot:
This PR aims to mitigate and close #378.

For 2018 versions, I just incremented the `checkpoint file version` to prevent gromacs-2018 to read gromacs-colvars-2018 .cpt files.

For 2020 versions, I used  a variable and their new enum structure. This variable auto-increment itself. This will prevent gromacs-2020 to read gromacs-colvars-2020 .cpt files and so on (gromacs-2021 to read gromacs-colvars-2021 .cpt file, etc).

For cases where someone want to read a 'old' gromacs-colvars.cpt file with a new Gromacs (for example, a gromacs-2020 to read a gromacs-colvars-2018 .cpt file), the code will crash. We need to write something like this somewhere in the depot: 

---------------------------------
## Known Issues with Gromacs

1. Gromacs-colvars version can't read checkpoint files (.cpt) created by a regular Gromacs version. This is due to the colvars data added to the .cpt file which cannot be read by a regular Gromacs version (see Issue #378). Depending on how you tried to achieve that, you will encounter either a compatibility error :

```
-------------------------------------------------------
Program:     gmx <...>, version XXXX.X
Source file: src/gromacs/fileio/checkpoint.cpp (line 935)

Fatal error:
Attempting to read a checkpoint file of version YY with code of version YY
```

Or a crash of the gmx program:
```
-------------------------------------------------------
Program:     gmx <...>, version XXXX.X
Source file: src/gromacs/fileio/checkpoint.cpp (line 703)

Fatal error:
Count mismatch for state entry box, code count is 9, file count is 1
```

-------------------------

- [ ] Before merging, the above statement will have to be committed.


giacomofiorin:
Thanks for the clear explanation!

To avoid the chance of future overlaps, would it be appropriate to increment `cpt_version` by a very large number, e.g. the integer representation of the Colvars version?  For example, using the current Colvars version in `master`, `cpt_version = 17 + 20201109` for 2018 and `cpt_version = cptv_count -1 + 20201109` for 2020?  I am a bit concerned that `cptv_Colvars` may overlap with a new feature in the main GROMACS tree.

jhenin:
Yes, this is not robust to major changes in Gromacs, however it is sufficient to differentiate between plain and Colvars-enabled Gromacs within a major version. The idea of a large increment is interesting. I'll note that the basic design of cpt files is not meant for portability between major versions, so it will misbehave in various ways, Colvars or no Colvars, so I'm also okay with a change that's in the spirit of the current code.

giacomofiorin:
Yes, pretty much everyone I know only uses checkpoint files within the same major version, and the presence or absence of Colvars constitutes a change in major version in itself.

I put this recommendation in the doc (and also formatted the Gromacs code example in the same way that they are for the other engines).

HubLot:
> To avoid the chance of future overlaps, would it be appropriate to increment `cpt_version` by a very large number, e.g. the integer representation of the Colvars version? For example, using the current Colvars version in `master`, `cpt_version = 17 + 20201109` for 2018 and `cpt_version = cptv_count -1 + 20201109` for 2020? I am a bit concerned that `cptv_Colvars` may overlap with a new feature in the main GROMACS tree.

Yes, it's a good solution. This will prevent the new versions of Gromacs from reading old gromacs-colvars .cpt files.
I'll do some tests and commit

HubLot:
So, I added a value of '1000' for cpt file on Gromacs-colvars version. 
A readable round number will make easy to keep track of regular cpt file version.

PR #384, opened 2020-11-25
Use a t_atoms struct for holding atoms information in proxy

HubLot:
This patch aims only the Gromacs 2018.X version.

It replaces the t_mdatoms one which was prone to bug when using large
number of MPI threads/ranks.
It now uses the same syntax as the 2020.X version.

I also removed 2 unused parameters of the `calculateForces()` function.

### DIFF
--- a/gromacs/gromacs-2018.3.patch
+++ b/gromacs/gromacs-2018.3.patch
@@ -266,7 +266,7 @@ diff -aur src/gromacs/mdlib/sim_util.cpp src_colvars/gromacs/mdlib/sim_util.cpp
 +        set_pbc(&pbc, inputrec->ePBC, box);
 +        inputrec->colvars_proxy->update_data(cr, step, pbc, box, bNS);
 +
-+        inputrec->colvars_proxy->calculateForces(cr, mdatoms, box, t, x, forceWithVirial);
++        inputrec->colvars_proxy->calculateForces(cr, box, x, forceWithVirial);
 +    }
 +
      if (inputrec->bPull && pull_have_potential(inputrec->pull_work))
@@ -396,7 +396,7 @@ diff -aur src/programs/mdrun/md.cpp src_colvars/programs/mdrun/md.cpp
 +        }
 +
 +        ir->colvars_proxy =  new colvarproxy_gromacs();
-+        ir->colvars_proxy->init(ir,ir->init_step,mdatoms, observablesHistory, prefix, filenames,filename_restart, cr, MASTER(cr) ? as_rvec_array(state->x.data()) : nullptr);
++        ir->colvars_proxy->init(ir,ir->init_step,top_global, observablesHistory, prefix, filenames,filename_restart, cr, MASTER(cr) ? as_rvec_array(state->x.data()) : nullptr);
 +    }
 +    else
 +    {
--- a/gromacs/gromacs-2018.3/colvarproxy_gromacs.cpp
+++ b/gromacs/gromacs-2018.3/colvarproxy_gromacs.cpp
@@ -34,7 +34,7 @@
 colvarproxy_gromacs::colvarproxy_gromacs() : colvarproxy() {}
 
 // Colvars Initialization
-void colvarproxy_gromacs::init(t_inputrec *ir, int64_t step,t_mdatoms *mdatoms,
+void colvarproxy_gromacs::init(t_inputrec *ir, int64_t step,gmx_mtop_t *mtop,
                                ObservablesHistory* oh,
                                const std::string &prefix,
                                gmx::ArrayRef<const std::string> filenames_config,
@@ -96,7 +96,7 @@ void colvarproxy_gromacs::init(t_inputrec *ir, int64_t step,t_mdatoms *mdatoms,
   // GROMACS timestep
   timestep = ir->delta_t;
   // Retrieve the topology of all atoms
-  gmx_atoms = mdatoms;
+  gmx_atoms = gmx_mtop_global_atoms(mtop);
 
   // Read configuration file and set up the proxy only on the master node.
   if (MASTER(cr))
@@ -431,9 +431,7 @@ void colvarproxy_gromacs::update_data(const t_commrec *cr, int64_t const step, t
 
 
 void colvarproxy_gromacs::calculateForces( t_commrec           *cr,
-                                           t_mdatoms           *mdatoms,
                                            matrix               box,
-                                           double               t,
                                            rvec                 *x_pointer,
                                            gmx::ForceWithVirial *forceWithVirial)
 {
@@ -550,7 +548,7 @@ int colvarproxy_gromacs::check_atom_id(int atom_number)
     log("Adding atom "+cvm::to_str(atom_number)+
         " for collective variables calculation.\n");
 
-  if ( (aid < 0) || (aid >= gmx_atoms->nr) ) {
+  if ( (aid < 0) || (aid >= gmx_atoms.nr) ) {
     cvm::error("Error: invalid atom number specified, "+
                cvm::to_str(atom_number)+"\n", INPUT_ERROR);
     return INPUT_ERROR;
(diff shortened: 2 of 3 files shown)


jhenin:
Merci @HubLot !

giacomofiorin:
@HubLot In relation to this, can you also update the corresponding branch named `release-2018-colvars` on the cloned repo, https://github.com/Colvars/gromacs?

So far I am only updating automatically the "vanilla" branches (i.e. the ones that exist in the official Gromacs repos), but I'm happy to automate also the patched branches if you think that would be useful.

CC @jhenin

HubLot:
Sure but should I update only the proxy part or the whole colvars ? I'm not sure the best solution here.
If we update the whole colvars, it implies we will do it regularly and the number of release to keep updated will grow.

Maybe a compromise will be to keep only updated the release that Gromacs maintained. 
We do a last round for 2018 and keep update 2020 until Gromacs drops its support.

jhenin:
@HubLot I agree that the maintenance work should be minimized, or things will fall behind sooner or later.
@giacomofiorin In that spirit I think it would be great to automate updates to the patched branches as well. Thank you.

giacomofiorin:
Hi, I put this off for too long so I started automatically rebasing each `*-colvars` branch (when it exists) on top of the corresponding mainstream branches (currently this applies to `release-2020` and `release-2018`).

I also manually amended the commits by @HubLot to update the Colvars interface with his changes from this PR, which I write off as a one-time fix.

The following links now give you up to date snapshots (equivalent to their .zip counterparts):
https://github.com/Colvars/gromacs/archive/release-2020-colvars.tar.gz
https://github.com/Colvars/gromacs/archive/release-2018-colvars.tar.gz
and also these:
https://github.com/Colvars/gromacs/archive/v2020.5-colvars.tar.gz
https://github.com/Colvars/gromacs/archive/v2018.8-colvars.tar.gz
The latter are probably preferable, because they are based onto precise patch versions of Gromacs, instead of Git branches.

An interesting caveat is that although `v20XX.Y` is a *tag* of the corresponding `release-20XX` branch, having an additional commit containing the Colvars patch requires having a *branch* named `v20XX.Y-colvars`. I wouldn't mind having the commit automatically amended (or better yet, recreated from scratch every day using the latest patch from https://github.com/Colvars/colvars).  This way the Colvars code in each patched version would be always up to date (both proxy and module). I have not yet set this up, and would like your opinion on this.

If you agree, this automation may take care of @HubLot's concern that maintaining up-to-date Colvars patches on top of old Gromacs releases (read: 2018) would end up being too much work. I would say that only versions for which a patch exists in the Colvars repository get updated: after we remove the patch from the Colvars repository, the automatic update of that Gromacs version simply stops.

Small side note: I'm changing the patch name from `2018.3` to `2018.X` for consistency.

HubLot:
Hi,

Thanks for the update @giacomofiorin!

> https://github.com/Colvars/gromacs/archive/v2020.5-colvars.tar.gz
> https://github.com/Colvars/gromacs/archive/v2018.8-colvars.tar.gz
> The latter are probably preferable, because they are based onto precise patch versions of Gromacs, instead of Git branches.

Quick question, why those new tags are not showing up in the tag section on our forked repo (https://github.com/Colvars/gromacs/tags) ? I see the amended commit in the "release-xxxx-colvars" branches but that's all.

Note that the list of pre-patched gromacs-colvars release should be updated in the `README.md`.

> An interesting caveat is that although `v20XX.Y` is a _tag_ of the corresponding `release-20XX` branch, having an additional commit containing the Colvars patch requires having a _branch_ named `v20XX.Y-colvars`. I wouldn't mind having the commit automatically amended (or better yet, recreated from scratch every day using the latest patch from https://github.com/Colvars/colvars). This way the Colvars code in each patched version would be always up to date (both proxy and module). I have not yet set this up, and would like your opinion on this.

If I understood well, no it's not necessary to have  a branch `v20XX.Y-colvars`. The commit containing the colvars patch is already in the branch `release-XXXX-colvars`. So, once you checkout the tag `v20XX.Y` you can cherry-pick the previous commit on top of it and then create the new tag from it (https://github.com/Colvars/colvars/blob/master/devel-tools/create_gromacs-colvars-releases.md)
If you manage to do the above automatically, I think it's a good solution, and removed my concern. The only downside is that the content of the tag `v20XX.X-colvars` will not be always the same. To me, it's not a big deal since you already handle "colvars" as a "rolling-release".


> I would say that only versions for which a patch exists in the Colvars repository get updated: after we remove the patch from the Colvars repository, the automatic update of that Gromacs version simply stops.

I agree to stop the automatic update on 'deprecated' versions but I wouldn't remove the patch files though. It's better to keep them to keep a track of the 'history' but it's a personal point of view.
Note that, we should also remove those versions from the `update-colvars-code.sh` also because at some point, the compilation will failed to due divergences between colvars sources and the proxy.

In all cases, I think we should explain a little bit better how we handle gromacs-colvars versioning, either in the `README.md` or in a complementary file like `GROMACS-Releases.md`

giacomofiorin:
@HubLot Correct, it's not necessary to have a proper branch `v20XX.Y-colvars`. But its history will diverge from `v20XX.Y` nonetheless, so you end up with a detached head, which is not ideal.

Then again, thanks for pointing out that having these commits as proper branches would lead to them not being listed as "Releases/Tags" in the GH webpage. That would end up being more confusing to people downloading the code, so I deleted the branches and tagged those detached heads by deleting the old tags with the same name and recreating them.

One added benefit that I just noticed is that if the tags are deleted and recreated anew, they are listed at the top of the "Releases" page, which is probably what we want.

For the same reason, I don't think it would be appropriate to keep all minor releases of a series (e.g. 2020.1 through 2020.5), so I deleted the tags for all but the most recent minor version. If you're downloading the code with the intent of compiling a Colvars-patched GROMACS, you will want only the latest one. (There will always be people asking for an older minor release i.e. the same major release minus bugfixes: usually there are no good reasons for doing that, and if there are any I'd let people post an issue and explain them.)

To be clear, I'm not suggesting this to avoid manual work, since an automated update can easily take care of multiple minor releases at once. It's just that having one minor release for each series is more amenable to putting them in a webpage that isn't entirely dedicated to listing all possible releases:
https://manual.gromacs.org/documentation/
If you feel strongly about having patched versions of all minor releases, I can still arrange that.

HubLot:
Yes, having a detached HEAD for the tag is not ideal but i don't think there is a simpler solution.

I agree with you about keeping only the most minor version pre-patched as a release.
If users want others minor versions, they still can do it manually with the `update-colvars-code.sh` script.

giacomofiorin:
Okay, I just updated the cron job to re-apply the Colvars patch on top of the local branches `release-20{18..21}` and the two tags  `v2018.8` and `v2020.5`. The former are also rebased onto the upstream branches from the official GitLab repository, while the latter simply have the Colvars commit reapplied each time onto the same point version.

Another caveat of using a detached HEAD is that eventually our Gromacs repo will get polluted with individual commits that aren't pointed to by anything after their tags get removed. There may be a way to remove them, but I haven't looked seriously into it. It's most likely a harmless nuisance.

Now, my updater script assumes that a `-colvars` branch or tag exists, and that it has exactly one commit on top of the corresponding Gromacs branch or tag. So I'll need to manually update new branches or tags at least once before they can get updated automatically. For branches, no problem: there is a new one per year and they are not used very much other than by developers (a delay will be fine there).

For tags, we may at some point lag behind Gromacs (e.g. we may not have `2020.6-colvars` readily available when `2020.6` comes out). But a quick manual check before switching is probably appropriate here. I think that [assing the Gromacs and Colvars test suites should be sufficient. I assume that all future official Gromacs releases will pass all their internal tests, of course!

HubLot:
Thank you @giacomofiorin !

It looks to me very clean.
I agreed having detached commits in the gromacs repo is not ideal but it's just a fork so I think it's harmless too.

To be clear, your update script take care of getting a new pre-patched versions automatically  and nothing should be done manually (besides when a new tag comes up)?

Hopefully, this setup will not last for too long thanks to the work on the MDModule version :)

giacomofiorin:
Right now, the latest Colvars patch (per the `master` branch of this repo) is automatically applied onto the latest commits of each Gromacs series branch (i.e. `release-20{18..21}`) and onto the specific commits tagged `v2018.8` and `v2020.5`.

These are all hard-coded, so I will need to create `release-2022-colvars` and `v2020.6-colvars` manually before they get updated automatically. Not a major problem.

For the future, it is assumed that the patch script won't encounter any conflicts. But it might when some of your changes to Gromacs files are integrated upstream in either `release-2022` or `release-2021`. So we may need to make `update-colvars-code.sh` deal with that when the time comes.

Issue #352, opened 2020-05-18
Accept singular and plural form of some keywords


jhenin:
Mainly in biases that act on (often) one and (sometimes) several colvars.

One option to implement: a version of get_keyval that accepts two spellings / versions of the key.

giacomofiorin:
Not high enough priority, closing.

PR #387, opened 2021-01-13
Implement centerToOrigin

jhenin:
When we want to cancel center of mass motion and the target position of the fitting group is (0, 0, 0). Useful in particular for polar angles. See #386 

TODO possibly: harmonize `centerReference` and `rotateReference` ?

### DIFF
--- a/doc/colvars-refman-main.tex
+++ b/doc/colvars-refman-main.tex
@@ -1745,7 +1745,7 @@ with respect to $\{\mathbf{x}_{i}(t)\}$.
     \texttt{atoms~\{...\}} block}{%
     Defines the group of atoms of which the RMSD should be calculated.
     Optimal fit options (such as \texttt{refPositions} and
-    \texttt{rotateReference}) should typically NOT be set within this
+    \texttt{rotateToReference}) should typically NOT be set within this
     block. Exceptions to this rule are the special cases discussed in
     the \emph{Advanced usage} paragraph below.
     }
@@ -1840,10 +1840,10 @@ atom group block. This allows for the following non-standard cases:
 
 \begin{enumerate}
 \item applying the optimal translation, but no rotation
-(\texttt{rotateReference off}), to bias or restrain the shape and
+(\texttt{rotateToReference off}), to bias or restrain the shape and
 orientation, but not the position of the atom group;
 \item applying the optimal rotation, but no translation
-(\texttt{centerReference off}), to bias or restrain the shape and
+(\texttt{centerToReference off}), to bias or restrain the shape and
 position, but not the orientation of the atom group;
 \item disabling the application of optimal roto-translations, which
 lets the RMSD component describe the deviation of atoms
@@ -2634,7 +2634,7 @@ $(x_1, y_1, z_1, \cdots, x_n, y_n, z_n)$.
     Group of atoms}{%
     Block \texttt{atoms \{...\}}}{%
     Defines the atoms whose coordinates make up the value of the component.
-    If \texttt{rotateReference}, \texttt{centerReference}, or \texttt{centerToOrigin} are defined, coordinates
+    If \texttt{rotateToReference}, \texttt{centerToReference}, or \texttt{centerToOrigin} are defined, coordinates
     are evaluated within the moving frame of reference.}
 \end{cvcoptions}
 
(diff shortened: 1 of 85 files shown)


HanatoK:
@jhenin 
Looks good to me, but I can see some numerical differences between the TCL implementation and the one after applying this patch. It turns out the TCL implementation of spherical angles uses centers of mass, but this patch only move the centers of geometry to the origin. Is that a problem?

jhenin:
Good question @HanatoK . The atom group centering feature uses the center of geometry because it minimizes the (unweighted) RMSD. To use the COM, we'd need to add a user option to request that, and add an alternate code path there (in `cvm::atom_group::calc_apply_roto_translation()`). It might be a good idea, but it's a little more user-visible complexity. This all was started to make user input simpler with respect to #386 . What I like about the idea is that is solves the issue for polar angles but it adds the feature to all components. Essentially that would make the "ref" group of all distance components redundant.

HanatoK (approved):
Well, actually I think using COG instead of COM does not introduce too much hysteresis in most cases. It would be better to offer an option to move the COM to the origin, but that may change too much code. The current implementation is acceptable.

doc/colvars-refman-main.tex
```diff
@@ -4214,6 +4215,19 @@ \subsection*{Overview}
...
 }
 
+\item %
+  \keydef
+    {centerToOrigin}{%
+    atom group}{%
+    Implicitly remove translations for this group by keeping its center at the origin}{%
+    boolean}{%
+    \texttt{off}}{%
+    This option implies \texttt{centerReference}.
```
giacomofiorin, line 4225: "implies `centerReference` and automatically turns it on if needed"?

doc/colvars-refman-main.tex
```diff
@@ -4181,6 +4181,7 @@ \subsection*{Overview}
 \cvsubsec{Moving frame of reference.}{sec:colvar_atom_groups_ref_frame}
 
 The following options define an automatic calculation of an optimal translation (\texttt{centerReference}) or optimal rotation (\texttt{rotateReference}), that superimposes the positions of this group to a provided set of reference coordinates.
+Alternately, \texttt{centerToOrigin} applies a translation to place the geometric center of the group at (0, 0, 0).
```
giacomofiorin, line 4184: It seems here that `centerToOrigin` is just a short-hand for `centerReference` without the need to specify a `dummyAtom` group with coordinates on the origin. So perhaps why not present it like that instead of calling it an "alternative"?

src/colvaratoms.cpp
```diff
@@ -384,12 +388,16 @@ int cvm::atom_group::parse(std::string const &group_conf)
 
   // We need to know about fitting to decide whether the group is scalable
   // and we need to know about scalability before adding atoms
-  bool b_defined_center = get_keyval(group_conf, "centerReference", b_center, false);
-  bool b_defined_rotate = get_keyval(group_conf, "rotateReference", b_rotate, false);
+  bool b_defined_center = get_keyval_feature(this, group_conf, "centerToOrigin", f_ag_center_origin, false);
+  b_defined_center |= get_keyval_feature(this, group_conf, "centerReference", f_ag_center, is_enabled(f_ag_center_origin));
+  if (is_enabled(f_ag_center_origin) && ! is_enabled(f_ag_center)) {
```
giacomofiorin, line 393: In here, it is assumed that `ref_pos_cog` is (0, 0, 0) by default, which is true but only because its class has a proper default constructor and `cvm::atom_group::center_ref_pos()` is never called without a set of reference positions.

I would set `ref_pos_cog` explicitly as the origin right when `f_ag_center_origin` is enabled. (Actually, this is one of the situations where a map of functions in the `colvardeps` may be useful to handle callbacks: add it to the C++11 wish list?) This way, if we add COM support instead of COG later, we could do it consistently without risking to introduce subtle bugs.

giacomofiorin (approved):
A couple of minor comments/suggestions. I of course fully agree with changing the relevant `bool` flags to features. Can you flesh out what you mean by "harmonize"?

giacomofiorin:
Hi, I see now from #386 that you mean adding "To" in the keyword names, e.g. `rotateToReference`.

That looks okay to me, but I would not be optimistic that you can safely deprecate `rotateReference` in favor or `rotateToReference`. There are many snippets of Colvars input embedded in NAMD tutorials that are already not maintained: even if they were all updated, the old syntax would still survive in user scripts shared by colleagues. At the very least, I would add `rotateToReference` as a synonym for `rotateReference`, mark in the doc the Colvars version where this is changed, and keep both for the future.

We had a few instances where keywords had to change for a very good reason, but this always came at some cost later on:
https://www.ks.uiuc.edu/Research/namd/mailing_list/namd-l.2020-2021/0927.html
Also, if I look at the LAMMPS manual and mailing list, there are many messages from people who use the current doc with older versions. Even when it is clearly labeled which version the doc is for, there will always be those who don't pay attention.

So while I agree that adding "To" would be better, please do think about how important that is.

HanatoK:
@giacomofiorin 
I have a question: does Colvars use date as the version string? Even now I do not know how to check the version string without running a simulation and check the log, although checking the source code may be a clumsy option. Why not use a simpler version string as PLUMED does? Also, I think it is good to make the version string appear at the according section title of the NAMD manual.

giacomofiorin:
Sorry, I was too vague. I meant writing the Colvars version *and* the corresponding NAMD version, similar to this section:
https://colvars.github.io/colvars-refman-namd/colvars-refman-namd.html#sec:colvars_config_changes
Arguably, a reverse table that lists a corresponding map Colvars version -> NAMD version may be a useful addition.

About why the date... Until fairly recently we were completely out of the loop about the NAMD and VMD release schedule. Quite literally @jhenin and I used to learn of a new release from the users' mailing list. So the simplest approach was to use a monotonically increasing version and update it often, to identify which particular Colvars snapshot was picked for inclusion into a particular NAMD release. Hence the date, and not something more compact like "`Major.Minor.Patch`" or "`Year.Patch`" like the Gromacs developers chose to switch to. It certainly wasn't the only choice: just the one that felt simplest at the time.

It is worth mentioning that the Colvars "section" of the NAMD manual is 100 pages out of 300. On top of that, embedding much-needed input examples in the documentation was very difficult because we also do not control the LaTeX environment where the VMD and NAMD UG were built. For these reasons, future VMD and NAMD versions will contain only links to the Colvars doc hosted externally:
https://github.com/Colvars/colvars/pull/375
and for better clarity/accuracy there will be a specific version of the doc for each package's release, see e.g. for the upcoming VMD 1.9.4:
https://colvars.github.io/vmd-1.9.4/colvars-refman-vmd/colvars-refman-vmd.html
These webpages all contain the Colvars version right at the top (as they have for a few years now).

The comparison to Plumed is difficult to make here, since by choice it is not distributed together with the codes that it interfaces to. So an easily identifiable version for it is both needed and possible to achieve, because they can plan their own release schedule independently.

jhenin:
Beautiful explanation of the Colvars versioning policy @giacomofiorin !

@HanatoK You can call `cv version` in a script to get the version string.

I changed the center and rotate keywords. Ideally I would have implemented this as a version of get_keyval that accepts aliases (see #352). For now I've added a parse mode that prints a deprecation warning. The value specified for the legacy keyword is taken as default for the current keyword:
```
colvars:       # centerToOrigin = off [default]
colvars:       # centerReference = on
colvars:       Warning: keyword centerReference is deprecated. Check the documentation for the current equivalent.
colvars:       # centerToReference = on [default]
colvars:       # rotateReference = on
colvars:       Warning: keyword rotateReference is deprecated. Check the documentation for the current equivalent.
colvars:       # rotateToReference = on [default]
```
If the new keywords are used, the legacy keywords are transparent.

giacomofiorin:
@jhenin A deprecation warning implies that at some point the old keyword will be discontinued. I am not sure that the confusion that will come out of that is really worth it. Why not just accept both forms forever, as you suggested in #352?

Also in the examples, please try to keep the old syntax until a NAMD version that supports the new syntax is available. NAMD people are confused enough as it is. Or better yet, take the version of the doc for NAMD 2.14 and give it its own subdir:
https://colvars.github.io/namd-2.14/colvars-refman-namd/colvars-refman-namd.html
at which point we can offer a link to it from the top of the *current* doc page, whose URL is versionless:
https://colvars.github.io/colvars-refman-namd/colvars-refman-namd.html
The former will still feature `centerReference`, the latter `centerToReference`, but people who access the latter will have a link to the former available. Those who think in terms of NAMD versions might consider the latter a "development" version, even though we always tried to maintain it stable for the reasons explained above. While imprecise, this distinction would help people also realize that the Colvars codebase changes between NAMD versions (and new features don't just suddenly appear out of nowhere).

As for LAMMPS, it is less of a problem because each version already includes a matching copy of our PDF manual and hey, people actually *read it* (actually it would be good to link or copy the HTML version, too). However, we can do the same thing as with NAMD and VMD if consistency helps.

In summary, I'd merge this latest commit only *after* we have already up a NAMD 2.14-specific version of the doc. I would also kindly ask that you also decide whether you want VMD 1.9.4 and NAMD 2.14 advertise different versions. You could tweak the `Makefile` of the website repo, which currently copies over the current version into `vmd-1.9.4` folder.

jhenin:
Thanks @giacomofiorin , I agree with all your comments.

I'm just unsure what you mean with this:
> I would also kindly ask that you also decide whether you want VMD 1.9.4 and NAMD 2.14 advertise different versions.

giacomofiorin:
I mean, acknowledging that the doc version for NAMD 2.14 will only contain `rotateReference` (because that's what's implemented), should the doc version of VMD 1.9.4 advertise `rotateToReference` or try to be consistent with NAMD 2.14 for the time being?

jhenin:
Ah I see. I don't feel a lot of pressure to have consistency between a version of NAMD and a version of VMD, since as far as I know their release schedules are completely independent, so their lifetimes will not be aligned anyway.

giacomofiorin:
@jhenin I see that you created the `namd-2.14` subdirectory of the manual. We'll also need to add a link to it from the "main" version. Add a comment here if you want to take a stab at it so that we don't duplicate. I suspect that defining a LaTeX macro to indicate the additional path level (i.e. "/namd-2.14", "/vmd-1.9.4" or nothing at all for the "main" version) should be sufficient. The macro can then be used here:
https://github.com/Colvars/colvars/blob/134cbecb72be7f3ff1792ac8085249959a7822e9/doc/colvars-refman.tex#L133

giacomofiorin:
Edit: I just fixed the link above.

PR #393, opened 2021-02-25
Access atomic data via scripting interface

giacomofiorin:
This PR adds the capability to retrieve the current values of atomic data arrays in the `colvarproxy` class, to test their current content and evaluate whether a certain input may lead to unstable dynamics (or if it has already, find out which atomic coordinates may be the culprits).

Additionally, the root-mean-square and maximum force are also computed and returned. Hopefully this will help significantly in setting up a simulation, and possibly during a simulation as well if we e.g. print warnings about a large atomic force before things blow up in the MD integrator.

Because the scripting interface still deals only with string representations at the moment, internal data can't be modified yet. However, the new feature can be useful to achieve things like the example below. @jhenin Perhaps the same or similar summary can also be produced by the Dashboard GUI?
```
puts "Total Colvars energy:"
puts "  [cv getenergy]"
puts "RMS of applied atomic forces:"
puts "  [cv getatomappliedforcesrms]"
puts "Largest applied atomic force:"
puts "  [cv getatomappliedforcesmax]"
puts "Atom with largest applied atomic force:"
puts "  [cv getatomappliedforcesmaxid]"
```

While adding the new methods I've introduced type-specific stringification functions, like `set_result_int_vec(...)`. These have already eliminated some duplicated code, and will eventually prove useful when the script interface will also use the script language's optimal representation for each output. Eventually, the only uses of the current function `set_result_str(...)` should be when really a string of text is returned.

Little rant: to remain VMD-compatible, the templates that I added can't use `std::is_same<>`, so I had to make several explicit template specializations (one for each instance!). This is silly, but couldn't be avoided I'm afraid. Eventually when C++11-isms are allowed in VMD as well, I am hopeful that those templates could be streamlined further.

### DIFF
--- a/doc/colvars-refman-main.tex
+++ b/doc/colvars-refman-main.tex
@@ -603,14 +603,43 @@ This is normally not required during a simulation if \refkey{colvarsRestartFrequ
 Because not only a state file (used to continue simulations) but also other data files (used to analyze the trajectory) are written, it is generally recommended to call the \texttt{save} method using a prefix, rather than a complete file name:
 \cvscriptexampleinputcv{save}{"$<$job$>$"}
 
- See \ref{sec:cvscript_tcl_cv} for a complete list of scripting commands used to manage the Colvars module.}\fi % End \ifdefined\cvvmdornamd
+\cvsubsubsec{Accessing atomic data}{sec:cv_command_atomic_data}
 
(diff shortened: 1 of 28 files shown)


jhenin (approved):
Nice!

giacomofiorin:
Thanks for the fixes!

PR #397, opened 2021-03-05
Display energy and atomic force statistics in the Dashboard window

giacomofiorin:
Dashboard side of #393 

This is a working implementation, however there could be a couple of improvements:
1) The display could use something smarter than a text box, or the layout could be improved in general.
2) The location of these text boxes may be moved out of the "Actions" tab to somewhere else, to take up less real estate.


jhenin:
I agree about the placement. I think it should live in a tab of its own, that would eventually contain everything related to biases.

This is closely related to #306 . I'm adding a comment there as well because I think it can move forward.

jhenin (changes requested):
This should be moved to a separate tab, to avoid cluttering the main tab.

PR #401, opened 2021-03-10
Allow running some Tcl commands without VMD or NAMD

giacomofiorin:
It has become quite tedious to have to run the NAMD regtests for the purpose of generating the scripting API documentation (after correcting for the "failed" test). Also, building VMD or NAMD at all is really unsuitable for this purpose, let alone automated code tests.

This PR adds a few tweaks to allow building Colvars as Tcl plugin as:
```
cmake -DCOLVARS_TCL -DTCL_DIR=/usr ~/colvars/cmake
make -j
tclsh
load libcolvars.so
```
After that, one can simply update the Tcl scripting doc as:
```
cd doc
source gen_cvscript_doc.tcl
```

The library definitely isn't fully functional when built this way, but there is also near-zero risk that we'd be assuming that without a MD engine with the code as it is.


jhenin:
The implementations of pure virtual functions are de facto a new implementation of the proxy. How about keeping colvarproxy abstract and putting these into a new derived class, called, uh, maybe colvarproxy_standalone? That opens the door for future extensions to make that version of the proxy more functional.

giacomofiorin:
I would still want the ability to reuse code from the base class into the
derived classes (current or future).  So the new derived class should also
be a specialization, not almost as generic as the base class.  The new
implementations that I added are still quite generic.  But I agree that
further specializations should go in a derived class.  With that
understanding in mind, can you check if there are any functions that would
really need to go in a derived class?

If unclear at this time, are you open to keeping them in the base class for
now and specialize them as needed?  All of the currently proxies have their
own specializations already, so for those the point is moot.

I am still of course in favor of documenting clearly what functions are
required for a proxy implementation vs. optional functions.  Pure virtual
functions go beyond documentation and into actual enforcement.  Either is
good for long term maintainability...  But there is also a need for short
term maintainability, and to me building VMD or NAMD to properly test the
Tcl scripting backend is getting really cumbersome.

giacomofiorin:
To recap the conversation (for documentation purposes), the changes in this PR do *not* constitute a new implementation of the proxy class: they simply remove the constraints against building an incomplete proxy implementation for testing purposes.

If such incomplete implementation was ever used in actual computations, an error would still be raised (at runtime rather than at link time as before).

PR #408, opened 2021-06-03
Add a table to map engine versions vs. Colvars versions (fixes #392)

### LINKED ISSUE #392: Add a table to map engine versions vs. Colvars versions
For NAMD, LAMMPS and VMD, each release contains a specific Colvars version, which is the one that the vast majority will use (very few will patch the existing release if a version of Colvars is already there).  Due to functionality changes via the Colvars version, one may need to know the Colvars version before running, and without checking the source code.

The link:
https://github.com/Colvars/colvars/releases
is accurate but it points to the source code tree, without just giving the Colvars version string itself.

It is expected that this issue will worsen with NAMD 3, which is a non-monotonic upgrade of NAMD 2.

For [patched Gromacs-Colvars releases](https://github.com/Colvars/gromacs/releases) this could be taken care of by letting the updater script add a short `README-colvars.txt` to the top-level GROMACS directory.

giacomofiorin:
A Bash script in the `doc` folder populates a Markdown document `README-versions.md`.

Once finalized I'd like to copy it over to the website repository, so that it can be automatically converted to HTML with the URL https://colvars.github.io/README-versions.html

There are probably several places where that page could be linked, including the ref manual, so that people reading it could navigate to the version they need.  Once we have a few more versions of the doc built (right now we pretty much only have [NAMD 2.14](https://colvars.github.io/namd-2.14/colvars-refman-namd/colvars-refman-namd.html)), we could also link those versions of the manual from the document introduced in this PR.

### DIFF
--- a/README-versions.md
+++ b/README-versions.md
@@ -1,0 +1,37 @@
+## List of Colvars versions included in simulation/analysis packages
+
+### Versions included in LAMMPS
+LAMMPS version | Colvars version
+-------------- | ---------------
+stable_29Oct2020 | [2020-09-17](https://github.com/Colvars/colvars/releases/tag/lammps-stable_29Oct2020)
+stable_3Mar2020 | [2020-01-27](https://github.com/Colvars/colvars/releases/tag/lammps-stable_3Mar2020)
+stable_7Aug2019 | [2019-08-05](https://github.com/Colvars/colvars/releases/tag/lammps-stable_7Aug2019)
+stable_5Jun2019 | [2019-04-26](https://github.com/Colvars/colvars/releases/tag/lammps-stable_5Jun2019)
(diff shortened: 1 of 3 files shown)


giacomofiorin:
For the record, the conversion to HTML didn't work for Markdown tables, so I moved the contents of README-versions to a Wiki page instead.

Perhaps the Wiki is a good home for those pieces of information that fulfill all these conditions:
1. Deal with source code management over a longer time span than issues/PRs.
2. Are general enough to be interesting to users.
3. Do not deal with *using* the code (that's what the main doc is for).
4. Are not suitable for a README.

Because of the work put in to ensure that Colvars builds out of the box with all supported engines, there isn't much to say in terms of "installation instructions". But if that changed at any time, probably an editable Wiki page (or editable page in general) would be more appropriate than the manual that remains fixed over time (because dependencies and compilers do change over time).

PR #411, opened 2021-07-23
Patch for Gromacs 2021.X

### LINKED ISSUE #410: Support of Gromacs 2021
Hi, @jhenin @giacomofiorin @HubLot ,

Just a question: is there any plan to support Gromacs 2021?

Best,
Haohao

HubLot:
- Tests passed at 10e-4 precision
 - Version 2021, 2021.1, 2021.2 tested

Before merging, it will be better to include the REMD support in GROMACS done here (https://github.com/HubLot/colvars/tree/REMD_support).
We'll discuss about it next week with @jhenin and I'll open a PR.

close #410

### DIFF
--- a/gromacs/gromacs-2021.x.patch
+++ b/gromacs/gromacs-2021.x.patch
@@ -1,0 +1,397 @@
+diff -aur src/gromacs/CMakeLists.txt src_colvars/gromacs/CMakeLists.txt
+--- src/gromacs/CMakeLists.txt	2021-05-05 15:57:05.000000000 +0200
++++ src_colvars/gromacs/CMakeLists.txt	2021-07-13 16:26:18.103064768 +0200
+@@ -108,6 +108,7 @@
+ add_subdirectory(compat)
+ add_subdirectory(mimic)
+ add_subdirectory(modularsimulator)
++add_subdirectory(colvars)
+ if (NOT GMX_BUILD_MDRUN_ONLY)
+     add_subdirectory(gmxana)
+     add_subdirectory(gmxpreprocess)
+diff -aur src/gromacs/fileio/checkpoint.cpp src_colvars/gromacs/fileio/checkpoint.cpp
+--- src/gromacs/fileio/checkpoint.cpp	2021-05-05 15:57:05.000000000 +0200
++++ src_colvars/gromacs/fileio/checkpoint.cpp	2021-07-22 11:26:13.998639031 +0200
+@@ -72,6 +72,7 @@
+ #include "gromacs/mdtypes/pullhistory.h"
+ #include "gromacs/mdtypes/state.h"
+ #include "gromacs/mdtypes/swaphistory.h"
++#include "gromacs/mdtypes/colvarshistory.h"
+ #include "gromacs/modularsimulator/modularsimulator.h"
+ #include "gromacs/trajectory/trajectoryframe.h"
+ #include "gromacs/utility/arrayref.h"
+@@ -177,7 +178,9 @@
+  * Backward compatibility for reading old run input files is maintained
+  * by checking this version number against that of the file and then using
+  * the correct code path. */
+-static const int cpt_version = cptv_Count - 1;
++/* COLVARS : add a value of 1000 for Colvars support and
++ * prevent regular GROMACS to read colvars .cpt files */
++static const int cpt_version = cptv_Count - 1 + 1000;
+
+
+ const char* est_names[estNR] = { "FE-lambda",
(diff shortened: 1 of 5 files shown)


HubLot:
REMD support is in PR #413
I'll amend this PR once #413 is merged.

giacomofiorin:
@HubLot Will look at this in detail after you amend as you mentioned.

Out of curiosity, what is the regtest error for the unbiased-MD test `000_distance`? If that is also high-ish, then it's a divergence on the GROMACS side, which complicates things. I'm testing the library with NAMD and LAMMPS with a tolerance 10^-6, so the library code is fairly consistent but eventually I'd like to begin using reruns alongside MD runs, so that we can compare the different backends on an identical trajectory.

giacomofiorin:
TODO: integrate the citation-reporter changes introduced in #419. Currently, in NAMD the citations are printed the first time that `colvarproxy_namd::setup()` is called, i.e. right before the run starts. In LAMMPS, this is done at the *end* of a run, so that the report on the features that were just used (and not just the module) will be acknowledged (#418 may change this).

I'm not 100% sure when the appropriate time is for GROMACS, but I would lean toward a behavior similar to NAMD. GROMACS users are just as likely not to run through the specified number of steps and just use checkpointing to continue.

HubLot:
Hi,

I just updated the PR with the REMD support added in #413.

About the precision, I should have precised that all tests in the library folder passed with 10^-6 precision.
Only tests in the interface have lower precision: 10^-5 for the MPI/tMPI tests and 10^-4 for the restart in tMPI.

I'll add the new citation-reporter changes in a following commit.

HubLot:
Hi, 

I include the citation in the Gromacs proxy, I tried to mimic the NAMD behavior so the output will be like this:

```
[...]
colvars:   # outputTotalForce = off [default]
colvars:   # outputAppliedForce = on
colvars:   # subtractAppliedForce = off [default]
colvars:   # runAve = off [default]
colvars:   # corrFunc = off [default]
colvars: ----------------------------------------------------------------------
colvars: Collective variables initialized, 1 in total.
colvars: No collective variables biases were defined.
colvars: ----------------------------------------------------------------------
colvars: Collective variables module (re)initialized.
colvars: ----------------------------------------------------------------------
colvars: Re-initialized atom group for variable "one":0/0. 4 atoms: total mass = 54.028, total charge = -0.72.
colvars: Re-initialized atom group for variable "one":0/1. 4 atoms: total mass = 54.028, total charge = -0.4.
colvars: 
colvars: SUMMARY OF COLVARS FEATURES USED SO FAR AND THEIR CITATIONS:
colvars: 
colvars: - Colvars module:
colvars: - Optimal rotation via flexible fitting:
colvars: - distance colvar component:
colvars:   Fiorin2013 https://doi.org/10.1080/00268976.2013.813594
colvars: 
colvars: The restart output state file will be "test.restart.colvars.state".
colvars: The final output state file will be "test.colvars.state".
colvars: Opening trajectory file "test.colvars.traj".

starting mdrun 'Protein'
20 steps,      0.0 ps.
[...]
```

I didn't include GROMACS citations because it's already done in the GROMACS file log and I thought it was redundant

giacomofiorin:
Ok, great, thanks for taking care of that. Please add also a line similar to this:
```
colvars->cite_feature("Colvars-GROMACS interface");
```
so that your own implementation is credited in the future.

Regarding the citation to the GROMACS main paper, I don't think it's too bad to have one redundancy. All other engines provide code that Colvars relies on (in a few cases even to compute collective variables). Some of this code is already acknowledged because you're citing the engine's main paper anyway, but certain contributions may not be as well described in the main paper. A relevant example for GROMACS would be a paper describing MDModules, once it is completed. Absent that, the GROMACS main paper could serve as placeholder: what do you think?

giacomofiorin:
I'll need a few days to properly take a look at the checkpointing, if you don't hear from me by the end of next week go ahead and merge.

HubLot:
Thanks for the feedback. I added the code line you suggested.
I'm ok to add also the main paper of GROMACS too, it's true it's not so much of a redundancy.

giacomofiorin:
FYI I ran the tests with GCC 11 and some of the tests passed at **10e-3**, not 10e-4. We've seen this before e.g. when significant changes were introduced in LAMMPS or NAMD 3 (the NAMD 2.x is quite stable in that regard and allows us to keep using 10e-6 or lower).

I also created the patched 2021 release series, and will keep it updated using the same script as for 2018 and 2020. Eventually I mean to move that to GitHub Actions.

PR #418, opened 2021-08-11
Extend support for fix_modify to remaining script API commands

giacomofiorin:
This PR extends the interface introduced in #216 while taking advantage of the scripting API enhancements from #316.

The LAMMPS command `fix_modify` is used as the main command, which behaves much in the same way as the `cv` command in the Tcl codes (VMD and NAMD). Compared to Tcl, however, functionality is more limited: the commands' string output cannot be captured and used as-is in conditionals (only printed to a separate log file). EDIT: this will be overcome in future PRs by passing references to LAMMPS variables.

An LAMMPS-specific limitation has always been that the full initialization of the module needs to be done upon the first `run` command, because of the order in which all the LAMMPS data structures are initialized. With this change, full initialization of Colvars is still deferred until when `run` starts, but the empty module can still be used and queried before that happens.

To properly document the new scripting feature, I introduced a distinction in the manual between scripting API and callbacks. These are two separate capabilities, and at the moment we support only the former in LAMMPS.

Other changes that affect other engines are mostly internal to the `colvarproxy` class.

Notable new abilities for LAMMPS users are the following:
- [x] Defining new collective variables after the first initialization of the fix, even if this means requesting additional atoms. 
- [x] Modify the internal parameters of the fix as well as those of the module (i.e. it is possible to redefine `seed` or `tstat`, and to update/change a restraint). The only restriction is that a single `fix_modify` command cannot do both in a single line.
- [x] Define Colvars configuration inline in the LAMMPS script, including expansion of LAMMPS variables.

Current limitations:
- Although it is possible to use `fix_modify` at any time, its results will only reflect the state of the module at the latest `setup()`, ignoring any configuration that may have been added more recently because of deferred initialization.

Testing :
- [x] Check for correct behavior using MPI-parallel runs (with and without partitions).
- [x] Update `fix_colvars.rst`


akohlmey:
> Compared to Tcl, however, functionality is more limited: the commands' string output cannot be captured and used as-is in conditionals (only printed to a separate log file).

There are ways around this limitation. You could have a "variable" or "output" keyword that takes a variable name as argument and then can have the return value/text/string/whatever assigned to that variable. Check out what the KIM package in LAMMPS does with e.g. `kim query`.

akohlmey:
On top of using `fix_modify` you could also have a custom colvars wrapper function in the library interface. While that doesn't give you access to library functionality in LAMMPS input files, but you could make it available from Python (or other script languages including Tcl (via swig)).

giacomofiorin:
Yes, passing a LAMMPS variable as argument is one of the option I was thinking about. Thanks for pointing out the `kim` example!  And of course, the library interface (C and Python) is the optimal way to go in general.  Now that you and @rbberger have streamlined a lot the installation process, wrappers would be much easier to add.

At the moment I'd like to have all use cases that *don't* need the deeper library interface sorted out this way.

If possible I would like to have the fix initialized after the `fix` command, so that subsequent uses of `fix_modify` can operate on it immediately. Is there any case when a fix's `setup()` function is called before the next `run` command?

akohlmey:
> If possible I would like to have the fix initialized after the `fix` command, so that subsequent uses of `fix_modify` can operate on it immediately. Is there any case when a fix's `setup()` function is called before the next `run` command?

No. Any such initialization would have to be called from either the constructor or - if it needs access to the virtual function tables somehow - from the post_constructor() function.

giacomofiorin:
Ok, so what I guess we can do is create both the fix *and* the module at the same time, but only initialize the latter as empty, and wait until `setup()` to fully initialize it.

The advantage would be that we can allow multiple calls to `fix_modify` supply multiple configuration parameters and/or files, which would help when some of those parameters/files are given upon certain conditions only.

giacomofiorin:
@akohlmey After further testing, the changes seem to work okay for me.  However, given that the LAMMPS Summer 2023 stable release is due this week, it seems easier to not submit it at this time.  Including these and other improvements together in the next patch release seems like the best option.

akohlmey:
> However, given that the LAMMPS Summer 2023 stable release is due this week

I have to push that date. Probably to mid-July. Nevertheless, this seems better done *after* the stable.

devel-tools/build_test_library.cmake
```diff
@@ -85,6 +85,7 @@ execute_process(
   -D BUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
   -D WARNINGS_ARE_ERRORS=ON
   -D CMAKE_VERBOSE_MAKEFILE=ON
+  -D ENABLE_COVERAGE=ON
```
jhenin, line 88: I expected coverage reporting to be useful for manually working on the tests. What's its purpose inside the CI workflow, is there any automated treatment?

gromacs/gromacs-2022.x/colvarproxy_gromacs.cpp
```diff
@@ -95,7 +95,7 @@ void colvarproxy_gromacs::init(t_inputrec *ir, int64_t step, const gmx_mtop_t &m
   updated_masses_ = updated_charges_ = true;
 
   // GROMACS timestep
-  timestep = ir->delta_t;
+  set_integration_timestep(ir->delta_t * 1000.0);
```
HubLot, line 98: I would keep the comment saying GROMACS is uses picoseconds  for the time step hence the 1000 factor.

jhenin (commented):
Looks excellent. Just one question about CI.

HubLot (approved):
I'm okay with the timestep change

devel-tools/build_test_library.cmake
giacomofiorin: I'm using that same CMake script also for manually checking the code on the local worktree.

An automated treatment would require additional tools, which I didn't set up yet. Right now, the library tests only parse the input, but do not run actual computations. So at the moment the flag is useful, just not yet at the stage where we would run a coverage analysis unattended. Do you feel that there is a reason *not* to include it at all?

jhenin: I see - my only concern was that there could be an impact on the CI run time. If you don't think that's the case, then I'm good.

giacomofiorin: No, the code may run a bit slower, but it compiles in the same amount of time, and ccache can still skip that altogether if there is a hit (I just checked again).

jhenin (approved):

PR #419, opened 2021-08-13
Generate citation report while running computations

giacomofiorin:
This PR adds a database of citations to papers referencing different pieces of the Colvars library, and introduces a mechanism to print during a run a report of all papers that should be cited.  The change is motivated in part by recent improvements in the LAMMPS [CiteMe class](https://github.com/lammps/lammps/pull/2545).  There is also a growing number of features that are not described in the Fiorin, 2013 Mol Phys paper, especially when written by people who weren't co-authors of it. These ought to be cited, in addition to and together with it.

Relevant citations are listed into their own BibTeX file, which is parsed by a Python script to generate a C++ source file that fills the database. Features listed in that file should be references verbatim in the code when they are used.

Here are the proposed criteria:

1. **_The code should cite only papers describing software implementation_**.  *Methodology* papers of course would still be cited in the *user documentation*.  For example, LAMMPS prints at runtime citations to relevant code but the manual also lists many more references to the theory behind that code.  There may be downsides in some cases, e.g. people who don't read the manual at all may not know where a method comes from.  But it is hard to deny that the academic publication and peer review process already *massively favors methodology work over implementation work*.  To correct this, there needs to be a place where the former will not dilute the latter.

2. **_A very brief 1-line max description of the code feature being credited is added to the report alongside the citation_**, to give the user a clear idea of what a paper is being cited for.  Most classes already have a clear correspondence with physical/mathematical quantities.  It would be of major help if those descriptions also matched section titles in the manual.  When multiple features reference the same paper, these should be all listed individually but the paper included once.

3. When dealing with **_inheritance and composition_** of objects, things may get a bit complicated.  Some classes are only thin wrappers around existing ones, while others significantly supersede the base class.  I would advocate for making the 1-line descriptions as accurate as possible, and **_reference all pieces of code that make up a user-facing feature unless it is abundantly clear that some pieces are really minor_**.  This is not a big issue at this time because most such "minor pieces" are part of the library itself, and I and @jhenin get credited anyway via the Fiorin, 2013 citation.  But there are other cases, especially when dealing with external codes that would normally be overlooked. E.g. I would like to add a citation to the Wells2007 paper when using GridForces in NAMD, as long as part of that code is used.

4. **_Longer term, some code eventually may be replaced altogether and the citation replaced as well_**.  As a minor example, the linear bias was originally coded by Andrew White in https://github.com/Colvars/colvars/pull/18, but since then I've rewritten it entirely (of course, the significant feature that he introduced, ALB, remains associated with the White2014 paper).  So this is not a big deal now, but it may be later on.  There is not a lot of pressure to supplant existing code, but sometimes it will happen just like new methodology takes over old methodology.

Possible current issues:
- Quite many features are associated with the Fiorin 2013 paper, so it is likely that I have overlooked a few cases where a significant part of the code was described elsewhere: I apologize in advance if that's the case, and would definitely want to correct these before merging this PR.
- There are several features for which I really couldn't find a good citation yet.  I put those at the very ~~beginning~~ end of the file, so that we can keep checking later as new papers are published.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -30,6 +30,16 @@ jobs:
     - name: Test build recipes
       run: bash devel-tools/check_build_recipes
 
+    - name: Convert BibTeX references to code
+      shell: bash
+      working-directory: doc
+      run: |
+        make update-code-refs
+        if [ -n "$(git status --porcelain ../src/colvarmodule_refs.h)" ] ; then
+            echo "Please update the code references in the doc folder" >& 2
+            exit 1
+        fi
+
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v1
       with:
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all pdf clean clean-all
+.PHONY: all pdf clean clean-all update-code-refs
 
 PDF=colvars-refman-lammps.pdf colvars-refman-namd.pdf colvars-refman-vmd.pdf colvars-refman-gromacs.pdf
 BIBTEX=colvars-refman.bib
@@ -23,3 +23,6 @@ clean-all: clean
 version-list:
 	./print_engine_versions.sh > ../README-versions.md
 
+
+update-code-refs:
+	python3 extract_code_refs.py > ../src/colvarmodule_refs.h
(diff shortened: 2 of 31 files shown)


giacomofiorin:
The integration of the new feature citation mechanism with NAMD and LAMMPS is now complete IMO. For VMD, I added a scripting command `featurereport` that could be invoked interactively or via the Dashboard.
For GROMACS, I'd wait until after #411 is merged before touching the files.
I'll leave open for another week or so for comments.

jhenin (approved):
These look very good to me. I think this can be merged and Gromacs support can be added, using the native interface of that back-end (which may have been inspired by lammps?).

akohlmey:
> using the native interface of that back-end (which may have been inspired by lammps?).

I think that Gromacs had this kind of thing first, which then inspired the CP2k folks, which in turn inspired the LAMMPS interface.

PR #420, opened 2021-09-01
Build backends automatically

giacomofiorin:
This PR introduces automated builds for the supported backends (GROMACS, LAMMPS, NAMD and VMD) for CI purposes. This is generally useful, but I'm specifically motivated by (1) upcoming work needed on #418 and (2) this is becoming easier to do for the two UIUC codes, which continue to restrict direct downloads but some of their dependencies do not now (e.g. Charm++). I just created private repositories of NAMD and VMD under the Colvars org for this purpose (hopefully the clone works). 

Although a work in progress, this PR is *not* marked as a draft, just so that the Actions workflow itself may be tested.

Some issues:
- The current regression tests are a bit sensitive to the compiler used and changes in the MD engines, causing divergence of the MD trajectories for identical Colvars code. (Occasionally a test fails at 10<sup>-5</sup> precision). This should be addressed before including the regression tests in the workflow. (E.g. perhaps a simple functionality tests using the same inputs may be run first?)
- Testing automatically the interactive and graphical behavior of VMD is difficult, to say the least. @jhenin perhaps some of the Dashboard routines can still be tested automatically? If so, I can include Tk in the build environment, too.


jhenin:
Good question about the Dashboard routines. Can we create a graphical window in that environment? If yes, I can probably script some typical tasks.

giacomofiorin:
@jhenin I didn't see much specific about running X11 apps on the Actions' instances. More often than not, a full set of graphical libraries takes a bit to install via package managers, and you're better off packing them into a container to speed that up. The Singularity container that I'm pulling in these workflows has these already in, and that's the same one that I'm using to build VMD on my workstation.

I tested that container/VMD combination on both the workstation (CentOS 7, NVIDIA GPU) and my laptop (Fedora 33, Intel GPU) and it works okay. For NVIDIA GPUs, you're much better off passing `--nv` to the Singularity command both in terms of stability and performance. If you prefer to use Docker (I can't) I am sure that it has equivalent functionality. CentOS 7 is the platform where John will most likely make the next Linux x86_64 builds (if he isn't already), so I encourage you to try this container out going forward. (Note: don't take CentOS 7 as a general recommendation for all software, 'coz it gets pretty political quickly...)

As for testing graphical routines without a display, I noticed unfortunately that some work is needed. Doing naively the following:
```
unset DISPLAY
singularity exec CentOS7-devel.sif <vmd-centos7-build>
...
vmd > package require cv_dashboard
1.3
vmd > cv_dashboard
...
invalid command name "winfo"
```
so clearly Tk does not play nice with that arrangement.

Perhaps using a tool like Xpra (look it up) may help, but I haven't yet had the time to test it. Do you want to try?

giacomofiorin:
All codes except GROMACS (which version to track remains to be defined) are now being built during the workflow, and the existing regression tests with LAMMPS and VMD being run. All pass as long as a consistent toolchain is used (see container) so I'm inclined to put them in.

I didn't include yet the NAMD tests, because of two small issues:
- `000_distancedir_harmonic-ddir-moving` needs fixing, because I made its input deliberately inconsistent to raise an error, but seeing as NAMD ends up returning with an error I'd better just make the input valid.
- `000_distance-wall-bypassExtended-off` was generated with a different GCC version and exhibits the 10<sup>-5</sup> divergence that I mentioned above. @jhenin Okay if I regenerate the reference files with the same GCC as the majority tests? (There are a few more that were likely generated with recent GCCs, but I don't feel like touching them as long as the tests pass.)

giacomofiorin:
P.S. I went ahead and regenerated the `000_distance-wall-bypassExtended-off` regtest output using the CentOS 7 build and the current code snapshot. See if it looks right to you, then I'll add the NAMD tests to the workflow and we can merge.

jhenin:
@giacomofiorin I'm fine with the updates to reference data for regtest.

About Xpra and testing graphical commands in Singularity, I have a technical gap to close before I can be efficient on that front.

giacomofiorin:
@jhenin While looking back at this topic I found this action, based on [XVFB](https://www.x.org/releases/X11R7.6/doc/man/man1/Xvfb.1.xhtml). It looks like it may be a simple enough approach to testing VMD and the Dashboard (including calls to Tk) with a headless display:
https://github.com/marketplace/actions/gabrielbb-xvfb-action
JavaScript devs seem to be using it reasonably often.

The Singularity side should work okay: I'm using containers fairly often to run graphical apps, including VMD (we're still stuck at CentOS 7 in the lab). If you use a Nvidia GPU, you will probably need the `--nv` flag: otherwise, on my laptop (Intel GPU) no extra steps are needed to run VMD inside a container. If you have a Ubuntu 20.04 physical machine or VM ready, you should be able to confirm this quickly.

PR #423, opened 2021-09-14
Set GMX_VERSION_STRING_OF_FORK while patching when it is available

giacomofiorin:
Adapt the Colvars patching script so that a descriptive label (e.g. `Colvars-2021-09-13` at the moment) is added to the Gromacs version string. The relevant CMake variable was introduced in the 2020 release series, so the script checks for it first before patching.

Having adapted the patching script affects the pre-patched releases at https://github.com/Colvars/gromacs, but also the use case of somebody patching a Gromacs source tree on their own. 

See: https://github.com/Colvars/gromacs/pull/2

### DIFF
--- a/update-colvars-code.sh
+++ b/update-colvars-code.sh
@@ -101,6 +101,12 @@ else
   fi
 fi
 
+
+source $(dirname $0)/devel-tools/version_functions.sh
+
+COLVARS_VERSION=$(get_colvarmodule_version)
+
+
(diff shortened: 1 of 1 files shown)


mabraham:
Looks good, thanks!

jhenin (approved):
Looks good, does the job.

mabraham:
OK, tag me if I can help on the GROMACS end

PR #428, opened 2021-11-01
More complete documentation for index file format (NDX)

giacomofiorin:
The GROMACS forum thread at:
https://gromacs.bioexcel.eu/t/error-compiling-gromacs-2020-5-patched-with-colvar-module/3048
raised an interesting issue about the NDX file format (disregard the GCC 11 compilation issues, which affect GROMACS files, not Colvars: hopefully the GROMACS devs will fix those in their repository).

When a group name contains spaces it cannot be parsed correctly by Colvars. This is reasonable behavior, but wasn't explicitly documented yet.

What's most striking is that the GROMACS parser formally allows spaces, but then uses only the first word, e.g. if one has an index file that contains the two groups labeled `[ Protein C-alpha ]` and `[ Protein backbone ]`, they would end up with two distinct groups both named `Protein`.  So while the GROMACS doc doesn't say it, it doesn't support spaces either and doesn't warn the user about it.

To simplify things I pulled out the Tcl function from #374 and put it in a script in the `colvartools` folder, with an example in the doc.  Given that this is just enough Tcl to not upset most people, I would be inclined to merge this PR and abandon #374.

Most importantly, the function warns the user about setting a group name that contains spaces or tabs.

### DIFF
--- a/colvartools/write_index_group.tcl
+++ b/colvartools/write_index_group.tcl
@@ -1,0 +1,45 @@
+# Write a VMD selection into a GROMACS index file.
+
+# 1st argument is either a Tcl file channel or a file name: in the latter
+# case, content will be appended to that file.
+# 2nd argument is an atom selection proc, as returned by the atomselect
+# command.
+# 3rd argument is the name of the group.
+
+proc write_index_group { ndxfile sel name } {
+    # Check that the name does not contain spaces or tabs
+    if { ([string first " " "${name}"] >= 0) ||
+         ([string first "	" "${name}"] >= 0) } {
+        puts stderr "Error: group name may not contain spaces or tabs."
+        return
+    }
+
+    if { [catch [list tell "${ndxfile}"]] } {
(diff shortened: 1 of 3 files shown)


HubLot:
Looks good!
I didn't know about the GROMACS bug on ndx file...

giacomofiorin:
@HubLot That was a useful thread, wasn't it?

I don't know if the GROMACS thing is a "bug" or a "feature": the index files generated by `make_ndx` are correct, and they probably do not care about index files generated by other codes :-/  It's probably just easier to fix MDAnalysis so that it doesn't write spaces?

HubLot:
Yes clearly!
Right, `make_ndx` always add a `_` between words and you can't rename a selection with a space.
But it's neither a bug at the MDAnalysis side because it's a user choice to give a selection's name (by default, it's mdanalysis001). See [here](https://docs.mdanalysis.org/stable/documentation_pages/selections_modules.html)

PR #429, opened 2021-11-08
Implement OSRW

jhenin:
That is, measuring and biasing the force acting on an alchemical coordinate. Currently only implemented in the Tinker-HP proxy.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -1,6 +1,6 @@
 .PHONY: all pdf clean clean-all update-code-refs update-cvscript-cmdline-doc
 
-PDF=colvars-refman-lammps.pdf colvars-refman-namd.pdf colvars-refman-vmd.pdf colvars-refman-gromacs.pdf
+PDF=colvars-refman-lammps.pdf colvars-refman-namd.pdf colvars-refman-vmd.pdf colvars-refman-gromacs.pdf colvars-refman-tinkerhp.pdf
 BIBTEX=colvars-refman.bib
 
 all: pdf
--- a/doc/colvars-refman-gromacs.tex
+++ b/doc/colvars-refman-gromacs.tex
@@ -6,7 +6,6 @@
 % Colvars repository at GitHub.
 
 
-\newcommand{\GROMACS}{}
 \newcommand{\MDENGINE}{Gromacs}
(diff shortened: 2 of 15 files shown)


jhenin:
I think this is good for merging now, which will prevent accumulating conflicts with other branches. Any objections @giacomofiorin ?

giacomofiorin:
Well, not that it's not considered a "draft" any more, I think you should at least document `extendedMass`, which unlike the alchemical stuff will be functional out of the box. 

It may be worth mentioning it here:
```
\cvsec{Syntax changes from older versions}{sec:colvars_config_changes}
```

Also, be prepared to dealing with people who use both `extendedFluctuation` and `extendedMass` in their inputs.

Of those who read the manual, the majority still decides what keywords are relevant for their run. But the number of those who just throw in *every* keyword without consideration is growing.

jhenin:
Ah yes! That was an oversight, thanks. Will fix.

jhenin:
Documentation done. After thinking about it, I don't feel it belongs in the "Syntax changes" section, given that this only adds new keywords without changing the functionality of existing ones (and the section intro states "This is not a list of new features:
its primary purpose is to make you aware of those improvements that affect the use of old configuration files with new versions of the code.").

giacomofiorin:
That's true, but since the new keywords are alternatives to old ones, they
do change their behavior in that they are not required any more.

I don't feel hugely strong about this issue, just inviting to think about
it for a moment: you have already a bunch of input files for eABF
circulating, and the new keywords introduce a different way of doing the
same.

How about adding some examples (of a few lines each) for each method
directly in the manual?  That could go a long way.

jhenin:
The new keywords are really designed to handle a new situation - the alchemical case. Offering to specify the force constant as an alternative is just for convenience, but thanks to your remarks I realize now that it might be more confusing than convenient. I'm going to simplify things by offering two independent interfaces for the alchemical and the geometric cases, and removing the alternate possibilities and the redundant keyword.

jhenin:
In the process of updating the documentation, I've been led to add the Tinker-HP ref manual, which was necessary sooner or later.

jhenin:
@giacomofiorin OK to merge this?

PR #430, opened 2021-11-09
Improve Github Actions

HubLot:
Hi,

This a draft for improving the Github Actions of Colvars by making the actions more modularizable.

For the first workflow ("Build & CodeQL analysis"), I split it into 3 different jobs (with one being dependant of another).

For the backend workflow, I'm using the "Reusable workflow" feature introduced [recently](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows). I created a template workflow ("backend-template.yml") which is in charge of installing, building and testing colvars with a generic backend. Then, I call this workflow in the "test-backends.yml" where I specified the variable which belongs to the workflow tested.

This separate the test procedure and produce independant job for each backend. This increase readability in case of failure.

It's a still a work in progress so comments/improvements are welcomed!


.github/workflows/test-backends.yml
```diff
@@ -0,0 +1,91 @@
+name: "Build backends with an updated Colvars module"
+
+on: [push, pull_request]
```
giacomofiorin, line 3: @HubLot As far as I understood, the standard `pull_request` would run the workflows entirely from the PR's head, and since this can be located anywhere, the workflows wouldn't have access to the repository secrets. And unfortunately we need those as long as VMD and NAMD remain with their restrictive license and their repositories need to remain private.

giacomofiorin:
An additional improvement that I would like to have is moving CodeQL at the very end, and have the build tests run with the benefit of ccache: this way, if there is a problem there, the job would fail rather quickly. Same goes for more detailed code quality checkers like the Sun compiler (and eventually cppcheck?), which run no slower than CodeQL but are more likely to raise issues than CodeQL is.

.github/workflows/test-backends.yml
HubLot: you're right, I'm trying to see if there is a solution to that but with little hope. 
It will work for PRs with a branch coming from the official 'colvars' depot but not from a fork.

HubLot:
Exactly, in this PR, CodeQL is performed separately to speed up the actions.

----

To better test the CI modifications and access private NAMD/VMD repos, I'll close this PR and open a new one  based on a branch of the official colvars repo

giacomofiorin:
Hi, to access those repos you need the private keys, stored in:
https://github.com/Colvars/colvars/settings/secrets/actions
but I'm not aware of an easy way to retrieve them from there.  I'll send you a private message.

I'm assuming that you have downloaded VMD or NAMD yourself at some point and accepted their license terms, if not please do that. Also download [Oracle Developer Studio](https://www.oracle.com/tools/developerstudio/downloads/developer-studio-jsp.html) just once yourself.

Issue #388, opened 2021-01-20
Support Lepton or other kinds of custom functions in GROMACS


HanatoK:
Currently Colvars can use neither lepton custom functions nor TCL scripting when integrating with GROMACS. Is there any way to easily implement a CV without hacking the C++ source of Colvars in GROMACS?
I take a look at PLUMED, and they bundle the lepton source code. Can Colvars do the same thing?

giacomofiorin:
Hi, Lepton is definitely doable. @jhenin was looking into adapting the CMake recipe that I created for embedding Lepton in LAMMPS, but it should be relatively straightforward.

For Tcl, that would involve a bit of a discussion with the Gromacs devs. We strive to keep the same build system and do not add significant dependencies. They seem to have enough work set already with Python/gmxapi at the moment, which arguably will be much more useful than Tcl when it builds consistently on all current platforms without needing too many tweaks :-)

Since you took a look, can you confirm that the Lepton version bundled with Plumed is now the official version without local changes?

HanatoK:
@giacomofiorin 
No, PLUMED extended a few inverse hyperbolic functions. I proposed to merge those extensions to openmm's repository in #231 , but only atan2 was added back.

HanatoK:
@giacomofiorin 
No, PLUMED extended a few inverse hyperbolic functions. I proposed to merge those extensions to openmm's repository in #231 , but only atan2 was added back.

jhenin:
Yes I looked into adding Lepton, but I'm a beginner with CMake, so that was harder than I thought. There's also the question of proposing a directory structure that will be acceptable to Gromacs folks, so there were tough choices to make. I think we need to talk to gmx developers about it first. We'll do it asap and get back to you.

giacomofiorin:
@HubLot @jhenin If you want to take a stab at this, I would suggest trying with a build system where Lepton is downloaded externally (via the OpenMM repository as we have in the Actions) and then linked to GROMACS. This would avoid the issue of where to place it, although in my opinion `src/external` is a good enough place for our own patched releases.

Given that Lepton is small enough that you can also build it easily with CMake, check out out how this is done in LAMMPS, e.g. `cmake/Modules/Packages/MDI.cmake` for a recent example.

After the holiday, I can also take a closer look.

HanatoK:
[...]

PR #432, opened 2021-11-18
Embedded tcl

jhenin:
This is a fairly simple way to allow the use of Tcl scripts for all back-ends, or without a back-end at all. One major benefit is allowing for unit tests using Tcl scripts. Also helps #388

Needs some coordination with other efforts, in particular #462.

Fixes segfault when building library with Tcl enabled and running file_io unit test.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -13,6 +13,10 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 
 option(WARNINGS_ARE_ERRORS "Treats warnings as errors" OFF)
 
+# message(STATUS "Enabling sanitize-address")
+# add_compile_options(-fsanitize=address)
+# add_link_options(-fsanitize=address)
+
 if(NOT DEFINED LEPTON_DIR)
   set(LEPTON_DIR ${COLVARS_SOURCE_DIR}/lepton)
 endif()
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -17,6 +17,9 @@ option(ADDRESS_SANITIZER "Build with Address Sanitizer for memory debugging" OFF
 if(ADDRESS_SANITIZER)
   add_compile_options(-fsanitize=address)
   add_link_options(-fsanitize=address)
+  if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "GNU"))
+    message(WARNING "-DADDRESS_SANITIZER flag not tested yet for this compiler.")
+  endif()
 endif()
 
 if(NOT DEFINED LEPTON_DIR)
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -13,9 +13,11 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 
 option(WARNINGS_ARE_ERRORS "Treats warnings as errors" OFF)
 
-# message(STATUS "Enabling sanitize-address")
-# add_compile_options(-fsanitize=address)
-# add_link_options(-fsanitize=address)
+option(ADDRESS_SANITIZER "Build with Address Sanitizer for memory debugging" OFF)
+if(ADDRESS_SANITIZER)
+  add_compile_options(-fsanitize=address)
+  add_link_options(-fsanitize=address)
+endif()
 
 if(NOT DEFINED LEPTON_DIR)
   set(LEPTON_DIR ${COLVARS_SOURCE_DIR}/lepton)
--- a/namd/src/colvarproxy_namd.h
+++ b/namd/src/colvarproxy_namd.h
@@ -59,9 +59,9 @@ protected:
   cvm::real amd_weight_factor;
   void update_accelMD_info();
 
-  void init_tcl_pointers();
-
 public:
+
+  virtual void init_tcl_pointers();
 
   friend class cvm::atom;
 
--- a/namd/src/colvarproxy_namd.h
+++ b/namd/src/colvarproxy_namd.h
@@ -61,7 +61,7 @@ protected:
 
 public:
 
-  virtual void init_tcl_pointers();
+  void init_tcl_pointers() override;
 
   friend class cvm::atom;
 
@@ -122,7 +122,7 @@ public:
     return simparams->dt;
   }
 
-  virtual cvm::real get_accelMD_factor() const {
+  cvm::real get_accelMD_factor() const override {
     return amd_weight_factor;
   }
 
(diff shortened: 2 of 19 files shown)


src/colvarmodule.cpp
```diff
@@ -386,10 +388,18 @@ int colvarmodule::parse_global_params(std::string const &conf)
   parse->get_keyval(conf, "scriptingAfterBiases",
                     scripting_after_biases, scripting_after_biases);
 
+#if defined(COLVARS_TCL)
+  std::string source_Tcl_script;
+  if (parse->get_keyval(conf, "sourceTclFile", source_Tcl_script)) {
```
giacomofiorin, line 393: Since you're getting closer to finishing this, my concern here is that this would force the execution of arbitrary Tcl code in the middle of the Colvars initialization.

I would *at the very least* defer that execution until Colvars is properly set up, and ideally control when that code is executed from the simulation engine.

If you have a Tcl interpreter driving the whole thing as in VMD or NAMD to a partial extent, it would be simpler to just use `source` IMO.

jhenin: > Since you're getting closer to finishing this, my concern here is that this would force the execution of arbitrary Tcl code in the middle of the Colvars initialization.
> 
> I would at the very least defer that execution until Colvars is properly set up, and ideally control when that code is executed from the simulation engine.

Good point. This needs deferring the check for the existence of the force callback until everything is initialized, which I'm doing now.
 
> If you have a Tcl interpreter driving the whole thing as in VMD or NAMD to a partial extent, it would be simpler to just use source IMO.

I agree. This is definitely reserved for the other engines that are not Tcl-driven at all, or to write a Colvars input that is fully portable between one of these two codes and one of the other engines.

giacomofiorin: The `colvarmodule` object itself is constructed separately from the configuration parsing. We just don't use this separation much yet, although I would do that for #418.

So you can test for the presence of the callback functionality after construction, but *before* the configuration is parsed.

However, pretty much everyone will use their Tcl code to modify existing variables or biases, so those would need to be defined before you source the script.

jhenin:
Good, I'll use override.
Two thoughts on the architecture:
- I wonder if colvarproxy_tcl would be better defined as a member of colvarproxy rather than a base class, since it is an optional component of the proxy. Instantiating it when Tcl is not available is awkward (e.g. having an object full of placeholder data to have the right size even if it is not usable).
- Some components of colvarproxy_tcl are generic, and colvarscript has Tcl-specific functions. The distribution of tasks between the two is not obvious to me. I would like to clarify it, either by moving everything Tcl-specific to colvarproxy_tcl, or by moving everything that never needs a backend-specific override to colvarscript.

giacomofiorin:
Both observations make sense. The current layout of the `colvarproxy_*` classes was never meant to be more than an incremental improvement, i.e. organizing into groups all the many functions of the single massive `colvarproxy` class that we used to have previously. Moving to using members instead of base objects would improve organization even more. E.g. if you have only 

I'm fine with clarifying what Tcl features are used where. IMO the distinction between what `colvarscript` and `colvarproxy` in this regard is that the former handles *scripting commands* (i.e. Tcl calling Colvars), while the latter handles *callbacks* (i.e. Colvars calling arbitrary scripted functions). If that distinction was blurred, it was mostly because we were initially handling all of the "Tcl stuff" in `colvarproxy_vmd` and `colvarproxy_namd`, but it doesn't need to stay that way, especially now that you want to put Tcl in TinkerHP.

If you want to tidy up the callback code, I would strongly consider making a base callback class, from which to derive a Tcl specific one. I don't have an opinion on the exact relationship of this class with the proxy class, i.e. if it's the colvars and biases directly that handle callbacks without the proxy knowing it, it's fine by me. The proxy's role is to manage the interface with the MD engine, but not necessarily with other libraries.

jhenin:
Sounds good. I would like to postpone the architecture changes for a separate PR after this and #462 are merged, to sync those branches and reduce merge conflicts - and because I don't know when I'll have the time for larger rearrangements.

giacomofiorin:
> Sounds good. I would like to postpone the architecture changes for a separate PR after this and #462 are merged, to sync those branches and reduce merge conflicts - and because I don't know when I'll have the time for larger rearrangements.

Exactly! You can see that this was also an unfinished task back in #122.

giacomofiorin:
I'll review this today.

namd/src/colvarproxy_namd.h
```diff
@@ -59,10 +59,10 @@ class colvarproxy_namd : public colvarproxy, public GlobalMaster {
   cvm::real amd_weight_factor;
   void update_accelMD_info();
 
-  void init_tcl_pointers();
-
 public:
 
+  virtual void init_tcl_pointers();
```
giacomofiorin, line 64: Override

src/colvarproxy_tcl.cpp
```diff
@@ -36,8 +35,18 @@ colvarproxy_tcl::~colvarproxy_tcl()
 
 void colvarproxy_tcl::init_tcl_pointers()
 {
-  cvm::error("Error: Tcl support is not available in this build.\n",
-             COLVARS_NOT_IMPLEMENTED);
+  // This is overloaded by NAMD and VMD proxies to use the local interpreters
```
giacomofiorin, line 38: Currently this is only overloaded by NAMD

jhenin: It's also overloaded in VMD, by a function that does nothing.

cmake/CMakeLists.txt
```diff
@@ -13,6 +13,10 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 
 option(WARNINGS_ARE_ERRORS "Treats warnings as errors" OFF)
 
+# message(STATUS "Enabling sanitize-address")
+# add_compile_options(-fsanitize=address)
+# add_link_options(-fsanitize=address)
+
```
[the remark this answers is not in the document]
jhenin: Done

src/colvarproxy_tcl.h
```diff
@@ -10,6 +10,17 @@
...
 
+#if defined(NAMD_TCL) || defined(VMDTCL)
+#define COLVARS_TCL
+#endif
+
+#ifdef COLVARS_TCL
+#include <tcl.h>
+#else
+// Allow for placeholders Tcl_Interp* variables
+typedef void Tcl_Interp;
```
giacomofiorin, line 21: I think that this line most of all is consistent with Tcl philosophy! :-D

giacomofiorin (approved):

PR #434, opened 2021-11-26
More flexible use of boundaries from colvars by grids

### LINKED ISSUE #403: upper/lower_boundary feature blocking histogram with specific grid params
With histogramGrid, a histogram can in principle be defined on a colvar without defined boundaries. However this fails with the message `User-controlled feature "upper_boundary" cannot be enabled automatically`.
The histogram bias should be able to enable it. Maybe making the boundary features dynamic will suffice, however I need to check this for unwanted side effects.

jhenin:
Goal: close #403 

Initialize lower_boundary to 0 and upper_boundary to width.
Re-calculate periodicity of grid after parsing specific config.

Added regression tests.


giacomofiorin:
Hard-coding the data type of the boundaries as scalar numbers could be acceptable in the end, seeing as nobody spent the effort to implement grids for non-scalar types yet 🙄

However, you now have a mismatch with the `colvar::cvc` class still using the more abstract type `colvarvalue` to represent the boundaries. Pointers to those objects are shared between `colvar::cvc` and `colvar`, correctly for now but there is no telling when a nasty error will come up.

After you sort out the failed regtest, please make the types of the boundary objects consistent.

jhenin:
Fair enough.

jhenin:
Tests are good now!

giacomofiorin:
A'ight, the rest looks okay.

PR #435, opened 2021-11-29
Improving Github Actions

HubLot:
Hi,

This a draft for improving the Github Actions of Colvars by making the actions more modularizable. This a copy of #430 

For the first workflow ("Build & CodeQL analysis"), I split it into 3 different jobs (with one being dependant of another).

For the backend workflow, I'm using the "Reusable workflow" feature introduced recently. I created a template workflow ("backend-template.yml") which is in charge of installing, building and testing colvars with a generic backend. Then, I call this workflow in the "test-backends.yml" where I specified the variable which belongs to the workflow tested.

This separate the test procedure and produce independant job for each backend. This increase readability in case of failure.

So far, it works for LAMMPS, NAMD & VMD tests.
Note that due to the NAMD/VMD private repository, Actions concerning these repos won't work on forks ~~(and PRs coming from those forks)~~

To Do:
  - [x]  add sun compiler
  - [x] add gromacs tests


HubLot:
I think the PR is ready to merge now.
I added the sun compilation and the gromacs 2020 tests

jhenin (approved):
Nice. The PR description still says that NAMD/VMD builds won't work on forks, but now they will.

giacomofiorin:
I'm also happy in general with the improved structure, but there are a few things that I'd still like to do:
1. With CodeQL now a separate job, it would make sense to first build the library quickly with the benefit of `ccache`, and then build it a second time via CodeQL going through it. The overhead of the first step would be very small, but it would catch errors more quickly. The `build-legacy` job could reuse the same cache (unfortunately Sun Pro does not support it, so it should go last, before or after CodeQL).
2. For clarity, perhaps it is time to rename `codeql-analysis.yml` to `test-library.yml`? The original name only come because the very first Actions workflow came from a CodeQL template but no other reason.
3. When building the backends concurrently, the cache key pattern is the same for all. This may result often in one backend picking a cache that is up to date for another backend. Can you perhaps test if the following would work in `backend-template.yml`?
```
      - name: Configure Ccache
        uses: actions/cache@v2
        with:
          path: |
            ~/.ccache
            ~/ccache_CentOS7-devel
          key: ${{ runner.os }}-build-${{ job.id }}-${{ github.sha }}
          restore-keys: |
            ${{ runner.os }}-build-${{ job.id }}-
            ${{ runner.os }}-build-
            -build-
```

HubLot:
> Nice. The PR description still says that NAMD/VMD builds won't work on forks, but now they will.
Yes, it should work on PR from fork but not from fork 'alone', meaning actions on forks commit will not succeed. I tried to resolve that but i don't think it's possible yet (due to the `secrets` in the Actions, see [here](https://github.community/t/make-secrets-available-to-builds-of-forks/16166) and [here](https://github.community/t/allow-secrets-to-be-shared-with-forks-from-trusted-actions/16525))

giacomofiorin:
> Yes, it should work on PR from fork but not from fork 'alone', meaning actions on forks commit will not succeed. I tried to resolve that but i don't think it's possible yet (due to the `secrets` in the Actions, see [here](https://github.community/t/make-secrets-available-to-builds-of-forks/16166) and [here](https://github.community/t/allow-secrets-to-be-shared-with-forks-from-trusted-actions/16525))

I don't think that would be a good idea to share the secrets with any forks: that would be equivalent to indiscriminately distributing the code under restrictive licenses. 

Which raises the good point that as others pull into their forks, their tests would fail unless they manually disable the workflows on their fork's GitHub settings. Is there a way to disable the jobs that depend on the presence of certain secrets?

Alternatively, it may just be easier to run the backends' jobs when in the "main" repo on the Colvars org, but disable those for the forks altogether.

HubLot:
thanks @giacomofiorin  for the feedback.
I agree with all points raised. I'll do the modifications proposed.


> Which raises the good point that as others pull into their forks, their tests would fail unless they manually disable the workflows on their fork's GitHub settings. Is there a way to disable the jobs that depend on the presence of certain secrets?

I was looking to that also, I'll continue the digging. Maybe the `if` condition could apply to `secrets` also.

> Alternatively, it may just be easier to run the backends' jobs when in the "main" repo on the Colvars org, but disable those for the forks altogether.

True but having the Actions run on PRs is very convenient

giacomofiorin:
> True but having the Actions run on PRs is very convenient

I think those would always work as long as they target the main repository, by matching the `pull_request_target` and running with the secrets, too. It's the `push` condition that wouldn't have access to the secret if the push happens in a fork.

Actually, I'm also processing now that you lifted the restriction that the PR or push must be toward the `master` branch: so any push would be subject to running workflows. I think that would lead to a lot of failed tests in those branches where we have work in progress. Can you restrict the `push` to match `master` only?

giacomofiorin:
@jhenin I accidentally clicked on re-request review but didn't meant to, sorry.

.github/workflows/test-backends.yml
```diff
@@ -21,6 +21,7 @@ jobs:
   namd:
     name: Test NAMD
     uses: Colvars/colvars/.github/workflows/backend-template.yml@test_ci
+    if: ${{ secrets.PULL_NAMD_KEY }}
```
jhenin, line 24: :+1:

jhenin: That's because I liked it

HubLot:
An update on the PR:

 - I renamed the codeql-analysis file
 - I modified the ccache step as suggested by @giacomofiorin 
 - To disable jobs which depends on secrets on forks, I added this statement to the job : `if: ${{ github.repository_owner == 'Colvars' }}` which work on my fork (see [here](https://github.com/HubLot/colvars/actions/runs/1531077657)).  This is more or less what @giacomofiorin  suggested. I hope this will not prevent Actions from fork's PRs but my guess is it shouldn't thanks to `pull_request_target` (which run the actions on the base branch). But this is hard to test without this PR being merged.

>Actually, I'm also processing now that you lifted the restriction that the PR or push must be toward the master branch: so any push would be subject to running workflows. I think that would lead to a lot of failed tests in those branches where we have work in progress. Can you restrict the push to match master only?

It was on purposed I lifted the restriction. For me, it's not a big deal to have a CI working on WIP-branches and I think it will be convenient too to have the CI running on those branches to spot more quickly the errors.
If you think it's too much, one can skip the CI with keywords in the commit ([here](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/)) or we can think of pattern-name for branches to be excluded or included in the CI (https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#on)

HubLot:
>1. With CodeQL now a separate job, it would make sense to first build the library quickly with the benefit of ccache, and then build it a second time via CodeQL going through it. The overhead of the first step would be very small, but it would catch errors more quickly. The build-legacy job could reuse the same cache (unfortunately Sun Pro does not support it, so it should go last, before or after CodeQL).

I didn't modify this part because building twice the colvars lib will require to duplicate some steps of the build job (installation of tcl and lepton) and the whole job "Build Colvars & CodeQL Analysis" is rather fast: 4min with 2m43 for the CodeQL analysis which is done after the build and which will be cancelled if the build fails. So it's already fast and I'm not sure having a duplication is best here.

jhenin:
@giacomofiorin I share your concern that some particularly "dirty" work in progress branches will trigger an avalanche of failed tests. Now, knowing that we can opt out on a temporary basis just by adding [skip ci] to a commit message in the push, that seems sufficient to me.

giacomofiorin:
> I didn't modify this part because building twice the colvars lib will require to duplicate some steps of the build job (installation of tcl and lepton) and the whole job "Build Colvars & CodeQL Analysis" is rather fast: 4min with 2m43 for the CodeQL analysis which is done after the build and which will be cancelled if the build fails. So it's already fast (and `build-legacy` too with ~35s) and I'm not sure having a duplication is best here.

Fair enough, but FYI the legacy build doesn't need Lepton at this stage.

It is there pretty much to appease the VMD build system, and counts squarely in the column of "extra work that we need to do to support VMD builds for systems that have all but extinct".

giacomofiorin:
> @giacomofiorin I share your concern that some particularly "dirty" work in progress branches will trigger an avalanche of failed tests. Now, knowing that we can opt out on a temporary basis just by adding [skip ci] to a commit message in the push, that seems sufficient to me.

The no. 1 enemy of the continuous-integration is false-positive test results overtaking true-positive ones. If you are consistent with using that tag, then it's all good.

Note that we have quite a few stale branches. If you happen to continue work on some of them, please also consider running the tests locally before pushing a new commit. The beauty of using a container for the numerical tests is that you don't need the particular Linux OS that GitHub uses.

giacomofiorin:
Ha! Gotcha! You wrote `main` in the first draft and then edited it!

giacomofiorin:
One important thing: I pushed some changes to an open PR (which is likely to happen when amending it to fix errors caught by CI) and that triggered double tests.

Is there a way to avoid the duplication? Say that I don't care to this level about the electricity that Azure consumes, but I would prefer to read everything important just once.

HubLot:
I looked into it and it seems there is no simple way to avoid duplication... (besides playing with `if` statements). It's discussed [here](https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012) and [here](https://github.community/t/duplicate-trigger-on-the-same-commit-push-and-pull-request/18473)

giacomofiorin:
Ok, the second discussion mentions a rather advanced tools that may be useful for other purposes, but the `if` conditional from the first discussion seems to fit the bill. After a few syntax mistakes (I'm writing code for Actions like many people write their MD simulation inputs) it seems to be running. We'll see later on the open PRs.

Should it not work, I also shortened the workflow names to "Test library" and "Test backends" simply, because now that there are many jobs the summary in the little balloon becomes more useful to locate which one is failing and then click on it directly.

giacomofiorin:
Oh and BTW we don't have yet caches specific to each backend. These are the timings of a successful job from last week.
https://github.com/Colvars/colvars/actions/runs/1536671721
image

All backends are using the cache of the job that finishes first (which is normally VMD because the tests are fewer).

giacomofiorin:
Ok that's more like it :-)
image

jhenin:
Now the tests on pull_request_target are skipped, but only some of them:
image

giacomofiorin:
That PR still looks like it's based on top of an earlier `master`. I would test on a more current one.

giacomofiorin:
It may also have to do with the if syntax, since for NAMD and VMD there are
two conditions that need to be merged.

giacomofiorin:
But basically, this is showing an additional problem: skipped tests still show up in the summary balloon, which really defeats the main purpose that I was trying to achieve. It looks like it only makes a difference in terms of how many Actions minutes you want to spend. But we still have *17 tests* to scroll over...

I guess I could just simply up on trying to keep the list short, and rely on the website to point to the failed tests more quickly.

The main goal for me at  this point is maintaining correct caches, so that even if there are redundant tests they come in quickly. I'll take a look again after the current PRs are merged.

PR #438, opened 2021-12-08
Handle biases in VMD Colvars Dashboard

jhenin:
I'm not completely happy with the interface as it is. I considered several options, such as putting the list of biases directly below the list of colvars and not in a tab, but then I'm unsure how to distribute the respective Action buttons.

I think the biases will remain secondary in VMD. The main feature I'm looking for here is to be able to open a config file, edit the colvars, and save everything without losing any information. In that sense the list of biases is mostly for information purposes.

- [x] expose table of biases
- [x] implement energy timeline (and other actions?)
- [x] deal with (in)compatible versions of the library throughout
- [x] document
- [x] do not write unit keyword twice if specified in input
- [x] do not write auto-generated harmonicWalls (will require parsing colvar cfg for wall options - piece of cake)

### DIFF
--- a/doc/colvars-refman-main.tex
+++ b/doc/colvars-refman-main.tex
@@ -246,10 +246,8 @@ If variables are modified, added or deleted by an external script, hit ``Refresh
 Starting the Dashboard also enables trajectory animation using the left/right arrow keys within VMD's graphical window.
 Atomic coordinates can be modified using VMD's ``Mouse/Move'' functions, and the Colvars module can then be updated by pressing F5 directly from the graphical window.
 
-A dropdown list allows for changing the current unit system if no variables are defined.
-If some variables are already defined, it is recommended to edit the configuration for all of
-them at once (eg. pressing Ctrl-a, then Ctrl-e), checking that all quantities are expressed in the desired set of units, and adding the
-\texttt{units} keyword to the general parameters, outside of \texttt{colvar \{\}} blocks (\ref{sec:units}).
+A dropdown list allows for changing the current unit system if no variables are defined (see section~\ref{sec:units}).
+When changing units, it is strongly recommended to check the configuration for all colvars to ensure that all quantities are expressed in the desired set of units.
 
 The Colvars Module interacts with one VMD Molecule at a time.
 At the bottom of the \textit{Actions} tab, a  dropdown lets the user change which VMD molecule is associated with the Colvars module.
@@ -260,9 +258,9 @@ Auto-updating selections (see below) can be used to adapt the colvar definitions
 
 \cvsubsec{Loading / Saving configuration files}{sec:dashboard_load_save}
 
-This saves the configuration of all defined collective variables to a file.
-Neither biases, nor general parameters of the Colvars module (except the units) are saved: editing them is beyond the scope of the Dashboard.
-We recommend keeping them in separate configuration files, and reading them separately in biased MD simulations.
+This saves the current configuration of the colvars Module to a file.
+This includes comments found in the input, general parameters of the Colvars Module, the collective variables themselves, and collective variable biases.
+The resulting configuration file can be read by Colvars, either back in VMD, or within a supported MD engine.
 
 \cvsubsec{The configuration editor}{sec:dashboard_config_editor}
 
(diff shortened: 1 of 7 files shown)


doc/colvars-refman-main.tex
```diff
@@ -328,6 +326,19 @@ \subsection*{Overview}
 Parameters for fine-tuning these visualizations can be set in the \textit{Settings} tab.
 
 
+\cvsubsec{Monitoring colvar biases in the Dashboard}{sec:dashboard_biases}
+
+The main Dashboard window features a "Biases" tab that contains a list of currently defined biases, their current energy, and the colvars that they depend upon.
+A button enables creating a Timeline plot with the energies of currently selected biases, similar to the colvar Timeline plot.
+Only one plot can be active at a time.
+
+Note that the energy of history-dependent biases such as ABF or metadynamics may not be meaningfully computed in VMD.
```
giacomofiorin, line 335: Great point, good to have here.

vmd/cv_dashboard/cv_dashboard.tcl
```diff
@@ -197,8 +203,28 @@ proc ::cv_dashboard::apply_config { cfg } {
       }
     }
   }
-  # Overwrite old map
+  # Overwrite old covlar map
```
giacomofiorin, line 206: Little typo

giacomofiorin (approved):
Overall very good! 

One minor suggestion, the button "New" refers to a new colvar, but there is always the possibility in the future to have bias templates, so maybe it should say "New colvar" to be clear from the get-go?

Another nitpicky detail is that the list of colvars may sometimes get too long to fit in the window, and there is no horizontal scroll. I don't know whether you prefer to have a button that expands on the full list (as you have for vector CVs), or use a scroll bar, or nothing at all. I'm good either way. After all, that list may be long but is *static* and can be queried also from the command line if needed.

giacomofiorin:
I thought to upload a screenshot for future record:
image
This was taken with a local build.

With a UIUC build I can't use the new feature yet, but it definitely highlights the importance of using updated menu fonts:
image

giacomofiorin:
BTW the biasing energy of metadynamics can be queried if a state file is loaded first:
```
vmd > cv bias metadynamics1 energy
0
vmd > cv load AutoDiff/test.restart.colvars.state
colvars: ----------------------------------------------------------------------
colvars: Loading state from file "AutoDiff/test.restart.colvars.state".
colvars: This state file was generated with version 20XX-YY-ZZ
colvars:   Restarting collective variable "one" from value: 3.21426
colvars:   Restarting metadynamics bias "metadynamics1" from step number 40.
colvars:   Read 0 hills.
colvars: ----------------------------------------------------------------------
vmd > cv bias metadynamics1 energy
0
vmd > cv bias metadynamics1 update
5.25534912996650e-01
vmd > cv bias metadynamics1 energy
0.525535
```

Successive calls to update will not change this value, because the step counter is not incremented in VMD.

Perhaps you could clarify to say that the history-dependent biases may not be consistent with their values during a previous simulation, but only with the contents of the state file?

jhenin:
> One minor suggestion, the button "New" refers to a new colvar, but there is always the possibility in the future to have bias templates, so maybe it should say "New colvar" to be clear from the get-go?

Done.

> Another nitpicky detail is that the list of colvars may sometimes get too long to fit in the window, and there is no horizontal scroll. I don't know whether you prefer to have a button that expands on the full list (as you have for vector CVs), or use a scroll bar, or nothing at all. I'm good either way. After all, that list may be long but is _static_ and can be queried also from the command line if needed.

Both the window and the table column can be resized, potentially accommodating a lot of colvar names.

> With a UIUC build I can't use the new feature yet,

Strange, you should be able to. What doesn't work?

> BTW the biasing energy of metadynamics can be queried if a state file is loaded first:

Nice, that is true of ABF too.

giacomofiorin:
> Both the window and the table column can be resized, potentially accommodating a lot of colvar names.

Cool, maybe write it in the section `sec:dashboard_main`? Many of the most common VMD windows can't be resized e.g. VMD's main menu, and depending on which window manager you're using the resizing 

> Strange, you should be able to. What doesn't work?

On the latest pre-compiled build [1.9.4a55](https://www.ks.uiuc.edu/Research/vmd/vmd-1.9.4/files/alpha/vmd-1.9.4a55.bin.LINUXAMD64-CUDA102-OptiX650-OSPRay185.opengl.tar.gz), the bias sub-command `type` that I added to support this PR is missing there, and I get this:
image
Maybe you can `catch` that and not create the bias tab if there is an error?

jhenin:
> Maybe you can catch that and not create the bias tab if there is an error?

This is already `catch`ed, it works for me in VMD 1.9.3 (using a fallback method, which only works if biases are created through the Dashboard and not externally, so we have a chance to parse the complete config). Are you sure you are testing the current HEAD revision?

giacomofiorin:
Yes, I ran `update-colvars-code.sh` with argument the `plugins` sub-folder of the VMD installation tree.

Indeed it works for me with 1.9.3 also, but not with 1.9.4a55.

Side note: I forgot to bring up with the review that since you bumped the version to 1.4, the installation folder should be renamed accordingly. Both the `Makefile` for embedding in the VMD plugin tree and the `Makefile.local` for a user-driven installation are still at 1.3. I don't have any other copies of the Dashboard externally to VMD, so I don't think there is a conflict, but better fix the confusion.

jhenin:
Ok this happens because the code uses "cv bias name help type" to probe whether the command exists, but the scripting command does not return an error when that happens. I modified it to do so in 4762f1c4891e28095871068c73801630cd6410b1, but of course that will only be useful for future uses...

So I need a better logic, like the colvars version string.

giacomofiorin:
FYI d65aa36b works for me.

PR #442, opened 2021-12-21
Persistent bias count

### LINKED ISSUE #441: Non-unique bias names
If two "anonymous" (auto-named) biases of the same type are defined (eg. harmonic1 and harmonic2), and harmonic1 is deleted, then re-created, the new bias is assigned number 2, which creates two biases with the same name. For consistency, I see two possibilities:
- do not decrease the counter when biases are deleted, and assign ever-increasing numbers in order of creation (simple);
- maintain he list of "used" bias numbers and assign new biases the first free number

jhenin:
Fixes #441

### DIFF
--- a/src/colvarbias.cpp
+++ b/src/colvarbias.cpp
@@ -41,6 +41,7 @@ colvarbias::colvarbias(char const *key)
 
 int colvarbias::init(std::string const &conf)
 {
+  name = bias_type + cvm::to_str(rank);
   colvarparse::set_string(conf);
 
   size_t i = 0;
--- a/src/colvarbias.cpp
+++ b/src/colvarbias.cpp
@@ -19,12 +19,13 @@
 
 colvarbias::colvarbias(char const *key)
 {
-  bias_type = to_lower_cppstr(key);
+  bias_type = colvarparse::to_lower_cppstr(key);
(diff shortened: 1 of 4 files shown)


jhenin:
Thanks @giacomofiorin, this does fix the issue as I saw it in the context of the Dashboard.

You mentioned simplifying the void pointer conversions? Other than that, this looks good to me.

giacomofiorin:
Well, I had already started opening a PR, but didn't finish yet...  I wanted to add a couple of doc improvements, but those I can also push straight in.

The only way that I was considering removing the `void *` was to put it as a correctly-typed member of the usage class, or something else. But that's only marginally better.

Since you opened the PR from the branch, you obviously want this soon, so let's merge it.

jhenin:
Sometimes speed is good too :-)

giacomofiorin:
Just pushed the doc change. I had realized that `plot_colvars_traj.py` was actually not mentioned in the doc, but it's useful to do so now that more processing tools rely on trajectory data being fed as NumPy arrays.

For example, @fabsugar has implemented his new unbiasing method here: https://github.com/Faraldo-Gomez-Lab-at-NIH/FCAM

PR #451, opened 2022-02-15
Better management of repository secrets

giacomofiorin:
This PR attempts to simplify the management of the private repositories for NAMD and VMD in automated tests. These repositories are private due to the restrictive licenses of the two packages: however, people who have accepted their license terms, are welcome to use the private repositories listed under this org using ad-hoc SSH keys.

Using these keys, it is no longer necessary that the Actions workflows are run within this repository. Therefore, I modified the tests to stop using the `pull_request_target` trigger, and use `pull_request` instead. I hope it works, 'coz I haven't read the GitHub doc in detail. (I normally do, but not when the only purpose is to comply to a restrictive license: just poking around seems enough effort for that).

### DIFF
--- a/.github/workflows/backend-template.yml
+++ b/.github/workflows/backend-template.yml
@@ -93,6 +93,7 @@ jobs:
           sudo dpkg -i singularity_3.8.5-2_amd64.deb
 
       - name: Get container image of backends' dependencies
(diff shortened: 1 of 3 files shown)


giacomofiorin:
The PR's side seems okay, there is still a reliance on the backend template's copy on `master`, for which changes will take effect only upon merging.

PR #452, opened 2022-02-16
Add GROMACS 2021 to automated build tests

giacomofiorin:
Also use a PR to test that the correct triggers are picked up.

### DIFF
--- a/.github/workflows/test-backends.yml
+++ b/.github/workflows/test-backends.yml
@@ -13,6 +13,7 @@ env:
   PULL_VMD_PLUGINS_KEY: secrets.PULL_VMD_PLUGINS_KEY
 
 jobs:
+
   lammps:
     name: LAMMPS
     uses: Colvars/colvars/.github/workflows/backend-template.yml@master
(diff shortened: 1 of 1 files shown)


giacomofiorin:
Workflows are okay and library tests are also working on all backends. There are some failures on the interface tests:
- `000_MPI` fails because the build is not MPI based :-)
- `010_multi-sim` also fails because of script issues:
```
Entering 010_multi-sim ...Run special script.
cp: cannot stat ‘a/test.colvars.traj’: No such file or directory
cp: cannot stat ‘a/test.colvars.state’: No such file or directory
```
@HubLot Can you look into making them work with the current environment? (We can add a MPI-based build recipe to the tests if you think that it would be worth it).

Neither issue is related to the scope of this PR => merging.

giacomofiorin:
@HubLot @jhenin I made some minor changes to run select GROMACS interface tests based on how the build was made. But despite that, I couldn't reproduce the `000_tmpi` test any better than 1.0e-4 relative error. I don't know how the reference data was generated, but with automated tests in place I didn't want to relax the tolerance, so I just disabled this test specifically.

giacomofiorin:
Nah, some of the tests that pass on my desktop still fail. My guess is that this is a difference due to which SIMD instructions are built there vs. on the Azure instances. I just have to give up and re-disable the interface tests. Please do look into whether they are worth maintaining, otherwise they should be deleted.

I'll turn on GROMACS' own tests, which pass now thanks to #450.

PR #462, opened 2022-03-06
Allow running functionality tests without MD engine

giacomofiorin:
Existing (regression) tests already provide a ton of coverage, particularly when it comes to interoperability with the existing engines. But they are also more expensive to run, and the dependency on the engines may be problematic when these are undergoing very active development.

Most importantly, the following two important tasks have little to do with the MD engine:
- Ensuring that logical flags are consistent, e.g. dependency checking.
- Verifying numerical correctness, e.g. energy-gradients correspondence.

This PR is an initial implementation of the first task, which reads the existing configuration files into an environment without having actual atoms being selected or forces applied to them or measured from them. A stub implementation of the proxy class allows for a full initialization without raising errors, and is thus exercising both the parser and dependency solver. A realization of the second task would require a more functional proxy implementation, and will be part of a later PR.


jhenin:
This is already looking quite good! What is missing to remove the draft status?

jhenin:
This is already looking quite good! What is missing to remove the draft status?

giacomofiorin:
> This is already looking quite good! What is missing to remove the draft status?

Not much, really. When I last touched this branch there were GROMACS-related changes not merged in yet. I just updated, and I'm fine with merging after the tests run.

PR #464, opened 2022-03-07
Fix #463 using more fine-grained features

### LINKED ISSUE #463: Feature f_cv_gradient tries to collect gradients even when not needed, or unavailable.
In 9d41a2682f0f5841ecbfa0bf904cb858b18983d7, a condition matching `f_cv_collect_gradient` was substituted with `f_cv_gradient` to allow introspection from the Dashboard, but the two are not the same.

When gradients are collected during a MD simulation, this results in additional processing, because `f_cv_gradient` is on whenever forces are applied. Furthermore, for those variables where part of the computation is off-loaded to the MD engine you get a bunch of warnings like this:
```
Warning: colvar components communicated no atom IDs.
```

The original condition should be reverted, and the Dashboard should use lines like this to request the gradients' collection:
```
cv colvar <xi> set collect_gradient 1
```

jhenin:
- partially revert https://github.com/Colvars/colvars/commit/9d41a2682f0f5841ecbfa0bf904cb858b18983d7
- create atom_list feature that cascades through the ag, cvc and cv levels
 (side-effect is performed at cv level but availability is checked at ag level)
- use in Colvars Dashboard to control exactly as needed

### DIFF
--- a/src/colvar.cpp
+++ b/src/colvar.cpp
@@ -958,7 +958,7 @@ void colvar::do_feature_side_effects(int id)
     case f_cv_total_force_calc:
       cvm::request_total_force();
       break;
-    case f_cv_atom_list:
+    case f_cv_collect_atom_ids:
       // Needed for getting gradients vias collect_gradients
       // or via atomic forces e.g. in Colvars Dashboard in VMD
       if (atom_ids.size() == 0) {
@@ -1103,14 +1103,14 @@ int colvar::init_dependencies() {
     init_feature(f_cv_collect_gradient, "collect_gradient", f_type_dynamic);
     require_feature_self(f_cv_collect_gradient, f_cv_gradient);
     require_feature_self(f_cv_collect_gradient, f_cv_scalar);
-    require_feature_self(f_cv_collect_gradient, f_cv_atom_list);
+    require_feature_self(f_cv_collect_gradient, f_cv_collect_atom_ids);
     // The following exclusions could be lifted by implementing the feature
     exclude_feature_self(f_cv_collect_gradient, f_cv_scripted);
     exclude_feature_self(f_cv_collect_gradient, f_cv_custom_function);
     require_feature_children(f_cv_collect_gradient, f_cvc_explicit_gradient);
 
-    init_feature(f_cv_atom_list, "atom_list", f_type_dynamic);
-    require_feature_children(f_cv_atom_list, f_cvc_atom_list);
+    init_feature(f_cv_collect_atom_ids, "collect_atom_ids", f_type_dynamic);
+    require_feature_children(f_cv_collect_atom_ids, f_cvc_collect_atom_ids);
 
     init_feature(f_cv_fdiff_velocity, "velocity_from_finite_differences", f_type_dynamic);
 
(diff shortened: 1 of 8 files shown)


giacomofiorin (commented):
Generally good, although I would prefer `f_*_collect_atom_list`, to signal more clearly that it's a companion to `f_*_collect_gradient` and not something that has to do with how the objects are defined.

Do I understand correctly that with `f_*_collect_atom_list` activated, the list of atom IDs is always exported, regardless of whether they are processed on rank 0?

jhenin:
> Generally good, although I would prefer `f_*_collect_atom_list`, to signal more clearly that it's a companion to `f_*_collect_gradient` and not something that has to do with how the objects are defined.

Good idea, will do that.

> Do I understand correctly that with `f_*_collect_atom_list` activated, the list of atom IDs is always exported, regardless of whether they are processed on rank 0?

The only effect is to call `build_atom_list` at the CV level, that is, always on rank 0. But this can only be triggered if the atom group feature can be enabled, which requires that it be processed on rank 0 (exclusion with scalable_com). Speaking of which, it occurs to me that the exclusion should involve f_ag_scalable instead, to be more future-proof. I'll change that as well.

jhenin:
I see that now, `f_*_scalable` is always enabled if `f_*_scalable_com` is enabled. Would it make sense to make  `f_*_scalable_com` require ` f_*_scalable`, so that happens automatically? That would simplify this construct, where I think the logic is not very clear for the uninformed reader:
```
    enable(f_ag_scalable_com);
    enable(f_ag_scalable);
```
Then other, future types of "scalable" features would also require `f_*_scalable`.

giacomofiorin:
I'm okay with declaring a requirement between those two feature flags. If I remember correctly (it was a while ago) I didn't think I had an established way of doing that at the time, but there is one now.

Regarding VMD versions, I'm fine with requiring a later one. This is not a big deal as long as the alpha releases come out relatively frequently. The right way to do this would be to also distribute the Colvars library as a C++ plugin, but we cannot do that because John does not allow plugins to access VMD objects directly. And while I certainly do not agree with that restriction, dynamic linkage is always more cumbersome for users than static.

giacomofiorin:
I tried to do that in 646f1c1c. I still don't remember exactly, but for sure there wasn't a `dynamic` feature property, which is actually what would be appropriate to use here IMO.

jhenin:
Well, things are looking nicer now!

jhenin:
Well, it turns out I have at least one user who uses the Dashboard on a Mac and gets a fresh Dashboard plugin from this repo. Merging this would break the master branch for him. Maybe add a tag with the compatible VMD version?

giacomofiorin:
You mean a Git tag or a note in the README? I'm prefer the latter, especially for the Dashboard, which people can in theory install independently.

jhenin:
Possibly both: a note in the README with links to the relevant tags?

jhenin:
Sounds good!

jhenin:
Draft release: https://github.com/Colvars/colvars/releases/edit/untagged-c2249507808a446fdbbe

giacomofiorin:
Draft looks good, I made a couple of tweaks to minimize confusion between the plugin, the library, and the packaged application.

giacomofiorin:
I so wish we could just put a VMD tarball there!

Issue #460, opened 2022-02-23
Move library to new location in GROMACS tree


giacomofiorin:
The current location in the [patched releases](https://github.com/Colvars/gromacs) is `src/gromacs/colvars`, which is really not ideal. While it makes it a bit simpler to build the release, the code organization should clearly reflect where the components are developed. A more suitable location would be `src/external/colvars`.

I put a tentative change for the 2021 patch in [this branch](https://github.com/Colvars/gromacs/tree/release-2021-colvars-in-external). I can re-generate the corresponding patches, but @HubLot @jhenin it's best to get your input first, particularly considering the MDModules version of the interface, which I didn't test.

The core of it is the patch below, which allows including headers and linking the library without need to put it under the `gromacs` folder.
```
diff --git a/src/gromacs/CMakeLists.txt b/src/gromacs/CMakeLists.txt
index a4430e9dd6..81a764b554 100644
--- a/src/gromacs/CMakeLists.txt
+++ b/src/gromacs/CMakeLists.txt
@@ -142,6 +142,10 @@ if (WIN32)
 endif()
 list(APPEND libgromacs_object_library_dependencies thread_mpi)
 
+file(GLOB COLVARS_SOURCES ${PROJECT_SOURCE_DIR}/src/external/colvars/*.cpp)
+add_library(colvars OBJECT ${COLVARS_SOURCES})
+list(APPEND libgromacs_object_library_dependencies colvars)
+
 configure_file(version.h.cmakein version.h)
 if(GMX_INSTALL_LEGACY_API)
   install(FILES
@@ -189,6 +193,8 @@ else()
     add_library(libgromacs ${LIBGROMACS_SOURCES})
 endif()
 
+target_include_directories(libgromacs PRIVATE ${PROJECT_SOURCE_DIR}/src/external/colvars)
+
 # Add these contents first because linking their tests can take a lot
 # of time, so we want lots of parallel work still available after
 # linking starts.
```

HubLot:
Hi Giacomo,

The changes seems correct to me. We will need to modify the `update_colvars_code.sh` too
2 points though:
  - The  2 includes of colvars_proxy in `runner.cpp` and `sim_util.cpp` should be placed before the gromacs include as it's done elsewhere for similar external codes.
  - At some point, we may need to create a special cmake file under their `cmake` folder for colvars and lepton. We should discuss with them the proper way to do this. Anyway it's beyond the scope of this issue imo.

giacomofiorin:
[...]

PR #465, opened 2022-03-08
Update GROMACS patches to place Colvars in src/external

giacomofiorin:
Implement #460

For now I'd rather not do the extra work of updating the 2020 patch, too (unless there are reasons why that would be necessary).


giacomofiorin:
@HubLot The tests pass. Since you let me re-generate the GROMACS 2021.x patch, I did that with `git diff` on a GROMACS worktree, and as a result the `patch` command line is different (`-p1` instead of `-p0`). This means committing to formatting future patches for e.g. 2022 release series in the same way. Can you confirm if this is okay with you?

HubLot (approved):
Great job!
I'm okay using `git diff` and `patch -p1`

giacomofiorin:
For simplicity, and because 2020 is still maintained, I just extended the change of path to the 2020 series.

As agreed, I'll also stop auto-updating the 2018 release series at https://github.com/Colvars/gromacs

PR #467, opened 2022-03-11
Fix MSVC warnings

jhenin:


giacomofiorin:
@jhenin Thanks for fixing a bunch of warnings. I then raised the level to /W3, but then a ton more type conversion warnings popped up, all of which were [4267](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4267). I silenced those with command line flags, because we're already using a 64-bit integer to index an n-dimensional grid's buffer (I fixed one instance), and I find it extremely unlikely that we would overflow a 32-bit integer in the other use cases. 

So I started making the warnings become errors with MSVC as well. With the exception of silencing 4267, using `/W3` is already more thorough than LAMMPS and GROMACS are doing.

It may be worthwhile checking out some of the 4267 warnings that I silenced: to do so, you can remove the `/wdxxxx` options and push to a branch (best remove also `/WX` so you see all warnings).

Ok with merging this when the CI jobs are done?

giacomofiorin:
P.S. 4244 is a similar warning, which we don't have but Axel is also silencing in LAMMPS. I'm neutral about keeping it.

P.S. 2 If you want to troubleshoot the 4267s, it may be helpful to use pragmas in headers (like in `colvarbias_restraint.h`) to reduce the noise level.

jhenin:
I can live with the 4267s for now. Merging this, as it's already a good step forward.

PR #474, opened 2022-04-29
Allow normalizing eigenvector vector components to deal with unit change

### LINKED ISSUE #471: Issue with Eigenvector CV units in GROMACS when vector is declared from .xyz file
Dear colvars devs,

I have encountered a possible unit-related bug. When using an eigenvector CV with GROMACS, I get values that are 10 times lower than they should be. 
I am aware that in GROMACS the projection onto an eigenvector will be expressed in nm; what I mean here is that the value returned by GROMACS/colvars is 100 times smaller than the value computed in Angstroms by an independent program (whereas it should only be 10 times smaller). 

I am using a .xyz file to declare the coordinates of the eigenvector (with vectorFile). According to the manual, .xyz files are always considered to be in Angstroms. Therefore I think what happens is that the entries of this .xyz are assumed to be in Angstroms and are rescaled (divided by 10) by GROMACS. If so, this is a bug, because after this rescaling the norm of the eigenvector will no longer be equal to 1. As a fix, one could either remove this rescaling specifically for eigenvector CVs, or perhaps add an automatic normalization step for the eigenvector. 

Another remark: in the file `colvarcomp_distances.cpp`, line 1406, the instruction:

`cvm::log("The norm of the vector is |v| = "+cvm::to_str(eigenvec_invnorm2)+".\n");`

prints, as far as I understand, the inverse of the vector square norm (1/|v|²) rather than |v|. This is harmless with normalized vectors because all these numbers are equal to 1. But in my case, I obtain: 
`The norm of the vector is |v| = 100.73.` 

from the colvars log, which is consistent with the norm of my vector now being (approximately) 0.1 instead of 1 (1/0.1² = 100). 

Could you please investigate? I observed the bug in Gromacs-2021.5 with colvars version Feb9-2022 and Feb20-2022. 
Thank you very much ! 
Florian

giacomofiorin:
Fixes #471.

I also added some regression tests that were missing, and at least for NAMD I tried to use a much older version if the code allowed it.

@flblanc Can you see if the feature works for you? You would need to patch a vanilla GROMACS yourself, if there are any issues post a comment and I can create a temporary branch on the GROMACS fork.


jhenin:
Thanks @giacomofiorin 
Your comment there https://github.com/Colvars/colvars/issues/471#issuecomment-1095037472 was very true: we need two types of normalization depending on the context, and they are mutually incompatible. They are now implemented and documented. I also added a warning in the doc regarding the unit conversion issue.

giacomofiorin:
@jhenin I edited the doc to clarify that the units are those of the MD engine (we don't want folks to be confused about which units are being used by Colvars).

I also amended the tests' reference output after you corrected for the missing sqrt().

Ok to merge if the tests pass?

jhenin:
Yup, I just wanted to converge on the functionality before regenerating the test outputs. Good to go for me, pending the test results.

giacomofiorin:
@flblanc The `master` branch should have the needed change (it didn't until the PR was merged). You can check for the Colvars version to be `2022-05-02`, as reported when using it or in the file `src/colvars_version.h`.

The renaming of GROMACS' own `master` branch broke the cron script that I used to keep it up to date. I had tried to preemptively work around it, but I did a different mistake. I'll update the patched GROMACS versions shortly, so you shouldn't even need to patch GROMACS yourself.

PR #475, opened 2022-05-03
Remove fatal error condition

jhenin:
- replaced erroneous fatal error conditions with regular errors
- as a result, the error code FATAL_ERROR becomes unused
- simplify the source by removing unused member data and code


src/colvarmodule.h
```diff
@@ -40,9 +40,7 @@ You can browse the class hierarchy or the list of source files.
 #define BUG_ERROR       (1<<3) // Inconsistent state indicating bug
 #define FILE_ERROR      (1<<4)
 #define MEMORY_ERROR    (1<<5)
-#define FATAL_ERROR     (1<<6) // Should be set, or not, together with other bits
-//#define DELETE_COLVARS  (1<<7) // Instruct the caller to delete cvm
-#define COLVARS_NO_SUCH_FRAME (1<<8) // Cannot load the requested frame
+#define COLVARS_NO_SUCH_FRAME (1<<6) // Cannot load the requested frame
```
giacomofiorin, line 43: Not directly relevant to this PR, but these have rather generic names that could use being wrapped in a namespace. Bringing it up in case you have any idea.

giacomofiorin (approved):
I'm good with this. I'm adding a couple updates for the metadynamics ones.

src/colvarmodule.h
giacomofiorin: Ok with old school. Come to think, we may want to move them to a different header than `colvarmodule.h`, so that the calling code can use them without needing to define the whole main class.

giacomofiorin: Macros renamed. About moving them, it looks more work than I'm willing to do at this moment: not a high priority, since the renaming already takes care of clashes.

giacomofiorin:
@jhenin I can't ask you to "review" this last change since you opened the PR, but please take a look.

jhenin:
Yup!

PR #481, opened 2022-05-16
Make file-reading operations of input data abstractable

giacomofiorin:
This change is similar to what done [earlier](https://github.com/Colvars/colvars/commit/61e9fe6fd5e07c0c9f5eda46931dd00af9af76a0) for `std::ofstream` (whose implementation was not working correctly years ago for some platform that NAMD was running on at the time).

This PR removes explicit uses of `std::ifstream` throughout most of the code, using instead its base class. Initialization and maintenance of a `std::istream` object is done by 2-3 additional proxy functions, which also take care of error checking, leading to significant consolidation of duplicated code. The only remaining exceptions are for use cases where it is clear that the input is read from files that was written by Colvars itself as the _output_ of a simulation.

In the current design, stream objects are proxy members, and therefore access to them is restricted to the first thread only. There may be a need to expand this in the future, but I do not see it at the moment.

@HanatoK In the neural network class I removed a couple of exceptions, since the corresponding error check is done by code that should support VMD, but it should be okay to reintroduce them later.


src/colvar_UIestimator.h
```diff
@@ -687,45 +688,53 @@ namespace UIestimator {
...
-                std::ifstream count_file(count_filename.c_str(), std::ios::in);
-                std::ifstream grad_file(grad_filename.c_str(), std::ios::in);
+                std::istream *count_file_p =
+                    proxy->input_stream(count_filename, "count filename");
+                std::istream *grad_file_p =
+                    proxy->input_stream(grad_filename, "gradient filename");
 
-                count_file >> sharp >> dimension_temp;
-                grad_file >> sharp >> dimension_temp;
+                if (count_file_p != NULL && grad_file_p != NULL) {
```
jhenin, line 696: Could this test be reversed and the function exit upon failure, as done in other instances? This would avoid wrapping most of the function in the test body, reducing nesting depth and improving readability.

src/colvar_neuralnetworkcompute.cpp
```diff
@@ -146,13 +161,15 @@ void denseLayer::readFromFile(const std::string& weights_file, const std::string
...
     }
+    proxy->close_input_stream(weights_file);
+
     // parse biases file
-    std::ifstream ifs_biases(biases_file.c_str());
-    if (!ifs_biases) {
-        throw std::runtime_error("Cannot open file " + biases_file);
+    std::istream *ifs_biases = proxy->input_stream(biases_file, "biases file");
+    if (ifs_biases == NULL) {
+        return;
```
jhenin, line 169: My understanding is that here the return value is a pointer solely to be able to track errors through null pointers. Instead of a pointer, could input_stream() return a reference to an istream that is marked as bad (`stream.clear(ios::badbit)`) in case of error? This would preserve more of the syntax on the caller side.

src/colvar_UIestimator.h
giacomofiorin: Sure, why not?

src/colvar_neuralnetworkcompute.cpp
giacomofiorin: Ok, but that means that the entire framework always needs to rely on `ifstream` after all... You can't explicitly allocate an `istream` object.

src/colvarproxy_io.h
```diff
@@ -0,0 +1,157 @@
...
+  /// How often the simulation engine will write its own restart
+  int restart_frequency_engine;
+
+  /// \brief Currently opened output files: by default, these are ofstream objects.
+  /// Allows redefinition to implement different output mechanisms
+  std::list<std::ostream *> output_files;
+  /// \brief Identifiers for output_stream objects: by default, these are the names of the files
+  std::list<std::string>    output_stream_names;
+
+  std::map<std::string, std::istream *> input_streams_;
```
HanatoK, line 150: Why are only input streams stored in `std::map`, while output streams are stored separately in two `std::list`s?

src/colvarproxy_io.cpp
```diff
@@ -0,0 +1,298 @@
...
+    return input_streams_[input_name];
+  }
+  return NULL;
+}
+
+
+int colvarproxy_io::close_input_stream(std::string const &input_name)
+{
+  if (input_streams_.count(input_name) > 0) {
+    input_streams_.erase(input_name);
```
HanatoK, line 188: The pointer is only erased from the container, but should it be deleted before `erase` for avoiding memory leak?

src/colvarproxy_io.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+  if (!isf->is_open()) {
+    if (error_on_fail) {
+      cvm::error("Error: cannot open "+description+" \""+input_name+"\".\n",
+                 COLVARS_FILE_ERROR);
+    }
+    return NULL;
+  }
+
+  input_streams_[input_name] = isf;
```
HanatoK, line 170: The input stream is created here, but not deleted in the destructor. Is this a memory leak?

HanatoK (changes requested):
How can I test the code path in `colvarproxy_io.cpp`?

src/colvarproxy_io.h
giacomofiorin: The latter is much older, when we couldn't be sure that `std::map` would be available on so many exotic platforms. I plan on converting the latter to `std::map` as well.

src/colvarproxy_io.cpp
giacomofiorin: Good point!

src/colvar_neuralnetworkcompute.cpp
HanatoK: I think `istream` also supports being converted to `bool` for checking these bits (https://en.cppreference.com/w/cpp/io/basic_ios/operator_bool).

giacomofiorin: @jhenin You came back with this request three times, which means that you feel very strongly about this design choice. I took a stab at it for a bit, but stumbled with two additional complications.
1. Returning a reference to an invalid object means that I could not initialize it right in the return statement, and I had to allocate a specific object just to return an invalid one. This object is return to threads > 0 if there is a bug, which I hope means that the code crashes before multiple threads crash the memory. In the back of my mind, I'm still concerned about deadlocks.
2. References cannot be reassigned, so those cases where you are reusing the same stream object become more complex, not simpler.

I will be busy for several days, so I put the work in progress in this branch and putting it back on draft status. At least the memory leaks should be addressed.

giacomofiorin:
> @giacomofiorin By the way, I see that VMD latest source code enables C++11 in the Makefile. Is it safe to use C++11 now? If so, we can use smart pointers to avoid the possible memory leaks.

That would be _really_ nice! Probably also possible in some future. But **not yet**!!

The Lord of the VMD did add a `CXX11` flag to appease us, but that flag is not on by default and he has not finished implementing it... You need something like these [patches](https://github.com/giacomofiorin/vmd-patches) for it to work.

Sadly, until the final "official" VMD 1.9.4 builds are finally published, you should not take anything for granted regarding what VMD will support or not support.

giacomofiorin:
> How can I test the code path in `colvarproxy_io.cpp`?

@HanatoK As far as I understand your question, the code path is already exercised every time input files are being read, which indeed turned out a small inconsistency with how the output prefixed was extracted in GROMACS vs. other codes (fixed in a08ccda95c00fc83fb3f4817a6ceb457d7469259).

Regarding using `std::map` for output streams, a different PR will follow later (this one is about bringing the output streams up to speed).

Do you see anything else that looks worth changing/fixing?

HanatoK:
> > How can I test the code path in `colvarproxy_io.cpp`?
> 
> @HanatoK As far as I understand your question, the code path is already exercised every time input files are being read, which indeed turned out a small inconsistency with how the output prefix was extracted in GROMACS vs. other codes (fixed in [a08ccda](https://github.com/Colvars/colvars/commit/a08ccda95c00fc83fb3f4817a6ceb457d7469259)).
> 
> Regarding using `std::map` for output streams, a different PR will follow later (this one is about bringing the input streams up to speed).
> 
> Do you see anything else that looks worth changing/fixing?

Sorry for my late reply, but currently I really have no time to review the code before the end of June. I vaguely remember thant I cannot test the code in `colvarproxy_io.cpp` via NAMD, but I will confirm it if I have time.

src/colvarbias_abf.cpp
```diff
@@ -708,43 +712,30 @@ void colvarbias_abf::read_gradients_samples()
...
-      is.open(z_gradients_in_name.c_str());
-      if (!is.is_open())  cvm::error("Error opening eABF z-gradient file " + z_gradients_in_name + " for reading");
-      z_gradients->read_multicol(is, true);
-      is.close();
+      error_code |= z_samples->read_multicol(z_samples_in_name,
+                                             "eABF z-histogram file",
+                                             true);
+      error_code |= z_gradients->read_multicol(z_gradients_in_name,
+                                               "eABF z-gradient file",
+                                               true);
```
jhenin, line 734: Splendid.

jhenin (approved):
Excellent changes! This abstraction really brought benefits already.

src/colvarproxy_io.h
HanatoK: I am OK if you do not have time to unify the handling of `std::istream*` and `std::ostream*` at this time.

HanatoK (approved):
Looks good to me!

src/colvarproxy_io.cpp
```diff
@@ -0,0 +1,291 @@
...
+  input_buffer_ = NULL;
+  restart_frequency_engine = 0;
+  input_stream_error_ = new std::istringstream();
+  input_stream_error_->setstate(std::ios::badbit);
+}
+
+
+colvarproxy_io::~colvarproxy_io()
+{
+  delete input_stream_error_;
```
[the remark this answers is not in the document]
giacomofiorin: Good point, memory leaks for from this class would be rare, but still they should be deleted behind the scenes since they are traditional pointer without garbage collection (no point in commenting further about smart pointers).

PR #461, opened 2022-03-04
Support Lepton in GROMACS build

giacomofiorin:
This PR adds support for patching Lepton into the GROMACS build system and runtime.

A working code is at:
https://github.com/Colvars/gromacs/tree/release-2021-colvars-lepton
and will be converted into patches for this repository before merging.

PR #483, opened 2022-06-01
Update Lepton via patching procedure

giacomofiorin:
The Lepton copies in the LAMMPS and NAMD repositories have lagged behind a bit. This PR makes it easier to keep those copies in sync by reusing the automatic download of Lepton (introduced for GROMACS in #461).

I'm not sure what to do for VMD, so I left it untouched for the moment.

### DIFF
--- a/.github/workflows/backend-template.yml
+++ b/.github/workflows/backend-template.yml
@@ -86,6 +86,13 @@ jobs:
           path: ~/.singularity
           key: Linux-x86_64-containers-build-2022-03-03
 
+      - name: Checkout OpenMM (for Lepton library)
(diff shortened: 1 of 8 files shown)


giacomofiorin:
I just fixed the NAMD recipe, and it worked locally. If the CI tests pass, it's good to go for me.

jhenin:
Thanks, this looks very good to me. I have only one question, which is what to do in case we notice a bug or incompatibility that requires patching Lepton locally for Colvars purposes. However, I think that is an unlikely event, and if it were to happen, we could then build the patching process into this pipeline, waiting for the issue to be solved upstream.

PR #485, opened 2022-06-27
Add decoupling option for moving restraints

### LINKED ISSUE #353: More practical smoothing of changing force constant in restraint
Right now targetForceExponent in restraints smoothes the lambda dependence of the force constant according to: kλ=k0+λ^α(k1−k0) , where k0, kλ, and k1 are the initial, current, and ﬁnal values of the force constant.

This is ideal when the force constant starts from 0, that is for the coupling transformation. However when doing the uncoupling transformation it would be more convenient to start from λ=1 and decrease it. This can be done explicitly by providing lambdaSchedule, but could be done automatically with a "decreasingLambda" flag. And that would  allow for single-stage uncoupling, if anyone ever uses that.

Essentially, the current convention is very convenient in the situation that is not often used in practice.

jhenin:
useful for restraint free energy perturbation: only efficient use of targetForceExponent.

In the process, the keyword _targetForceExponent_, which was found by users to be extremely confusing, was renamed _lambdaExponent_. The old keyword is still parsed with the parse_deprecated mode. The doc is updated accordingly.

Closes #353


jhenin:
Now ready for prime time!

src/colvarbias_restraint.cpp
```diff
@@ -538,12 +550,13 @@ int colvarbias_restraint_k_moving::init(std::string const &conf)
     target_nstages = lambda_schedule.size() - 1;
   }
 
-  if (get_keyval(conf, "targetForceExponent", force_k_exp, force_k_exp)) {
-    if (! b_chg_force_k)
-      cvm::log("Warning: not changing force constant: targetForceExponent will be ignored\n");
+  if ((get_keyval(conf, "targetForceExponent", lambda_exp, lambda_exp, parse_deprecated)
+    || get_keyval(conf, "lambdaExponent", lambda_exp, lambda_exp))
+    && !b_chg_force_k) {
+    cvm::error("Error: cannot set lambdaExponent unless a changing force constant is active.\n");
```
giacomofiorin, line 556: Use `COLVARS_INPUT_ERROR` here.

src/colvarbias_restraint.cpp
```diff
@@ -513,7 +514,18 @@ int colvarbias_restraint_k_moving::init(std::string const &conf)
...
+  if (b_decoupling) {
+    starting_force_k = 0.0;
+    target_force_k = force_k;
+    b_chg_force_k = true;
+  }
+
   if (get_keyval(conf, "targetForceConstant", target_force_k, target_force_k)) {
+    if (b_decoupling) {
+      cvm::error("Error: targetForceConstant may not be specified together with decoupling.\n");
+      return COLVARS_ERROR;
```
giacomofiorin, line 527: Use `COLVARS_INPUT_ERROR` here.

giacomofiorin (approved):
Looks great, two really minor comments.

PR #486, opened 2022-07-08
Remember first step of moving restraint

jhenin:
This is needed for correct handling of moving restraints (continuous, using targetNumSteps) that don't begin at step 0.


jhenin:
A VMD test fails with a segfault. Locally I can't replicate the segfault but I do get an error:
> colvars: Error: trying to interpolate between two colvarvalues with a lamdba outside [0:1].

which is due to a negative lambda, because VMD does not guarantee that timestep numbers increase over time. Not sure how to handle this: generally speaking, time-dependent biases don't behave correctly in VMD in the overwhelming majority of cases.  In the case of moving restraints, we have no information on the mapping between VMD frames and simulation timesteps, so there is no way to recover the right time dependence. Should time-dependent biases just freeze when `simulation_running` is false?

jhenin:
Freshly rebased onto master. Provided that the tests pass, I'll request a final review @giacomofiorin

jhenin:
Well, after fixing my oversight, we have passing tests! Ready for merging IMO.

src/colvarbias_restraint.cpp
```diff
@@ -766,16 +766,20 @@ int colvarbias_restraint_harmonic::update()
...
   error_code |= colvarbias_ti::update();
 
-  // update parameters (centers or force constant)
-  error_code |= colvarbias_restraint_centers_moving::update();
-  error_code |= colvarbias_restraint_k_moving::update();
+  if (cvm::main()->proxy->simulation_running()) {
+    // update parameters (centers or force constant)
+    error_code |= colvarbias_restraint_centers_moving::update();
+    error_code |= colvarbias_restraint_k_moving::update();
+  }
```
giacomofiorin, line 775: @jhenin For consistency I think this check should also be done for the other derived restraint classes, i.e. `colvarbias_linear` and `colvarbias_harmonic_walls`.

If so, maybe a simpler approach would be to do the check inside the two `*_moving::update()` methods in the base classes, so that derived restraint classes automatically get it. But I'm not hung up on this latter suggestion if there's a strong argument against.

giacomofiorin (changes requested):
The overall change is justified and clearly implemented.

About the part that affects VMD specifically (disabling changes in restraint centers or force constant), I think it should be done consistently across classes and documented. See specific comment.

One "nice to have" feature may be optionally changing the step counter via scripting commands in VMD, i.e. to "simulate a simulation". But that's clearly outside the scope of this PR.

jhenin:
@giacomofiorin followed your advice and moved the test into the update_* functions

PR #487, opened 2022-07-11
Use map of output streams, and return references to its elements

giacomofiorin:
As mentioned in the conversation for #481, update the registry of output files to use a `std::map` rather than two paired lists, which are no longer worth using any more.

### DIFF
--- a/src/colvarbias_meta.cpp
+++ b/src/colvarbias_meta.cpp
@@ -36,7 +36,6 @@ colvarbias_meta::colvarbias_meta(char const *key)
   : colvarbias(key), colvarbias_ti(key)
 {
   new_hills_begin = hills.end();
-  hills_traj_os = NULL;
 
   hill_weight = 0.0;
   hill_width = 0.0;
@@ -335,13 +334,12 @@ colvarbias_meta::~colvarbias_meta()
   colvarbias_meta::clear_state_data();
   colvarproxy *proxy = cvm::proxy;
 
-  if (proxy->get_output_stream(replica_hills_file)) {
+  if (proxy->output_stream_exists(replica_hills_file)) {
     proxy->close_output_stream(replica_hills_file);
   }
 
-  if (hills_traj_os) {
+  if (proxy->output_stream_exists(hills_traj_file_name())) {
     proxy->close_output_stream(hills_traj_file_name());
-    hills_traj_os = NULL;
   }
 
   if (target_dist) {
@@ -397,9 +395,12 @@ colvarbias_meta::add_hill(colvarbias_meta::hill const &h)
   }
 
   // output to trajectory (if specified)
-  if (hills_traj_os) {
-    *hills_traj_os << (hills.back()).output_traj();
-    cvm::proxy->flush_output_stream(hills_traj_os);
+  if (b_hills_traj) {
+    // Open trajectory file or access the one already open
+    std::ostream &hills_traj_os =
+      cvm::proxy->output_stream(hills_traj_file_name());
+    hills_traj_os << (hills.back()).output_traj();
+    cvm::proxy->flush_output_stream(hills_traj_file_name());
   }
 
   has_data = true;
@@ -429,12 +430,14 @@ colvarbias_meta::delete_hill(hill_iter &h)
     }
   }
 
-  if (hills_traj_os) {
+  if (b_hills_traj) {
     // output to the trajectory
-    *hills_traj_os << "# DELETED this hill: "
-                   << (hills.back()).output_traj()
-                   << "\n";
-    cvm::proxy->flush_output_stream(hills_traj_os);
+    std::ostream &hills_traj_os =
+      cvm::proxy->output_stream(hills_traj_file_name());
+    hills_traj_os << "# DELETED this hill: "
+                  << (hills.back()).output_traj()
+                  << "\n";
+    cvm::proxy->flush_output_stream(hills_traj_file_name());
   }
 
   return hills.erase(h);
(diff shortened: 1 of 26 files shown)


giacomofiorin:
To summarize, I got confused a bit about which functions need to be overridden and which overloaded. The former gets a little tricky depending on the compiler. Eventually I fixed my mistakes and squashed them and the fixes back into one commit. 

Since we're dealing with a templated class (`colvar_grid<...>`) as a base, we're not really using base-class pointers and vtables, except in three cases (`value_input()`, `value_output()` and the destructor). But even those don't need to be `virtual`, either. I'd like to simplify this if possible, and instantiate them explicitly in the derived classes before moving on with this PR. Comments welcome, especially when concerning the ABF code.

src/colvarproxy_io.cpp
```diff
@@ -209,60 +213,43 @@ std::ostream * colvarproxy_io::output_stream(std::string const &output_name,
...
-    return NULL;
+    return *output_stream_error_;
   }
 
-  std::ostream *os = get_output_stream(output_name);
-  if (os != NULL) return os;
-
-  if (!(mode & (std::ios_base::app | std::ios_base::ate))) {
-    backup_file(output_name);
+  if (output_streams_.count(output_name) > 0) {
```
jhenin, line 219: Could use output_stream_exists() ? There are a few instances of this.

src/colvarbias_meta.cpp
```diff
@@ -331,13 +330,12 @@ colvarbias_meta::~colvarbias_meta()
   colvarbias_meta::clear_state_data();
   colvarproxy *proxy = cvm::proxy;
 
-  if (proxy->get_output_stream(replica_hills_file)) {
+  if (proxy->output_stream_exists(replica_hills_file)) {
     proxy->close_output_stream(replica_hills_file);
   }
```
jhenin, line 335: This could be done without testing for the existence first, since close_output_stream handles it gracefully.

src/colvarmodule.cpp
```diff
@@ -1092,17 +1084,18 @@ int colvarmodule::calc_scripted_forces()
...
     return cvm::error("Error: in writing restart file.\n", COLVARS_FILE_ERROR);
   }
   proxy->close_output_stream(out_name);
-  if (cv_traj_os != NULL) {
+
+  if (proxy->output_stream_exists(cv_traj_name)) {
     // Take the opportunity to flush colvars.traj
-    proxy->flush_output_stream(cv_traj_os);
+    proxy->flush_output_stream(cv_traj_name);
   }
```
jhenin, line 1097: This could be done without testing if the flush_output_stream function was made more permissive, like close_output_stream, and ignored streams that aren't open. I'm not sure the test will be very effective at catching bugs, so I'd rather simplify the calling code.

src/colvarproxy_io.cpp
```diff
@@ -288,17 +277,29 @@ int colvarproxy_io::close_output_stream(std::string const &output_name)
...
+int colvarproxy_io::close_output_streams()
+{
+  if (! io_available()) {
+    return COLVARS_OK;
+  }
+
+  for (std::map<std::string, std::ostream *>::iterator osi = output_streams_.begin();
+       osi != output_streams_.end();
+       osi++) {
+    (reinterpret_cast<std::ofstream *>(osi->second))->close();
```
jhenin, line 300: How does this reinterpret_cast behave if the underlying object is not of type ofstream? Could that call an invalid function pointer and cause a crash?

src/colvarproxy_io.cpp
giacomofiorin: Done.

I also did the same for input streams, where the usage is very similar, even though it didn't come up in #481

src/colvarproxy_io.cpp
giacomofiorin: Since the map is private, the only chance that the wrong type of object can be stored there is if a derived proxy class overrides functions inconsistently. It's always possible, but it'd be a one-time error that would be caught when implementing that derived proxy class.

src/colvarmodule.cpp
giacomofiorin: Done

src/colvarbias_meta.cpp
giacomofiorin: Done

giacomofiorin:
@jhenin I addressed your comment, feel free to look again but do keep in mind that this is still a draft.

Before finishing it, I would like to eliminate the last remaining virtual functions from the `colvar_grid` template class. Are you okay with that? Do the existing ABF tests cover a reasonable combination of features?

jhenin:
I think the tests should do the job, yes.

giacomofiorin:
@jhenin I have reconsidered switching from deriving the class template to instantiating an object of it. Replacing the many existing member functions with wrappers around the instantiated object would be tedious but conceptually simple.

But then there's the trickier issue that `read_raw` and `write_raw` need internally to call the _derived_ functions `value_input()` and `value_output()`. They need to do that because `colvar_grid_gradient` and `colvar_grid_count` are intertwined, and need to call each other at a relatively low level. Changing that design seems possible, but beyond scope for this PR.

Tests pass, I'll just update the Actions workflow because GitHub is nagging and wrap this up.

src/colvarproxy_io.cpp
jhenin: Wouldn't it be safer to use dynamic_cast for these?

giacomofiorin: Ok, fair point. I was initially avoiding it because it is not as flexible with multiple inheritance, and we can't control what other classes will be used from here. But in the base one we do know that the inheritance is simple. Changing it.

jhenin (approved):

PR #494, opened 2022-09-01
Don't track NAMD Makefile, patch it

### LINKED ISSUE #490: The latest (lepton) version causes compilation error of NAMD
Hi,

When the latest devel version of NAMD was patched with the master (with downloaded lepton) version of Colvars, the compilation yielded the following error:
`Make.depends:1: *** recipe commences before first target. Stop.`

However, everything works if an old version (2022.5) of Colvars (without lepton) was used.

Any suggestion about how to address this issue is appreciated.

Thanks,
Haohao

jhenin:
Should fix #490


giacomofiorin:
It looks like the tests that use volumetric map-based variables fail because the NAMD 3 branch is currently _behind_ the NAMD 2 branch as far as Colvars support is concerned.

We could skip those tests by detecting the NAMD specific version, but I don't think that's very reliable since NAMD does not provide it (yet). The proposed change on [Gerrit](https://charm.cs.illinois.edu/gerrit/c/namd/+/3926) was never picked up and now that they moved to GitLab it is probably even less of a priority.

Earlier I was thinking of using the CVC map to list programmatically the available functions: this would have a direct benefit when troubleshooting input errors (so being worth implementing) but also useful for testing which tests to run. I can't implement this for a few weeks at least, so up to you whether to keep this PR open as a draft or just disable the NAMD 3 interface tests for the time being.

PR #499, opened 2022-09-22
Cmake debug

jhenin:
Option to build library with debug flags.


jhenin:
Good test to add @giacomofiorin ! I rebased onto the right branch as requested :-)

PR #500, opened 2022-10-20
Fix cvm not properly destructed in VMD

jhenin:
This fixes an issue in VMD where the module cannot be re-allocated after deleting, because the proxy pointer remains non-NULL. This is only a tentative fix, as I'd much rather locate the cause of the issue, because this used to work not so long ago.

### DIFF
--- a/src/colvarmodule.cpp
+++ b/src/colvarmodule.cpp
@@ -1228,8 +1228,8 @@ int colvarmodule::setup()
 
 colvarmodule::~colvarmodule()
 {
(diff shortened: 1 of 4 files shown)


jhenin:
Git bisect told me that the faulty commit is: ad45675076ae0c58a0470efff68151db4baca2f2
Rewriting the fix accordingly.

giacomofiorin (approved):
Thanks for catching that oversight!

PR #502, opened 2022-10-25
Accept multiple config files from NAMD (preliminary work)

jhenin:
Together with an equally tiny patch in NAMD (below), this makes it possible to parse multiple config files in NAMD. With current NAMD, this doesn't change the behavior.

```diff
diff --git a/src/SimParameters.C b/src/SimParameters.C
index f9e89889..8df00654 100644
--- a/src/SimParameters.C
+++ b/src/SimParameters.C
@@ -1874,7 +1874,7 @@ void SimParameters::config_parser_constraints(ParseOptions &opts) {
     opts.optionalB("main", "colvars", "Is the colvars module enabled?",
       &colvarsOn, FALSE);
     opts.optional("colvars", "colvarsConfig",
-      "configuration for the collective variables", PARSE_STRING);
+      "configuration for the collective variables", PARSE_MULTIPLES);
     opts.optional("colvars", "colvarsInput",
       "input restart file for the collective variables", PARSE_STRING);
```

### DIFF
--- a/namd/src/colvarproxy_namd.C
+++ b/namd/src/colvarproxy_namd.C
@@ -113,9 +113,9 @@ colvarproxy_namd::colvarproxy_namd()
            cvm::to_str(COLVARPROXY_VERSION)+".\n");
   colvars->cite_feature("NAMD engine");
   colvars->cite_feature("Colvars-NAMD interface");
-    // Construct instance of colvars scripting interface
+
   errno = 0;
-  if (config) {
+  for ( ; config; config = config->next ) {
     colvars->read_config_file(config->data);
   }
 


giacomofiorin:
I want to understand more: is there a use case where the Tcl command `cv configfile` is unsuitable?

If there is, okay with me to make `colvarsConfig` essentially doing the same thing. But also keep in mind that making it repeatable means that people may use it after the first `run` command, and run into this:
```
FATAL ERROR: Setting parameter extraBonds from script failed!
```
Repeating keywords is one of the most confusing aspects of the NAMD input parsing. If you go in and touch that, I think it's better to be consistent and add all the related keywords to `ScriptTcl` as well.

Lastly, doc updates are missing.

jhenin:
Good point about repeating the keyword in script mode... although it says:
`FATAL ERROR: Setting parameter colvarsConfig from script failed!`
which is not quite as confusing - but still a little.

The reason for adding it is the following: when runnin colvars with multiple inputs and alchemy together, the current situation requires that the two are interleaved:
```
alch on
<some alchemy keywords>

colvars on
colvarsConfig first.colvars
cv configfile second.colvars

runFEP $a $b $c $d
```
because both alchemy and colvars require a mix of pre-script parameters and scripting commands.

giacomofiorin:
> Good point about repeating the keyword in script mode... although it says: `FATAL ERROR: Setting parameter colvarsConfig from script failed!` which is not quite as confusing - but still a little.

Fair enough, I had copy-pasted it from a Slack message that a colleague who got confused by this. I tried to explain the logic behind NAMD parsing of "traditional" keywords vs. Tcl commands, and another colleague used the "exploding head" emoji... I may be to blame for how I tried to "clarify" it, but you'll probably agree that it isn't easy.

> I'm aiming for the cleaner version:
> 
> ```
> colvars on
> colvarsConfig first.colvars
> colvarsConfig second.colvars
> 
> alch on
> <some alchemy keywords>
> runFEP $a $b $c $d
> ```

What about this version? 
```
colvars on
cv configfile first.colvars
cv configfile second.colvars

alch on
<some alchemy keywords>
runFEP $a $b $c $d
```

jhenin:
> What about this version?

That doesn't work, you can't enable alchemy after entering script mode. This is what a user tried first, and reported to me, which prompted this PR.

giacomofiorin:
Ok then, I have to agree that we need to bite the bullet.

I'd like then to enable `colvars`, `colvarsConfig` and `colvarsInput` in script mode, by adding them to `ScriptTcl.C`. Or at least `colvarsConfig` and `colvarsInput`. What are your thoughts?

jhenin:
Yes, I agree about enabling all 3 commands. The nice thing is, they can be implemented as wrappers around colvarscript commands so they are transparent to the user without any further code duplication.

giacomofiorin:
Yeah that's the idea. The only complication will be making sure people can tell easily which version of the code they have.

giacomofiorin:
Also we need to make sure we can create a new GlobalMaster after initialization

jhenin:
> Also we need to make sure we can create a new GlobalMaster after initialization

Yes, that's worth a try. If that isn't possible, we can enable everything but the `colvars` command, that will be a good step already.

jhenin:
Are there any proposed changes that affect files that are tracked in this repo? If not, I propose we merge this, which is neutral in the status quo, and work on a separate PR for NAMD.

giacomofiorin:
Sure, I edited the title slightly, though, so that we don't end up thinking later that the work is done when it isn't.

giacomofiorin:
I'm keeping the branch open, I suggest we push future work there unless you prefer a different name.

jhenin:
I expect the rest of the implementation to happen on a branch of the NAMD repo, but who knows, there might be extra work needed on this side.

PR #509, opened 2022-12-28
Safer detection of Windows with recent Microsoft Visual Studio versions

akohlmey:
Compiling colvars in LAMMPS fails to compile on Windows with Visual Studio 2022 Community Edition 17.4 / MSVC++ v19.34 because the WIN32 define is not present (anymore?).

Using _WIN32 instead of WIN32 as predefined preprocessor macro to detect Windows works and is what is documented.
https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

Update: I think I have figured out why `#if defined(WIN32)` worked before: CMake sets this. I had a corrupted CMakeCache.txt file and all the predefined compiler switches were lost. That cause the failure to compile, but with this change it doesn't matter whether CMake sets the define because _WIN32 is set by the compiler.

### DIFF
--- a/src/colvarbias_meta.cpp
+++ b/src/colvarbias_meta.cpp
@@ -14,7 +14,7 @@
 #include <algorithm>
 
 // used to set the absolute path of a replica file
-#if defined(WIN32) && !defined(__CYGWIN__)
+#if defined(_WIN32) && !defined(__CYGWIN__)
 #include <direct.h>
 #define CHDIR ::_chdir
 #define GETCWD ::_getcwd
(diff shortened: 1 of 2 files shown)


giacomofiorin:
Thanks for catching this and particularly for looking into why `WIN32` works. I had started trying to figure out how to link the test executables in Windows (which would be good to have anyway), but it seems that even after achieving that things would have succeeded anyway thanks to CMake.

MSVC seems to require a lot of babysitting, which is probably why the CMake folks decided to define the macro themselves...

I assume that `_WIN32` has been defined by all versions of MSVC (otherwise, someone from Illinois may have troubl building the new VMD release, whenever that happens).

giacomofiorin (approved):

PR #510, opened 2023-01-06
[Draft] add support for GROMACS 2022

HubLot:
Hi,

Here a PR to add support of GROMACS 2022 (better late than never).
I add to modify the patch script because the version number is now in the global CMakeLists.txt.

Issue: I'm not able to compile on the GROMACS 2022 tree. I have `include` errors from the proxy file:
```
gromacs-2022.4_test-colvars/src/external/colvars/colvarproxy_gromacs.cpp:8:10: fatal error: gromacs/mdtypes/inputrec.h: Aucun fichier ou dossier de ce type
 #include "gromacs/mdtypes/inputrec.h"
```

I suspect it's due to the move of Colvars in the `external` folder and  their new CMake modifications (related to this global issue (https://gitlab.com/gromacs/gromacs/-/issues/3288)). 
I tried to find a solution by adapting the CMake files but without success (but I'm really a newbie with Cmake).
Do you have a clue to solve this?

### DIFF
--- a/.github/workflows/test-backends.yml
+++ b/.github/workflows/test-backends.yml
@@ -118,3 +118,15 @@ jobs:
       path_compile_script: devel-tools/compile-gromacs.sh
       test_lib_directory: gromacs/tests/library
       rpath_exe: install/bin/gmx_d
+
+  gromacs-2022:
+    name: GROMACS 2022
+    if: github.event_name == 'pull_request' || contains(github.event.head_commit.message, 'test-gromacs-2022')
+    uses: ./.github/workflows/backend-template.yml
+    with:
+      backend_name: GROMACS-2022
+      backend_repo: gromacs/gromacs
+      backend_repo_ref: release-2022
+      path_compile_script: devel-tools/compile-gromacs.sh
+      test_lib_directory: gromacs/tests/library
+      rpath_exe: install/bin/gmx_d
--- a/gromacs/gromacs-2022.x.patch
+++ b/gromacs/gromacs-2022.x.patch
@@ -1,0 +1,506 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index cd748c9..31e3b62 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -588,6 +588,10 @@ include(gmxManageLmfit)
+
+ include(gmxManageMuparser)
+
++include(gmxManageColvars)
++
++include(gmxManageLepton)
++
+ ##################################################
+ # Process SIMD instruction settings
+ ##################################################
+diff --git a/src/gromacs/CMakeLists.txt b/src_colvars/gromacs/CMakeLists.txt
+index 99cc804..40fbe29 100644
+--- a/src/gromacs/CMakeLists.txt
++++ b/src_colvars/gromacs/CMakeLists.txt
+@@ -139,6 +139,11 @@ if (WIN32)
+ endif()
+ list(APPEND libgromacs_object_library_dependencies thread_mpi)
+
++gmx_manage_colvars()
++gmx_manage_lepton()
++list(APPEND libgromacs_object_library_dependencies colvars)
++list(APPEND libgromacs_object_library_dependencies lepton)
++
+ if(GMX_INSTALL_LEGACY_API)
+   install(FILES
+ 	  analysisdata.h
+@@ -184,6 +189,8 @@ else()
+     add_library(libgromacs ${LIBGROMACS_SOURCES})
+ endif()
+
++gmx_include_colvars_headers()
++
+ if (TARGET Heffte::Heffte)
+     target_link_libraries(libgromacs PRIVATE Heffte::Heffte)
+ endif()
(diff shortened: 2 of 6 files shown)


HubLot:
A proper solution to the issue is to separate the Colvars library files and the proxy ones.

Colvars library files remains in `src/external` and those files don't need GROMACS headers to compile (like other libraries in this library).
Colvars-GROMACS proxy files are moved back to `src/gromacs` inside `src/gromacs/applied_forces/colvarproxy` since only those require GROMACS headers.
Furthermore, it helps separate what constitute the core of Colvars and what consitute the proxy between the library and the engine.

This what is done in the last commit (3392926)

HubLot:
Summary of the tests launched:
  - library tests: all passed
  - interface tests:
       - serial: segfault due to LINCS (need to investigate)
       - MPI: passed with 1e-5 precision
       - MPI_multi-sim: passed from 1e-3 to 1e-5 precision

giacomofiorin:
@HubLot Separating out the library files from the GROMACS interface code makes sense in general. I only put them in the same place to be easier to maintain patched worktrees, but clearly that's less of a concern than being able to build! :-) 

For the tests, having them pass at 1e-5 accuracy or worse does not seem very useful. Are you open to adding instead unit tests of the Colvars interface to the CMake targets of GROMACS? It would seem to me more consistent with the expectations of most users.

HubLot:
> @HubLot Separating out the library files from the GROMACS interface code makes sense in general. I only put them in the same place to be easier to maintain patched worktrees, but clearly that's less of a concern than being able to build! :-)

Indeed !

> For the tests, having them pass at 1e-5 accuracy or worse does not seem very useful. Are you open to adding instead unit tests of the Colvars interface to the CMake targets of GROMACS? It would seem to me more consistent with the expectations of most users.

Can you elaborate? You mean, having a folder `tests` inside `src/gromacs/applied_forces/colvarproxy` and run by GROMACS test suite?

giacomofiorin:
> Can you elaborate? You mean, having a folder `tests` inside `src/gromacs/applied_forces/colvarproxy` and run by GROMACS test suite?

Yes, exactly. All the "interface" tests would be in the right place there, because you can follow a bit more closely any changes in the `MDModules` infrastructure, which are likely to happen both within and outside your control.

I don't know if it would be also the right place for the library tests, but we can think about it later.

Also, looking at the rest of the `applied_forces` folder, why not rename the `colvarproxy` folder as simply `colvars`? This would make it clearer what the folder contains without needing to know the name of a class internal to Colvars. A simple README could point to the location of the library in `src/external` to clarify that not everything is in there. Although many of the modules in there (`awk`, `densityfitting`, etc) are fully contained within GROMACS, `qmmm` certainly isn't.

HubLot:
> Yes, exactly. All the "interface" tests would be in the right place there, because you can follow a bit more closely any changes in the `MDModules` infrastructure, which are likely to happen both within and outside your control.

Yes, it makes sense indeed but we won't be able to use colvars suite "as is" for running the tests, e.g spiff, Autodiff folder, reference files, etc, no?

> Also, looking at the rest of the `applied_forces` folder, why not rename the `colvarproxy` folder as simply `colvars`? This would make it clearer what the folder contains without needing to know the name of a class internal to Colvars. A simple README could point to the location of the library in `src/external` to clarify that not everything is in there. Although many of the modules in there (`awk`, `densityfitting`, etc) are fully contained within GROMACS, `qmmm` certainly isn't.

Agreed, I'll amend the PR.

giacomofiorin:
> Yes, it makes sense indeed but we won't be able to use colvars suite "as is" for running the tests, e.g spiff, Autodiff folder, reference files, etc, no?
> 

That's right, but once you know that you have a working CMake and CTest, why would you use these other tools? 😉  They make sense to support NAMD and VMD but that's about it.

HubLot:
Yes, I agree. 
I think it may be out of the scope of this PR though.

giacomofiorin:
> Yes, I agree. I think it may be out of the scope of this PR though.

Yes, that should be a separate PR covering also other GROMACS releases that are currently supported. If you have a good idea where to start from, feel free to go ahead. Otherwise, I'm happy to take a look as well: it is worth at least considering running all the tests through CTest.

This PR looks okay right now. Do you have anything else that you plan to add?

HubLot:
> > Yes, I agree. I think it may be out of the scope of this PR though.
> 
> Yes, that should be a separate PR covering also other GROMACS releases that are currently supported. If you have a good idea where to start from, feel free to go ahead. Otherwise, I'm happy to take a look as well: it is worth at least considering running all the tests through CTest.
> 
Okay I will look into it and let you know.

> This PR looks okay right now. Do you have anything else that you plan to add?

No, it's good. It may be worth squashing the PR though due to the smalls and not really informative commits.

giacomofiorin:
> No, it's good. It may be worth squashing the PR though due to the smalls and not really informative commits.

Nah, it's only 5 commits, and some of them are actually large.

giacomofiorin:
@HubLot Can you check the status of the CI jobs in https://github.com/Colvars/gromacs? I had a few hiccups in updating the patched branches, but after solving those I noticed some segfaults on macOS.

HubLot:
it crash also on my Linux workstation, i'll investigate

PR #512, opened 2023-01-17
fix Uninitialized variables in GROMACS source tree.

HubLot:
Variables `bColvars` and *colvars_proxy` were not initialized in the `inputrec` struct in the GROMACS source tree.

It may cause a crash when the `bColvars` undefined behavior is equal `True` and the program try to delete the `colvars_proxy` pointer (which was not allocated).
It happens in the GROMACS 2022 integration tests (see #510)

the PR initialized the variables and make also sure we don't delete a null pointer `colvars_proxy`. it adds also an error message when the `bColvars` is set to true and the `colvars_proxy` is null.

I don't think this behavior can happen in a regular GROMACS usage because I set the `bColvars` variable to false manually when no colvars input file were found in the mdrun commandline (https://github.com/Colvars/colvars/blob/master/gromacs/gromacs-2022.x.patch#L350)

### DIFF
--- a/gromacs/gromacs-2021.x.patch
+++ b/gromacs/gromacs-2021.x.patch
@@ -3,9 +3,9 @@ index 3715ce1656..65134d0568 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
 @@ -572,6 +572,9 @@ include(gmxManageTNG)
- 
+
  include(gmxManageLmfit)
- 
+
 +include(gmxManageColvars)
 +include(gmxManageLepton)
 +
@@ -19,7 +19,7 @@ index a4430e9dd6..385b688501 100644
 @@ -142,6 +142,12 @@ if (WIN32)
  endif()
  list(APPEND libgromacs_object_library_dependencies thread_mpi)
- 
+
 +# Add Colvars and Lepton targets, embed their object code in libgromacs
 +gmx_manage_colvars()
 +gmx_manage_lepton()
@@ -32,7 +32,7 @@ index a4430e9dd6..385b688501 100644
 @@ -189,6 +195,8 @@ else()
      add_library(libgromacs ${LIBGROMACS_SOURCES})
  endif()
- 
+
 +gmx_include_colvars_headers()
 +
  # Add these contents first because linking their tests can take a lot
@@ -58,8 +58,8 @@ index fadc3d283e..8a6ce56a6f 100644
 +/* COLVARS : add a value of 1000 for Colvars support and
 + * prevent regular GROMACS to read colvars .cpt files */
 +static const int cpt_version = cptv_Count - 1 + 1000;
- 
- 
+
+
  const char* est_names[estNR] = { "FE-lambda",
 @@ -1238,6 +1241,15 @@ static void do_cpt_header(XDR* xd, gmx_bool bRead, FILE* list, CheckpointHeaderC
      {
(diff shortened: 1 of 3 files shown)


gromacs/gromacs-2021.x.patch
```diff
@@ -502,30 +505,5 @@ index c2973bb1af..cb4d1da254 100644
 +           Generic data file
 + -colvars_restart [&lt;.dat&gt;]  (colvars.dat)    (Opt.)
 +           Generic data file
- 
+
  Options to specify output files:
- 
-diff --git a/.github/workflows/build_cmake.yml b/.github/workflows/build_cmake.yml
```
jhenin, line 508: So this part of the patch isn't needed anymore? It's better this way, we don't want to be altering these build recipes.

jhenin (commented):
Good change! Just checking where the change in CMake recipe for 2021.x comes from.

gromacs/gromacs-2021.x.patch
HubLot: Well this part should not be in the patch file since it concerns only the Colvars repo and not the GROMACS source tree

giacomofiorin: @HubLot This patch that you are removing was introduced in bbe94155 and meant to a CI error in standard GROMACS 2021: 
https://github.com/gromacs/gromacs/actions/runs/3629868210/jobs/6122561771
This error was due to a change in the macOS Azure instances available, and the GROMACS people fixed it in 2022 and later. But 2021 seems to be already lagging behind even though 2023 is still in preview...

Ideally, they should backport the CI fix in 2021.  But if we didn't do it in the patch, the CI jobs on the 2021 branch of our own fork  for reasons that are unrelated to Colvars. Whether or not you are in the mood to check the details (I probably wouldn't) it's impossible to see them without being logged into GitHub.

HubLot: Thanks for the explanation.
I understand the issue but this patch should not be in this file. This gromacs-2021.x.patch is meant to be used with a standard GROMACS source tree on a standard user computer. 
Right now, the `update-colvars-code.sh` is "failing" because `patch` binary doesn't know what do to with those lines (path *.github/workflow/...* is unknown).

So, it should remain in CI somehow

giacomofiorin: Ah, ok. In my view, the best-supported would be to download one of the pre-patched releases, rather than downloading a standard release and patching it ourselves.

That said, it's safer to add a check for this, which I did in `master`, more precisely f3808039. Can you rebase this branch over that?

jhenin: Thanks Giacomo, this works, although it adds a few very specialized lines to user-facing `update-colvars-code.sh` which is getting a little heavy. I wonder if that would be handled directly within the CI scripts ?

giacomofiorin: Feel free to move them, but only in a way such that those lines are run regardless of whether you are on a user desktop or an Actions instance.

jhenin: Fair enough. Since I don't see an simple way to do this, I'm ok with keeping it the way it is in current master.

HubLot: okay for the fix in the update-colvars.code.sh 
I'll rebase the branch

PR #513, opened 2023-01-25
Set target temperature through a common code path

giacomofiorin:
Minimize divergences between MD engines by storing the target temperature in a proxy member and provide a shared function `set_target_temperature()` to set it. The previous virtual function `temperature()` is now a `const` accessor `target_temperature()`. 

Additionally, there is now a script method ~~`settemperature`~~ `targettemperature` to set it in scripts, i.e. in post-processing or when we are temporarily unable to get it from the MD engine (see e.g. #218).

~~Questions:~~
- ~~Should `set_target_temperature()` be `virtual` or just remain a plain setter?~~
- ~~Name of the script function (`settemperature -> `settargettemperature`)?~~


jhenin:
> Should set_target_temperature() be virtual or just remain a plain setter?

Trying to think of a possible use for an override... The only one that comes to mind is an annealing method that would be driven from Colvars. That seems distant and unlikely, so I'd keep a plain setter for now, for simplicity. If the need arises to override it, making it virtual will be an easy change.

> Name of the script function (settemperature -> settargettemperature)?

I think a getter would be useful too. The function could be called `targettemperature` and be a getter/setter like `cv frame`.

gromacs/gromacs-2022.x/colvarproxy_gromacs.cpp
```diff
@@ -47,9 +47,14 @@ void colvarproxy_gromacs::init(t_inputrec *ir, int64_t step, const gmx_mtop_t &m
...
 
+  // From Gnu units
+  // $ units -ts 'k' 'kJ/mol/K/avogadro'
+  // 0.0083144621
+  boltzmann_ = 0.0083144621;
+
   // Get the thermostat temperature.
   // NOTE: Considers only the first temperature coupling group!
-  thermostat_temperature = ir->opts.ref_t[0];
+  set_target_temperature(ir->opts.ref_t[0]);
```
HubLot, line 57: Could we use  directly the attribute `target_temperature_`? To be consistent with botlzmann above?

gromacs/gromacs-2022.x/colvarproxy_gromacs.cpp
```diff
@@ -47,9 +47,14 @@ void colvarproxy_gromacs::init(t_inputrec *ir, int64_t step, const gmx_mtop_t &m
 
   angstrom_value = 0.1;
 
+  // From Gnu units
+  // $ units -ts 'k' 'kJ/mol/K/avogadro'
+  // 0.0083144621
+  boltzmann_ = 0.0083144621;
```
HubLot, line 53: It could be worth using the GROMACS constant here:  see `gmx::c_boltz`

HubLot (commented):
I just have 2 comments but otherwise it's looks good on the gromacs side

gromacs/gromacs-2022.x/colvarproxy_gromacs.cpp
HubLot: I know we always write the number before but we could benefit from the PR

gromacs/gromacs-2022.x/colvarproxy_gromacs.cpp
giacomofiorin: Yeah, initially I wanted to use a function also for the Boltzmann constant, which could have been done via `set_unit_system()`, but it's better to leave it be for now.

Let me try adding `set_physical_constants()`, which could be a function that gets called within `set_unit_system()` when/if these are changed.

giacomofiorin: Actually, scratch that, sorry. I don't want to mess with the units just yet. I'm going to set the Boltzmann constant from GROMACS as you suggested in the other comment.

For this comment specifically, I would prefer to keep the temperature (which is a user choice) separate from the physical constants (which are fixed and under the control of the unit system but not the user's currently). Better if these two are not consistent just yet.

HubLot: okay, fine by me!

gromacs/gromacs-2022.x/colvarproxy_gromacs.cpp
giacomofiorin: @HubLot I couldn't the definition of `c_boltz` anywhere. Is it auto-generated? I tried guessing from what similar source files include.

giacomofiorin: I tried but...
image

HubLot: okay so:
  - for 2020 & 2021: `#include "gromacs/math/units.h"` with `BOLTZ` as a name (without `gmx::`)
  - for 2022: `#include "gromacs/math/units.h"` with `gmx::c_boltz`

HubLot: didn't catch they changed the constant name in 2022 and above :)

giacomofiorin: Okay this works. Thanks!

jhenin:
Trying VMD...
```
vmd > cv targettemperature
9.04115e+271

```
Hmmm... This will need a little fix.

jhenin:
Force-pushed an update because of malicious behavior of multi-file search in VS Code, intentionally missing Gromacs files.

jhenin (approved):

giacomofiorin:
> Hmmm... This will need a little fix.

...or not.  Remember that in VMD it is the caller's responsibility to initialize class members? :-)

src/colvarmodule.cpp
```diff
@@ -2107,18 +2107,6 @@ int cvm::load_coords_xyz(char const *filename,
 // Wrappers to proxy functions: these may go in the future
```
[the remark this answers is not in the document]
giacomofiorin: Absolutely, but in another PR. #487 is an example where tidying things up ended up in the way of finishing a specific task.

jhenin:
>  Remember that in VMD it is the caller's responsibility to initialize class members? :-)

PR #520, opened 2023-04-06
Fix ref count

giacomofiorin:
The number of Colvars objects that use each atom was not being updated correctly due to a missing increment in reference count in a copy constructor. This is harmless because the refcounts are not used anywhere at the moment, but may cause subtle bugs in the future.

Scripting methods to get the numbers of atoms and groups currently being used by Colvars are added; for atoms, there are two distinct commands, one that returns the number of all atoms ever requested, the other for the one currently in use.

### DIFF
--- a/doc/cvscript-tcl.tex
+++ b/doc/cvscript-tcl.tex
@@ -144,6 +144,33 @@
 \texttt{-------}
 \\
 \texttt{E : float - Amount of energy (internal units)}
+\item \texttt{cv getnumactiveatomgroups}
+\\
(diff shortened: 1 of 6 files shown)


jhenin:
To make the relationship clearer, can the seemingly unrelated names "refcount" and "ncopies" be made the same - one way or another?

giacomofiorin:
> To make the relationship clearer, can the seemingly unrelated names "refcount" and "ncopies" be made the same - one way or another?

Done

PR #522, opened 2023-04-07
Use iostream::fail() to check for I/O error

### LINKED ISSUE #521: cv load with non-existent file fails silently
This code in `colvarmodule::setup_input()`
```
    std::istream *input_is = &(proxy->input_stream(restart_in_name,
                                                   "restart file/channel",
                                                   false));
```
has a flag which results in file errors being ignored. @giacomofiorin is this by design, or leftover from testing?

jhenin:
Fixes #521

### DIFF
--- a/colvartools/abf_data.cpp
+++ b/colvartools/abf_data.cpp
@@ -39,7 +39,7 @@ ABFdata::ABFdata(const char *gradFileName)
 
     std::cout << "Opening file " << gradFileName << " for reading\n";
     gradFile.open(gradFileName);
-    if (!gradFile.good()) {
+    if (gradFile.fail()) {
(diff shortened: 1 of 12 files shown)


jhenin:
> it is really not self-apparent why one should use fail() instead of bad(). Our use case has always been "test if there are any errors (no matter of what kind) and then read the stream".

The intended use of `bad()` is not very clear to me, that is, based on the name I would expect `bad()` to be the broadest way of catching errors. But I've accepted that it isn't :-) 

I did consider using the bool cast, which is semantically equivalent to `fail()`, and thought that `fail()` was more explicit and slightly easier to read, but I'm not strongly opinionated on this. I'm ok with testing the stream objects themselves.

PR #525, opened 2023-04-17
Implement reflecting boundary conditions for metadynamics bias

fabsugar:
Dear All,
the branch meta_reflection includes important modifications of the metadynamics code (colvarbias_meta.*) which allow to correct known artifacts of metadynamics at the boundaries of the collective variables (either natural boundaries or added through flat bottom potentials).
Namely, I implemented the reflection of the metadynamics Gaussians through the boundaries and an additional feature that removes the biasing forces beyond the boundaries (so that those forces smoothly go to zero beyond the boundaries). The implementation is specifically designed to be general and applicable to any dimensionality. The output pmf and bias energy includes those corrections (e.g. it remains flat beyond the border of a given collective variable while keeping the same shape across the others).
This feature also solves the need to add explicit Gaussians beyond the boundaries, which is computationally inefficient (and currently might need some fix).
The new feature is simple to use, it only requires the keyword "hills_reflection on" and automatically works on the default boundaries. Upon activating it the users will be able to run long metadynamics simulations of any dimensionality without developing the large errors at the boundaries that are currently present and that become larger and larger as the simulation goes on.
The new feature is fully tested on different systems up to a dimensionality of four.
The meta_reflection branch was merging fine with the master branch up to mid January, right now there is some minor conflict.
All the best,
Fabrizio


giacomofiorin:
Looking at the state of the history a squash-merge (i.e. merging all commits into one commit) may be appropriate when this PR is ready.

The conflict is near the function where the hills are added in the multiple walkers case.  Have you been able to test that combination of features before the conflicts became worse?

giacomofiorin:
P.S. I tried resolving the conflicts in the `meta_reflection_squash` branch.

fabsugar:
As far as I remember I didn't test the combination of meta reflection and
multiple walkers (I did test it with other multiple replica approaches).
I'll send a test with the version before the conflict and let you know.
Thank you,
Fab

giacomofiorin:
Sounds good. You could also start from `meta_reflection_squash`

fabsugar:
Sounds good, I'll do that,
Fab

PR #529, opened 2023-04-28
Log newline

### LINKED ISSUE #503: Should cvm::log systematically add a newline?
I find myself hunting for instances of cvm::log without an explicit "\n", and wondering if there are any cases at all where we don't want it. If not, the function could just add it, and if there are just a handful of cases where we don't want it, that could be a separate option to not add it.

jhenin:
Fixes #503

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@
 *.o
 *.so
 *.a
-*.x
 *.exe
 *.pyc
 .vscode/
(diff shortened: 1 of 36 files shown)


giacomofiorin:
This is hard to read. Was it absolutely necessary to remove the explicit trailing newlines everywhere? The most immediate outcome of that are merge/rebase conflicts with existing work branches.

giacomofiorin:
Superseded by #530

PR #530, opened 2023-04-28
Append newline to log and error messages if not already present

giacomofiorin:
Also make the indentation of error messages consistent with log ones


src/colvarmodule.cpp
```diff
@@ -1735,13 +1735,15 @@ std::ostream & colvarmodule::write_traj(std::ostream &os)
 
 void colvarmodule::log(std::string const &message, int min_log_level)
 {
-  if (cvm::log_level() < min_log_level) return;
+  if (cvm::log_level() < min_log_level || message.size() == 0) return;
```
jhenin, line 1738: Do we want to forbid the use of `cvm::log("")` to log a blank line?

jhenin: I mean, I thought you modified this line specifically to remove that option. Otherwise, leaving the line as it was before seems fine to me.

giacomofiorin: In all honesty I just wanted to not have to check the length of the string, but since the error message *must* be sent (even if empty due to a bug) there is no way around that. I amended them to be consistent.

PR #532, opened 2023-05-02
Add XYZ trajectory reading feature

jhenin:
tested in new unit test read_xyz_traj


giacomofiorin:
(*Copying here from single-commit's comment*)

We should probably disable these particular tests if C++11 or later is not available. We can't keep writing _everything_ in old C++.

jhenin:
Agreed on the principle. Now this one is written in old C++, and the cost is mild.
You are thinking of disabling the tests at the CI level?

giacomofiorin:
> Agreed on the principle. Now this one is written in old C++, and the cost is mild. You are thinking of disabling the tests at the CI level?

Yes I pushed a commit to disable the test building in the CMake scripts used in CI. I was thinking about this already because GCC was raising warnings about the `override` qualifiers in the stubs code.

Feel free to tweak it if you want this particular test to run, but I feel that C++98 code should probably just be checked in VMD.

jhenin:
I chose to take advantage of the recent input file map, and allow for sequential reading through repeated calls to the XYZ reading function. I could add a function similar to the example in the unit test, that reads the whole file in one go and returns an array of frames. That said, the first expected usage is sequential reading to play back MD trajectories for library tests.

giacomofiorin:
> I chose to take advantage of the recent input file map, and allow for sequential reading through repeated calls to the XYZ reading function. I could add a function similar to the example in the unit test, that reads the whole file in one go and returns an array of frames. That said, the first expected usage is sequential reading to play back MD trajectories for library tests.

The example looks good, and loading all frames in memory optimal for tests. I'm not sure I would want to immediately add support for loading all frames in memory, without knowing how large would the files be. It looks like the addition would be very quick, and can be decided when the need comes up.

P.S. I just rebased using the button.

jhenin:
Cool, then this is good for merging as far as I'm concerned.

giacomofiorin (approved):
Looks great. I added a commit to ensure a compatible CMake when I get back to trying re-enabling the unittests on Windows.

Issue #501, opened 2022-10-25
NAMD interface test 003_reinitatoms fails


jhenin:
NAMD 2.13 passes. NAMD 2.14 gives a segfault, and the devel version : "corrupted double-linked list".

This is afaik a rarely-used feature.

jhenin:
Related: how do we make sure NAMD interface tests are run regularly? Maybe run them in CI only when `colvarproxy` or `colvarproxy_namd` changes?

giacomofiorin:
That test occasionally fails, and given how many dusty corners of NAMD are being exposed by reinitatoms, the chances that this is due to Colvars are not high.

I'd keep an eye out for people mentioning sporadic errors in constant pH, although whatever race condition is there may not even show up for larger systems.

jhenin:
I have checked that this only happens when parsing a Colvars config after reinitatoms.
If I rebuild in debug mode, the segfault disappears. Neat.

jhenin:
Fixed by #537

PR #537, opened 2023-06-06
Initialize NAMD ID map entry immediately when requesting a new atom

### LINKED ISSUE #536: Incorrect RMSD when using "run 0" then "cv configfile" and then "run N" in NAMD
Hi! I am using NAMD by script running with Colvars, but the RMSD seems wrong. My Colvars configuration is
```
colvarsTrajFrequency 10
colvarsRestartFrequency 5000
indexFile alad.ndx
smp off

colvar {
  name dihedral_A
  dihedral {
    group1 {atomNumbers {5}}
    group2 {atomNumbers {7}}
    group3 {atomNumbers {9}}
    group4 {atomNumbers {15}}
  }
}

colvar {
  name dihedral_B
  dihedral {
    group1 {atomNumbers {7}}
    group2 {atomNumbers {9}}
    group3 {atomNumbers {15}}
    group4 {atomNumbers {17}}
  }
}

colvar {
    name rmsd
    rmsd {
        atoms {
            indexGroup Atoms
        }
        refPositionsFile ref.xyz
    }
}
```

I use the following commands in the NAMD input file :
```
colvars                 on

firsttimestep		0
run			0

cv configfile rmsd.in
run 40
```

But then the Colvars traj file is incorrect:
```
# step       
           0 
# step         dihedral_A             dihedral_B             rmsd                 
           0   -8.25111514934035e+01   7.82878040124970e+01   1.52341358796618e+00
          10   -8.13773511440876e+01   7.37152755486682e+01   1.52341358796618e+00
          20   -7.98939764048137e+01   7.21838555446005e+01   1.52341358796618e+00
          30   -7.80847181113126e+01   

7.41707606870538e+01   1.52341358796618e+00
          40   -7.63721010622145e+01   7.74879983813807e+01   1.52341358796618e+00
```
The rmsd column is incorrect and never updated. Any ideas?

The test file is attached.
[test_colvars_rmsd.tar.gz](https://github.com/Colvars/colvars/files/11663139/test_colvars_rmsd.tar.gz)

giacomofiorin:
Works around several issues arising when using NAMD's `GlobalMaster` infrastructure in restarted jobs. Even considering just atoms (i.e. no groups CoMs or volumetric maps), `GlobalMaster` can independently:
1. Gather atomic positions of certain atoms
2. Broadcast forces applied to atoms, which may or may not be the same atoms as above
3. When requested, gather total forces for atoms, which *may or may not be the same atoms as above*

For Colvars the three sets of atoms are meant to be the same for simplicity. However, other `GlobalMaster` based objects (e.g. `tclForces`) may request atoms differently. Therefore, we had a provision in place in `colvarproxy_namd` to request all those atoms within Colvars as well, since a73f32f4cdd3316b1e5f38f75224f2efc5b2e036.

What that provision was not accounting for is that the array of total forces (3) _survives_ a configuration reset. For example, when switching system in the `003_reinitatoms` test, the are 15 alpha carbons requested, but since they are not exactly the same set we end up with 29 total forces retrieved, potentially writing to invalid memory (because those "old" forces do not correspond to atoms currently requested) or raising an error otherwise. There is no easy way to empty that array of total forces without significant changes to NAMD, so I opted for requesting those atoms as well.

This solution may increase the communication overhead when running e.g. constant-pH NAMD where one cycles between multiple topologies, but using CPH in combination with total forces seems unlikely (and probably not appropriate, given its discontinuities).

Fixes #536 

Probably also #501 @jhenin do you mind running again in that same environment as when you opened that issue?

### DIFF
--- a/namd/src/colvarproxy_namd.C
+++ b/namd/src/colvarproxy_namd.C
@@ -287,18 +287,30 @@ int colvarproxy_namd::setup()
 
 int colvarproxy_namd::reset()
 {
+  if (cvm::debug()) {
+    cvm::log("colvarproxy_namd::reset()\n");
+  }
+
   int error_code = COLVARS_OK;
 
-  // Unrequest all atoms and group from NAMD
+  // Unrequest all positions, total forces, etc from NAMD
   modifyRequestedAtoms().clear();
+  modifyForcedAtoms().clear();
+  modifyAppliedForces().clear();
+
   modifyRequestedGroups().clear();
+  modifyGroupForces().clear();
+
 #if NAMD_VERSION_NUMBER >= 34471681
   modifyRequestedGridObjects().clear();
+  modifyGridObjForces().clear();
 #endif
+
+  requestTotalForce(false);
 
   atoms_map.clear();
 
-  // Clear internal Proxy records
+  // Clear internal atomic data
   error_code |= colvarproxy::reset();
 
   return error_code;
--- a/namd/src/colvarproxy_namd.C
+++ b/namd/src/colvarproxy_namd.C
@@ -166,7 +166,6 @@ int colvarproxy_namd::update_target_temperature()
 void colvarproxy_namd::init_atoms_map()
 {
   size_t const n_all_atoms = Node::Object()->molecule->numAtoms;
-  atoms_map.resize(n_all_atoms);
   atoms_map.assign(n_all_atoms, -1);
 }
 
--- a/namd/src/colvarproxy_namd.C
+++ b/namd/src/colvarproxy_namd.C
@@ -90,6 +90,8 @@ colvarproxy_namd::colvarproxy_namd()
           "the output restart file could be defined, exiting.\n");
   }
 
+  init_atoms_map();
+
   // initialize module: this object will be the communication proxy
   colvars = new colvarmodule(this);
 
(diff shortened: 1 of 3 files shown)


jhenin (approved):
Looks sound.

giacomofiorin:
@jhenin @HanatoK Although the change in itself is correct, it also seems like the fix is breaking the initialization in some other cases and more fixes are needed.

I'll push those here and re-request review after all the tests pass.

namd/src/colvarproxy_namd.C
```diff
@@ -160,14 +162,23 @@ int colvarproxy_namd::update_target_temperature()
 }
 
 
+
+void colvarproxy_namd::init_atoms_map()
+{
+  size_t const n_all_atoms = Node::Object()->molecule->numAtoms;
+  atoms_map.resize(n_all_atoms);
```
HanatoK, line 169: The `resize` is immediately followed by `assign`, so I think the `resize` can be removed?

HanatoK (approved):
Looks good to me.

namd/src/colvarproxy_namd.C
giacomofiorin: Done

giacomofiorin:
@jhenin I merged this since this is a clear bugfix. Because it may also fix #501 can you check in your case? I have not yet had issues (although in my case they were sporadic anyway) .

jhenin:
Fixes #501 for me!

PR #538, opened 2023-06-07
Close, but do not delete input streams

giacomofiorin:
Keep the entries in the map of input streams (see #481) accessible to objects derived from `colvarproxy`, as well as returning their names as a list.

@jhenin The latter convenience scripting function might also be useful in the Dashboard, by obtaining all the input files in a single query.  In the case of reused or templated configurations, it may be useful to print out an informational message listing those files for the user's benefit.

### DIFF
--- a/src/colvarproxy_io.cpp
+++ b/src/colvarproxy_io.cpp
@@ -207,13 +207,17 @@ std::istream &colvarproxy_io::input_stream(std::string const &input_name,
   }
 
   if (colvarproxy_io::input_stream_exists(input_name)) {
-    return *(input_streams_[input_name]);
+    if (! dynamic_cast<std::ifstream *>(input_streams_[input_name])->is_open()) {
+      // This file was opened before, re-open it.  Using std::ios::binary to
+      // work around differences in line termination conventions
+      // See https://github.com/Colvars/colvars/commit/8236879f7de4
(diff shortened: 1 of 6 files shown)


src/colvarproxy_io.cpp
```diff
@@ -207,14 +207,18 @@ std::istream &colvarproxy_io::input_stream(std::string const &input_name,
   }
 
   if (colvarproxy_io::input_stream_exists(input_name)) {
-    return *(input_streams_[input_name]);
+    std::ifstream *ifs = dynamic_cast<std::ifstream *>(input_streams_[input_name]);
+    if ( ifs && !ifs->is_open()) {
```
jhenin, line 211: The case ifs == NULL could mean that the pointer is to a valid istringstream. This method will likely be overridden in Gromacs and whenever other derived istream classes are used, but I'd rather have safeguards against null dereference.

giacomofiorin: Ok with me (the extra check doesn't hurt). However, I would very much to any derived class initializing an entry with a null pointer.

jhenin: So would I, this is just meant to catch null pointers resulting from a failure of dynamic cast.

PR #539, opened 2023-06-07
Full functional test without back-end

jhenin:
Reading trajectory from XYZ file using internal reader.


jhenin:
`colvarproxy_stub` is now able to perform meaningful colvar calculations, but is still missing some important components:

- reading system topology (esp. atom masses)
- reading PBC information

This is sufficient for some testing. The absence of PBC or masses reduces the coverage of the tests, though. If those two conditions were fulfilled, `colvarproxy_stub` would deserve the name `colvarproxy_standalone`.

giacomofiorin:
@jhenin The added functionality looks good enough to merge as-is in my opinion.

Note that masses, charges and PBCs are not the only features that remain to be covered in the standalone tests. Gradients or applied forces aren't being tested for correctness either, for example. For a single scalar variable or a single bias, I think you could read those from an additional XYZ file, but to cover every case would take a bit more work to set up. Maybe a different PR?

Renaming the class would be fine; what would be more complex is decide whether to make it a CLI tool useful beyond tests. This may or may not be redundant with VMD depending on the features.

I just amended your commit to resolve a conflict caused by merging #538, but otherwise do not plan to touch this branch significantly.

jhenin:
I agree with your points. I'll merge this now, and use cases will develop later.

PR #542, opened 2023-06-14
Support for properly close istringstreams in input_streams_

HubLot:
Hi,

In the GROMACS proxy, streams can also be `istringstreams` (in the MDmodule version).
This MR add the support too properly close (& reset) this kind of streams in case of the same stream is read twice during parsing.

This modification was in first part in the Colvars GROMACS proxy (in an override) but it may be better to have it directly in the parent class.

I wonder also if it would be nice to have a method to add `istringstreams` in `input_streams_` map directly in the `colvarproxy_io` class, like this:
```c++
void colvarproxy_io::add_input_string(std::string const &input_name, std::string const content)
{
  if (!io_available()) {
    cvm::error("Error: trying to access an input file/channel "
               "from the wrong thread.\n", COLVARS_BUG_ERROR);
    return;
  }
 if (!colvarproxy_io::input_stream_exists(input_name)) {
   input_streams_[input_name] = new std::istringstream(content);
  }
}
```

### DIFF
--- a/src/colvarproxy_io.cpp
+++ b/src/colvarproxy_io.cpp
@@ -238,8 +238,18 @@ int colvarproxy_io::close_input_stream(std::string const &input_name)
 {
   if (colvarproxy_io::input_stream_exists(input_name)) {
     std::ifstream *ifs = dynamic_cast<std::ifstream *>(input_streams_[input_name]);
-    if (ifs && ifs->is_open()) {
-      ifs->close();
+    if (ifs) {
+      if (ifs->is_open()) {
+        ifs->close();
+      }
+    }
+    else {
+      // From a string, just rewind to the begining
+      std::istringstream * iss = dynamic_cast<std::istringstream *>(input_streams_[input_name]);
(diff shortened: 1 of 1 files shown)


HubLot:
I was thinking about the second option.
The idea was just to provide a standard method so the specializations of the proxy can add easily a string as an input stream.

I don't see how it would need refactorization because after the use of this function, the regular input_stream(name_of_input) will be called as for now. But may be i miss something.

(In any case, it is just a suggestion)

jhenin (approved):
Simple, conservative, does the job.

giacomofiorin:
(Sorry for the accidental closure).

I would tend to agree with @jhenin that the specialized code does not necessarily need to be in the base class. However, it's a pain in the neck to integrate changes in multiple versions of the GROMACS proxy. And the MDModules version is not even tracked in this repository, at the moment. The more functionality is in the library, the less work is needed to deal with this.

giacomofiorin (approved):
Same here.

PR #543, opened 2023-06-15
Initialize input string streams alongside input file streams

giacomofiorin:
Implement suggestion raised in #542.

The logic of using `dynamic_cast` to check whether a pointer is `std::ifstream *` and `std::istringstream *` can easily become fragile in derived classes, so I removed the `virtual` qualifier. GROMACS proxy functions should wrap around these functions without overriding them. 

This arrangement could be changed later as far as I'm concerned, but I feel safer to not override for now.

### DIFF
--- a/src/colvarproxy_io.cpp
+++ b/src/colvarproxy_io.cpp
@@ -273,16 +273,14 @@ bool colvarproxy_io::input_stream_exists(std::string const &input_name)
 int colvarproxy_io::close_input_stream(std::string const &input_name)
 {
   if (colvarproxy_io::input_stream_exists(input_name)) {
-    std::ifstream *ifs =
-      dynamic_cast<std::ifstream *>(input_streams_[input_name]);
+    std::ifstream *ifs = dynamic_cast<std::ifstream *>(input_streams_[input_name]);
     if (ifs) {
       if (ifs->is_open()) {
         ifs->close();
       }
     } else {
       // From a string, just rewind to the begining
-      std::istringstream * iss =
-        dynamic_cast<std::istringstream *>(input_streams_[input_name]);
+      std::istringstream * iss = dynamic_cast<std::istringstream *>(input_streams_[input_name]);
       if (iss) {
         iss->clear();
         iss->seekg(0);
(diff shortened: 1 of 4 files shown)


src/colvarproxy_io.cpp
```diff
@@ -237,15 +286,16 @@ bool colvarproxy_io::input_stream_exists(std::string const &input_name)
 int colvarproxy_io::close_input_stream(std::string const &input_name)
 {
   if (colvarproxy_io::input_stream_exists(input_name)) {
-    std::ifstream *ifs = dynamic_cast<std::ifstream *>(input_streams_[input_name]);
+    std::ifstream *ifs =
+      dynamic_cast<std::ifstream *>(input_streams_[input_name]);
```
jhenin, line 290: Are these lines so long that they need to be chopped? I find it harder to read... what's the criterion?

jhenin:
I agree! My first intention was to override, but @HubLot  came up with a very simple solution that does not require it.

src/colvarproxy_io.cpp
giacomofiorin: They would go beyond column 80. That continuation style is what `clang-format` would use if we were enforcing it.

HubLot (commented):
great! I just have one comment

src/colvarproxy_io.cpp
```diff
@@ -228,6 +229,54 @@ std::istream &colvarproxy_io::input_stream(std::string const &input_name,
...
+  }
+
+  input_streams_[input_name] = new std::istringstream(content);
+
+  return *(input_streams_[input_name]);
+}
+
+
+std::istream &
+colvarproxy_io::input_string_stream(std::string const &input_name)
```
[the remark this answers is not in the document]
giacomofiorin: That's true, but the initialization can be ambiguous because of the optional arguments.

If you don't see a problem, I'm happy to make it less pedantic.

src/colvarproxy_io.cpp
giacomofiorin: The problem with this type of conventions is that most people agree that the old one is bad, but disagree on a replacement.

That said, of the two MD engines that specify one (GROMACS and LAMMPS) they seem to agree on 100, so I'm happy to fall in line.

PR #546, opened 2023-06-21
Add minimal .clang-format for library source code

giacomofiorin:
Specific options are to be consistent with most the existing codebase, while everything else is at default.


jhenin (approved):
+1 It's time we made these these choices explicitly!

PR #547, opened 2023-06-22
fix clang12 warnings

HubLot:
For the `colvarparse.h`it fixes:
```
/colvars/colvarparse.h:265:21: warning: use of old-style cast [-Wold-style-cast] 
     out.append(1, (char) ::tolower(in[i]) );
```

For the `colvartypes.h`, I'm not sure what is the best solution. The warning is present because there is a explicit destructor for those 2 classes so we need an explicit copy constructor and an explicit copy assignment
```
warning: definition of implicit copy assignment operator for 'vector1d<colvarmodule::rvector>' is deprecated because it has a user-declared destructor [-Wdeprecated-copy-dtor]
  inline ~vector1d()
```
So, I just defined the default ones for `vector1d` but maybe it needs more?
For `rmatrix`, I just removed the destructor since it was empty.

### DIFF
--- a/src/colvaratoms.h
+++ b/src/colvaratoms.h
@@ -167,7 +167,7 @@ public:
   atom_group(std::vector<cvm::atom> const &atoms_in);
 
   /// \brief Destructor
-  ~atom_group();
+  ~atom_group() override;
 
   /// \brief Optional name to reuse properties of this in other groups
   std::string name;
@@ -180,7 +180,7 @@ public:
   int init();
 
   /// \brief Initialize dependency tree
-  virtual int init_dependencies();
+  virtual int init_dependencies() override;
 
   /// \brief Update data required to calculate cvc's
   int setup();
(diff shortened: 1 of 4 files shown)


HanatoK:
If I remember correctly, `= default` is C++11

giacomofiorin:
> Erf... I'm so not used to pre C++11... I will correct this. Le 22 juin 2023, 18:38, à 18:38, HanatoK ***@***.***> a écrit:
> […](#)
> If I remember correctly, `= default` is C++11 -- Reply to this email directly or view it on GitHub: [#547 (comment)](https://github.com/Colvars/colvars/pull/547#issuecomment-1602987978) You are receiving this because you authored the thread. Message ID: ***@***.***>

Don't. Let's finish #548, merge it in `master` and rebase this and @HanatoK's work in #541 .

HubLot:
okay, i just rebased on master

jhenin:
All clear.

PR #548, opened 2023-06-22
Update required C++ language version to C++11

giacomofiorin:
Recent versions of GROMACS are warning at build time when a derived-class function is lacking `override`, or when the `NULL` macro is used for pointers. Rather than making additional efforts to work around these warnings and similar ones, it makes much more sense to just begin adopting the standard everywhere.

VMD is the only holdout, and the problem is not even with the application itself (which builds just fine with C++11) but rather with the very old OS used at UIUC to build VMD for distribution. We are unaware of their current status (last non-C++11 VMD alpha build is October 2021). However, they will have to address this sooner or later, and probably not because of a small library like Colvars but a much bigger one (i.e. CUDA).

Checklist:
- [x] Update README
- [x] Update documentation
- [x] Remove `#ifdef` guards, which are unreliable for Visual Studio
- [x] Remove pre-C++11 CI jobs

### DIFF
--- a/README-c++11.md
+++ b/README-c++11.md
@@ -25,6 +25,6 @@ We no longer support using extremely old compilers and OSes that are well past t
 
 ### Building VMD is quite complicated: could you just provide a working VMD build?
 
-In theory we *could*, but because of UIUC's restrictive license we are unfortunately unable to redistribute any version of VMD (modified or not).  It is worth noting that the UIUC license violates several of the best practice set forth by the National Institutes of Health (NIH), which funds the development of VMD:
+In theory we *could*, but because of UIUC's restrictive license we are unfortunately unable to redistribute any version of VMD (modified or not).  It is worth noting that the UIUC license violates several of the best practices set forth by the National Institutes of Health (NIH), which funds the development of NAMD and VMD:
 [https://datascience.nih.gov/tools-and-analytics/best-practices-for-sharing-research-software-faq](https://datascience.nih.gov/tools-and-analytics/best-practices-for-sharing-research-software-faq)
 We hope that this arrangement will change in the future, and we encourage you to let your preference known to the VMD and NAMD development teams.
--- a/README-c++11.md
+++ b/README-c++11.md
@@ -20,4 +20,11 @@ We no longer support using extremely old compilers and OSes that are well past t
 
 - **NAMD** follows the C++11 standard.
 
-- **VMD** precompiled builds *do not* use C++11 yet, and the optional `CXX11` flag in VMD's `configure` script does not work yet.  Please use the `c++11` patch from [this repository](https://github.com/giacomofiorin/vmd-patches/) to enable compilation of VMD.  What this patch does is to instruct the compilers to use C++11, if it's not the default.
+- **VMD** precompiled builds *do not* use C++11 yet, and the optional `CXX11` flag in VMD's `configure` script does not work yet.  Please apply the `c++11` patch from [this repository](https://github.com/giacomofiorin/vmd-patches/) to allow building VMD with C++11 whenever supported.
+
+
+### Building VMD is quite complicated: could you just provide a working VMD build?
+
+In theory we *could*, but because of UIUC's restrictive license we are unfortunately unable to redistribute any version of VMD (modified or not).  It is worth noting that the UIUC license violates several of the best practice set forth by the National Institutes of Health (NIH), which funds the development of VMD:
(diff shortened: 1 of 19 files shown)


giacomofiorin:
This PR is ready as far as I can see. Obviously, we only just began modernizing the code base, i.e. more work is needed but at least let's not spend that energy trying to bend over backwards.

@HanatoK @HubLot Your PRs are based on branches in your own repos, up to you whether you want to rebase them on top of this branch or not. For review purposes, I suggest we ignore C++11-related CI errors anyway.

cmake/CMakeLists.txt
```diff
@@ -11,6 +11,15 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 
 #set(CMAKE_VERBOSE_MAKEFILE ON)
 
+if(NOT DEFINED CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
jhenin, line 16: If we wanted to use modern CMake:
> [This method isn't bad for a final package, but shouldn't be used by a library.](https://cliutils.gitlab.io/modern-cmake/chapters/features/cpp11.html)

the more modern alternative is available for CMake 3.8 and later, so after 2017, which counts as quite recent in my book. I'm usually not a fan of pushing recent CMake versions because it's been an issue for me in the past, so I don't recommend this change now, but I'll keep an eye on this question.

jhenin (commented):
I like the simplification that results from this.

HubLot (approved):
Great!
Is it worth specifing in the global README.md (and also in REAMDE-c++11.md) that C++ is now required since version XXX and bump the colvars version ? One could add also a link to a commit to this last "non required C++11" version.
It could help people trying to find it (if those people exists :D)

HanatoK (approved):
The changes look good.

README-c++11.md
```diff
@@ -1,27 +1,30 @@
...
+- **VMD** precompiled builds *do not* use C++11 yet, and the optional `CXX11` flag in VMD's `configure` script does not work yet.  Please apply the `c++11` patch from [this repository](https://github.com/giacomofiorin/vmd-patches/) to allow building VMD with C++11 whenever supported.
+
+
+### Building VMD is quite complicated: could you just provide a working VMD build?
+
+In theory we *could*, but because of UIUC's restrictive license we are unfortunately unable to redistribute any version of VMD (modified or not).  It is worth noting that the UIUC license violates several of the best practices set forth by the National Institutes of Health (NIH), which funds the development of NAMD and VMD:
+[https://datascience.nih.gov/tools-and-analytics/best-practices-for-sharing-research-software-faq](https://datascience.nih.gov/tools-and-analytics/best-practices-for-sharing-research-software-faq)
+We hope that this arrangement will change in the future, and we encourage you to let your preference known to the VMD and NAMD development teams.
```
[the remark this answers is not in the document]
giacomofiorin: You're right, the invitation to reach out was especially naive on my part. Even very well articulated and constructive comments by advanced users can easily be discounted on the grounds that the "majority of the users are not asking for this". I'll amend.

cmake/CMakeLists.txt
giacomofiorin: Well, we _are_ already requiring CMake 3.13 since 60f6bc42 (added because you started creating symlinks in #539).

giacomofiorin: Update: I tried the more modern suggestion from that link, and it didn't allow specifying a later C++ version. I think that recipe is good for choosing a specific version for your package. However, we're not the ones deciding it and we rather need an easy way to adopt other people's choices during testing.

Leaving the `CMAKE_CXX_STANDARD` in place for now until a better solution is available.

giacomofiorin:
> Great! Is it worth specifing in the global README.md (and also in REAMDE-c++11.md) that C++ is now required since version XXX and bump the colvars version ? One could add also a link to a commit to this last "non required C++11" version. It could help people trying to find it (if those people exists :D)

Done, I listed the Colvars version as of today (assuming we're merging). As for giving a quick link to the people who need a version that works on old platforms, I kind of feel that they should do whatever they like, but without any assistance from us.

README-c++11.md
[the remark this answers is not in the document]
giacomofiorin: @giacomofiorin I amended the message, see if it looks okay to you?

PR #552, opened 2023-07-13
Initialize the list of variables early to restore --list-variables option

giacomofiorin:
Fixes a regression due to #479


HanatoK (approved):
Looks good.

jhenin:
In the absence of regression tests for these programs, I assume this is tested? If yes I suggest merging.
We can talk about testing for various scripts, but outside the scope of this PR, especially if it is a bugfix.

PR #554, opened 2023-07-19
Update containers for Linux CI builds

giacomofiorin:
Adding a new container based on CentOS Stream 9, with a package selection equivalent to the pre-existing CentOS 7 container. Although CentOS 7 supports some recent GCC versions, the rest of the OS is quite ancient.

Note: there is a continuing discussion about Red Hat Enterprise Linux, its clones, and other distributions such as Debian or Ubuntu. The choice of CentOS 9 in this PR is only intended out of convenience. 

With the addition of GCC 12 and Clang 16 alongside the older 4.8 and 3.3, there is a better mix of old and new compilers to test on.  For now the backends are still built on CentOS 7, because some regression tests fail slightly (1.0e-5), but otherwise all backends will build and run on the newer container as well.

Notably, when VMD is built with GCC 11, it will assume the default standard, which is C++17, i.e. no patching is needed to build VMD in compliance with modern C++. Maybe we could add Lepton to the build test as well?

Input welcome from all. @HubLot specifically should comment on whether we should build GROMACS on CentOS 9 or wait to change that until the MDModules interface is merged here.

### DIFF
--- a/.github/workflows/backend-template.yml
+++ b/.github/workflows/backend-template.yml
@@ -25,6 +25,10 @@ on:
         description: "Name of the branch of the backend repository"
         required: true
         type: string
+      container_name:
+        description: "Name of the container used to build on"
+        required: true
+        type: string
       vmd_plugins_repo:
         description: "Name of the VMD plugins repository"
         required: false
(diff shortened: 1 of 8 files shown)


HubLot:
Hi,

I'm definitively in favor of compiling GROMACS on a more recent distro. CentOS 9 is great.

PR #558, opened 2023-07-21
Unformatted (binary) checkpointing

giacomofiorin:
This PR introduces a new class `cvm::memory_stream` that has a similar API to STL streams, but in reality is a glorified wrapper to `std::memcpy()`.

To minimize the likelihood of inheritance-related issues, the class does not derive the STL stream classes, and instead re-implements select functions only.  This means that specific unformatted I/O functions needed to be added to the existing classes, but in most cases the existing code based on stream operators could be converted into templates.

The new implementation is CI tested on all backends by setting the environment variable `COLVARS_BINARY_RESTART` (see the related doc changes). This variable only controls writing of the state file, because the reading function checks for the magic integer before choosing between unformatted and formatted input. 

LAMMPS restart files now embed the unformatted (binary) Colvars state, but using the formatted state file on the side for just Colvars remains as an option consistent with the existing behavior.

Future improvements could involve writing or reading to/from a file stream directly, instead of a memory buffer (see comments by @HanatoK), however this is currently not allowed by LAMMPS (at least for reading) and GROMACS for the foreseeable future.


HanatoK:
It seems the main use case of `cvm::memory_stream` is to serialize the various C++ objects in Colvars. If so, why not consider using a 3rd serialization library such as [cereal](https://github.com/USCiLab/cereal)?

I would also want to know if `memcpy` is really necessary as it may hurt the performance, because in my impression it should be OK to `reinterprete_cast` double/float to char when writing to files. Copying to `external_output_buffer_` seems only to aggregate data.

giacomofiorin:
The intention is to support both checkpointing using files and (de)serializing to memory. A copy would be needed for the latter use case, and it would be probably wasteful for the former.

Do you expect that the performance overhead would be high enough that it would be worth implementing two different mechanisms?

Cereal looks awesome! It would be probably the best choice if we actually had the ability to choose freely about whether to add a dependency. Sadly, I don't think we do. We've been on a similar route with Lepton and the experience is somewhat of a mixed bag.

HanatoK:
> The intention is to support both checkpointing using files and (de)serializing to memory. A copy would be needed for the latter use case, and it would be probably wasteful for the former.
> 
> Do you expect that the performance overhead would be high enough that it would be worth implementing two different mechanisms?

I am not quite sure. What I was thinking about is something like 3D metadynamics simulations, and the grid object could be large. But I maybe wrong and perhaps the performance overhead is not a serious problem.

I am considering something similar to `std::string_view`. When performing `cvm::memory_stream::write_vector` we do not actually write the vector, but just hold a reference (or a shared pointer) to the original object, the position to write and the length of the array. Then when `cvm::memory_stream` is actually written to a file, we can call `reinterpret_cast` to write the data. What are the typical use cases when serializing to memory? Does it intend to communicate the objects via networks or MPI?

HanatoK:
While we cannot use `std::string_view` and `ranges` from C++20 and C++23, is it possible to implement similar things as follows:

```C++
#include <ranges>
#include <vector>
#include <string_view>
#include <cstring>
#include <iostream>
#include <sstream>

int main() {
  std::vector<int> a{1, 2, 3};
  std::vector<double> b{4.0, 5.0, 6.0};
  std::string_view a_sv(reinterpret_cast<char*>(a.data()), a.size() * sizeof(decltype(a)::value_type));
  std::string_view b_sv(reinterpret_cast<char*>(b.data()), b.size() * sizeof(decltype(b)::value_type));
  auto jv = std::ranges::join_view(std::vector{a_sv, b_sv});
  // the following two lines can be lazy evaluated
  std::stringstream ss; // can be fstream or other stream
  std::copy(jv.begin(), jv.end(), std::ostream_iterator<char>(ss, ""));
  std::string out = ss.str();
  std::cout << "Copied size: " << out.size() << std::endl; 
  // compare to the reference
  char* ref = (char*)malloc(sizeof(int) * a.size() + sizeof(double) * b.size());
  std::memcpy(ref, a.data(), a.size() * sizeof(int));
  std::memcpy(ref + a.size() * sizeof(int), b.data(), b.size() * sizeof(double));
  std::cout << "Ref size: " << sizeof(int) * a.size() + sizeof(double) * b.size() << std::endl;
  const int cmp = std::memcmp(out.data(), ref, out.size());
  std::cout << "Compare result: " << cmp << std::endl;
  free(ref);
  return 0;
}
```
see https://godbolt.org/z/MnGj1Edsz

giacomofiorin:
The most immediate use case would be checkpointing in GROMACS. As @HubLot can describe better than me, the MDModules interface places certain restrictions at the moment, which make it really difficult to access the checkpoint file stream directly. Copying data is just safer for now (or at least easier to maintain), but in the future a smarter "zero-copy" mechanism as you suggest may be possible if not even recommended.

I don't understand the example though: the first part uses `std::string_view` (C++17), what does the second part do differently from the code in this PR?

HanatoK:
> I don't understand the example though: the first part uses `std::string_view` (C++17), what does the second part do differently from the code in this PR?

I think it is possible to use an object of `std::vector<std::string_view>` (or a similar way that could be available only in C++11) to hold the memory buffer instead of `unsigned char *external_output_buffer_`. The second part is just for comparing the result with the outcome of `memcpy`.

giacomofiorin:
Sure, if there is a smarter container that works for C++11 it would be great to try it, especially if we could also use `std::copy` on it if required. In that case, supporting a C-style external buffer (which is currently not used by default) may not even be necessary.

At the moment the more time consuming part of the work is making all classes run their I/O via objects with similar semantics as STL streams, but are not derived from them.

giacomofiorin:
@HanatoK Do you know of a container that can provide what you have in mind for C++11?

HanatoK:
> @HanatoK Do you know of a container that can provide what you have in mind for C++11?

If you refer to `std::string_view`, apparently no, unless you implement a "fat pointer" if you don't want to rely on external libraries.

giacomofiorin:
@HubLot The code below from the LAMMPS interface can serve as boilerplate for GROMACS checkpointing, too. The writer allocates a memory stream, has Colvars write to it, and then writes to file. The reader copies the memory from a buffer that LAMMPS has just read from a file into an internal buffer where Colvars will read from.

https://github.com/Colvars/colvars/blob/dc08c905369b95d7bcfd7aef8ffb294149477e61/lammps/src/COLVARS/fix_colvars.cpp#L890-L919

The additional copy in the reader is because we can't guarantee that LAMMPS will still have that buffer by the time Colvars is able to read its contents (due to deferred initialization).

giacomofiorin:
@jhenin The fault of the update branch button is a merge commit, I just rebased it.

src/colvarmodule.cpp
```diff
@@ -794,7 +808,9 @@ int colvarmodule::calc()
       // Write restart file, if different from main output
       error_code |= write_restart_file(restart_out_name);
     } else {
-      error_code |= write_restart_file(output_prefix()+".colvars.state");
+      if (output_prefix().size()) {
+        error_code |= write_restart_file(output_prefix()+".colvars.state");
```
jhenin, line 812: Suggestion: use a different extension for binary restart files (`colvars.binstate`?), to make it obvious to the user and avoid mistakes.

jhenin (commented):
Nearly ready, just a couple suggestions. I should have a pass at the doc.

src/colvarmodule.cpp
```diff
@@ -118,6 +124,14 @@ colvarmodule::colvarmodule(colvarproxy *proxy_in)
 
   // set initial default values
 
+  binary_restart = false;
+  char const *env_var = getenv("COLVARS_BINARY_RESTART");
```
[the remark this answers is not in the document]
jhenin: Done.

src/colvarmodule.cpp
giacomofiorin: I looked into this for a bit, and the `.colvars.state` extension is used pervasively throughout the code. Making another extension work consistently would require significant restructuring, which I would rather not add to this PR.

My thinking behind not distinguishing the two was the following:
- LAMMPS and GROMACS have their own checkpoint/restart files and do not need `.colvars.state` files in any format;
- NAMD does not have a restart file, however NAMD users are dealing with the `.coor` file being in either PDB or binary format, so being confused would actually be consistent behavior!

Would you be okay with not supporting silent auto-detection of the format but raise instead an error when the file being read is not in the expected format? Whether you accidentally fail, or consciously neglect to set the environment variable in consecutive jobs, a scream test should be the appropriate course of action.

jhenin: Okay, thinking about it some more, I can't think of a case where a mistake would have serious consequences. So I'm going to backtrack on my comment, I think it works as it is.

jhenin: New idea: split the magic number into two halves. One to specify the file type, and one for the version. That way we can print a useful error message when trying to read an incompatible binary file.

giacomofiorin: The main problem is that the code version may not be enough to ensure that the binary file is compatible, as it could have been made assuming a different memory layout for complex types.  Safer to just declare "compatibility not guaranteed" in my opinion.

As far as the version of Colvars that produced the file, that's encoded in the `configuration { ... }` string block, which is the very first thing that appears both in text files and in binary files (after the magic number).

jhenin:
Rebased on master

giacomofiorin:
@jhenin I made some more updates to the doc also considering your concerns about terminology.

If you are okay can you go ahead and merge?

doc/colvars-refman-main.tex
```diff
@@ -958,13 +958,22 @@
 
 \cvsubsec{Input state file}{sec:colvars_input}
 
-Several of the sampling methods implemented in Colvars are time- or history-dependent, i.e.\ they work by accumulating data as a simulation progresses and use this data to change their biasing forces.  When continuing a simulation over consecutive runs, a \emph{state file} needs to be loaded into Colvars.
+Several of the sampling methods implemented in Colvars are time- or history-dependent, i.e.\ they work by accumulating data as a simulation progresses and use these data to determine their biasing forces.
+
+This requires that the Colvars information 
+
+a standard method for checkpointing (currently, these are only GROMACS and LAMMPS)
```
jhenin, line 965: Part of the sentence is missing, and I'm unsure what the intention was exactly.

giacomofiorin: Oops, working on multiple windows. My employer may not be ready yet to shut down, but my brain may be :-)

PR #559, opened 2023-07-24
Option to run NAMD regtests in GPU-resident mode

jhenin:
using the option CUDASOAintegrate in NAMD3
This is controlled by the new script option -gpu, via the environment variable NAMD_CUDASOA.
This required disabling zeroMomentum and tclForces.

For now, only the tests using the common NAMD script are affected. Many of the other tests could use it as well.

### DIFF
--- a/namd/tests/library/Common/common.namd
+++ b/namd/tests/library/Common/common.namd
@@ -70,7 +70,8 @@ if { [info exists xsc_file] > 0 } {
 }
 
 COMmotion               no
-zeroMomentum            yes
+# incompatible with TclForces as of 07/2023
+#zeroMomentum            yes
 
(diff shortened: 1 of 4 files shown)


jhenin:
This is not so much "ready for review" as ready for suggestions and improvements to make it final and merge it.

giacomofiorin:
I'm good with the fix in the Tcl script, but would prefer if  `-gpu` was a bit more specific. How about `-cudasoa`?

giacomofiorin:
I implemented the change I mentioned. Approve or revert depending on your preference and then merge?

giacomofiorin:
Seems like my commit is not showing up, but yours has the same change.

PR #566, opened 2023-08-14
Better handling of faulty configs in Dashboard

EzryStIago:
Motivation: when loading a config file into a VMD session, if there is an error (e.g. a missing ref file), the file simply doesn't load and the user has to edit the config file in a third-party text editor.

Preferred behavior: alert the user the issue and offer to open the config file in the dashboard

New behavior: when run_cv encounters an error, it still produces an error message but also returns a flag. This gets passed up the call stack to load which then asks the user if they want to open the file in the dashboard config editor (yes or no).

Possible points of concern:
1. Two error messages come up: the first is the cv error message that happens every time there's an issue with run_cv, the second asks if the user wants to edit the config file. This is preferable to the previous behavior, but suggests that revisiting the error handling paradigm might be useful (i.e. run_cv would return the error code and information while higher-level functions decide what to do with that information depending on the context)
2. I modified the logic of the edit_cv so that it can accept a config string. This makes the cvs argument do double-duty. What do you think of this approach? Would it be appropriate to rename cvs to args since it's now a more generic string?


jhenin:
Thanks @EzryStIago !

To answer your questions:
1. I recommend adding an optional run_cv parameter "silent" which defaults to false, and disables the error messages.
2. I agree that renaming cvs to args would be better.

jhenin:
I rebased this to clean it up and then realized I could not push to your fork, @EzryStIago. 
Now in PR #708

Issue #563, opened 2023-08-01
Description of a cvc is not updated after set_function_type


HanatoK:
I am trying to implement `debugGradients` for nested CVs (such as `aspathCV` and `azpathCV`). However, when I try to access the nested CVC's description by `cv[i_cv]->description` (see the code in https://github.com/HanatoK/colvars/blob/4cb61111581a9b3dcd7dea2c7282fc70ca6544e6/src/colvarcomp_apath.cpp#L306-L310), it prints lines like 

`Debugging gradients for cvc aspathCV0001 with respect to sub-CV cvc "002" of type , `

Full log can be found in https://github.com/HanatoK/colvars/blob/4cb61111581a9b3dcd7dea2c7282fc70ca6544e6/namd/tests/library/033_apathCV_distanceVec_derivatives/AutoDiff/test.colvars.out#L6905-L6912.

I inspect the code of constructor of CVCs, and it calls `colvar::cvc::init(conf);` to setup the description, but `colvar::cvc::set_function_type` can be called after that, and `description` is not changed. Is it possible to fix this issue?

jhenin:
Hey @HanatoK , please have a look at the PR I made to address this.

PR #568, opened 2023-08-19
Set CVC description consistently

jhenin:
Goal: address #563 

Create a function to update the CVC description consistently based on name and type, and call it when the type may have changed.


HanatoK:
Looks good to me!

giacomofiorin (approved):

jhenin:
Lifted draft status.

Issue #87, opened 2016-10-17
Implement fit gradients for non-scalar components (see also #61)


giacomofiorin:
These would be easily calculated if the atomic gradients were stored in the atom group. Solving issue #61 would allow to solve this easily, too.

Meanwhile, a `f_cvc_fit_gradients` feature could be used to print a Warning message.

jhenin:
Another benefit of that would be debugGradients on non-scalar cvcs.

jhenin:
This is a likely cause behind a long-standing bug reported by Chris Chipot, FE calculations on a scripted coordinate based on a vector cvc.

jhenin:
It would be nice to define a shortest path towards implementing this, if possible without all the changes envisioned in #61. But if necessary, that can be a good reason to speed up implementation of #61.

giacomofiorin:
Before moving on with all other things, can you reproduce the bug conclusively?

jhenin:
I can try, but I only expect a larger fittingGroup to distribute the counter-force on more atoms. Wouldn't the net counter-force would stay the same, as long as the fittingGroup describes the same frame of reference?

What I did though, is disable the fit gradients for a similar test where they were available, and I got similarly erroneous results. What I get from that test is that in that kind of applications (essentially receptor-ligand coordinates), the fit gradients are not optional.

giacomofiorin:
I'm clarifying the scope of the issue, until it is resolved.  The error applies to PMF calculations on vector variables with an external fitting group, which do not have a published implementation yet.  All current PMF calculations (which are restricted to one or more scalar variables) are unaffected.

jhenin:
It also affects pmf calculations on scalar variables that are scripted functions of vector components. That is a combination of documented features.

giacomofiorin:
If the script uses vector variables it is, of course, affected. Making a vector variable into a scalar through a script doesn't change the flow of the underlying code.

jhenin:
To avoid undetected incorrect behavior, the fit gradients should be available only when atomic gradients are calculated in the first place.

I see two levels at which it can be done:
- move the enableFitGradients option/feature up to the CVC class, where it can be dependent on atomic gradients being calculated;
- somehow enable atom groups to know whether they are provided with atomic gradients, so the check can be performed there.

To me the first option seems simpler, because I don't see cases where you'd want to control the fit gradients separately for different groups in one CVC, but maybe I'm missing something.

giacomofiorin:
[...]

PR #570, opened 2023-08-28
enable use of CVs defined by PyTorch neural network models 

zwpku:
This branch implements a class called `torchANN`, which allows to define cv components by loading pretrained PyTorch neural network models. 

#### Installation Steps

1. Download [LibTorch](https://pytorch.org/get-started/locally/). This package is required in order to enable the `torchann` class. First, download the code and unzip it.
 
    ```
         wget https://download.pytorch.org/libtorch/nightly/cpu/libtorch-cxx11-abi-shared-with-deps-latest.zip
         unzip libtorch-cxx11-abi-shared-with-deps-latest.zip
     ```

    In this way, the library is uncompressed under the current directory. Let's say it is located at /path/to/libtorch.  
  
2. Patch MD engine.  This step is done as usual using the script *update-colvars-code.sh*.  Enter the source code of Colvars package, and run:

    ```  
         ./update-colvars-code.sh /path/to/md-engine        
    ```

4. Compilation. This step depends on the engine to be compiled. 

    -  NAMD:  add "--with-colvars-torch --torch-prefix path/to/libtorch" to the argument of ./config  
    
       Assume packages that are required to build NAMD, e.g. charm, tcl/tcl-threaded, are already prepared.
       Then, one can compile the NAMD package with the following commands: 

    ```
        ./config Linux-x86_64-g++ --charm-arch multicore-linux-x86_64 --with-colvars-torch    \
              --torch-prefix /path/to/libtorch  --with-fftw3 --fftw-prefix /path/to/fftw
        cd Linux-x86_64-g++
        make 
    ```
    - GROMACS:  add "-DTorch_DIR=/path/to/libtorch/share/cmake/Torch" when running cmake

    An example of the command is:
 
    ``` 
        cmake .. -DCMAKE_INSTALL_PREFIX=/home/username/local/gromacs  \
                        -DFFTWF_LIBRARY=/home/username/mambaforge/lib/libfftw3f.so  \
                        -DFFTWF_INCLUDE_DIR=/home/username/mambaforge/include \
                        -DTorch_DIR=/path/to/libtorch/share/cmake/Torch/  \
                        -DCMAKE_CXX_COMPILER=/usr/bin/mpicxx \
                        -DOpenMP_gomp_LIBRARY=/home/username/mambaforge/lib/libgomp.so
    ```

    -  LAMMPS:  only installation by cmake is supported. In the directory of LAMMPS source code, run

     ```
          mkdir build && cd build
          cmake ../cmake -D PKG_COLVARS=yes -D COLVARS_TORCH=yes 
     ```

    and set the variable Torch_DIR in the file *CMakeCache.txt*.  When a cpu version of libtorch library is used, it may 
    also be necessary to set MKL path to empty: 

     ``` 
          MKL_INCLUDE_DIR:PATH=
     ```

     Alternatively, one could combine these steps in one command:

     ```
          cmake ../cmake -D PKG_COLVARS=yes -D COLVARS_TORCH=yes      \ 
              -D  Torch_DIR=/path/to/libtorch/share/cmake/Torch -D MKL_INCLUDE_DIR=
     ```
   
    After that, run make and make install to compile and install the package.

The class has only been tested using simple neural network models (i.e. an autoencoder on alanine dipeptide), under NAMD and GROMACS engines.  Feedbacks are welcome!

### A (trivial) example 

1. Create a PyTorch model

```python
import torch

class MyModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
    def forward(self, x):
        return x

model = MyModel()
scripted_cv_filename = f'./identity.pt'
torch.jit.script(model).save(scripted_cv_filename)
```

This Python script simply creates a model which is an identity map and save it to a file named *identity.pt*.

2. Define the COLVARS config file

This file defines two CVs using torchann class taking other cv components (here dihedral angles) as inputs.

```
colvarsTrajFrequency    10000
colvarsRestartFrequency 10000

colvar {
  name nn_0
  lowerBoundary -180.0
  upperBoundary 180
  width 5.0
  extendedLagrangian on
  extendedFluctuation 5.0
  extendedTimeConstant 200

  torchann {
    modelFile identity.pt
    m_output_index 0
    period 360

    dihedral {
      group1 { 
	atomnumbers 5
      }
      group2 { 
	atomnumbers 7
      }
      group3 { 
	atomnumbers 9
      }
      group4 { 
	atomnumbers 15
      }
    }

    dihedral {
      group1 { 
	atomnumbers 7
      }
      group2 { 
	atomnumbers 9
      }
      group3 { 
	atomnumbers 15
      }
      group4 { 
	atomnumbers 17
      }
    }

  }
}

colvar {
  name nn_1
  lowerBoundary -180.0
  upperBoundary 180
  width 5.0
  extendedLagrangian on
  extendedFluctuation 5.0
  extendedTimeConstant 200

  torchann {
    modelFile identity.pt
    m_output_index 1
    period 360

    dihedral {
      group1 { 
	atomnumbers 5
      }
      group2 { 
	atomnumbers 7
      }
      group3 { 
	atomnumbers 9
      }
      group4 { 
	atomnumbers 15
      }
    }

    dihedral {
      group1 { 
	atomnumbers 7
      }
      group2 { 
	atomnumbers 9
      }
      group3 { 
	atomnumbers 15
      }
      group4 { 
	atomnumbers 17
      }
    }
  }
}

abf {
  colvars nn_0 nn_1
  fullSamples	200
}
```


giacomofiorin:
@zwpku We made a few improvements to the Github Actions workflow in the master branch. Would it be okay to rebase this branch on top of it?

zwpku:
@giacomofiorin  I have merged the new workflow into this branch. By the way, is it possible to have this cvc enabled and checked in the Github Actions workflow?  It requires the libtorch package and slightly different commands for backend installations.  Thanks!

giacomofiorin:
Hi @zwpku any patching of the backend should be done via the `update-colvars-code.sh` script, which is also used to produced a modified worktree for each MD engine before we submit pull requests (for NAMD and LAMMPS only: VMD has no such organization).

However, including libtorch as a dependency of those packages would be a decision done by their maintainers, and it's not unlikely that they may say no.  Can you share contact information (privately if you prefer) so that we can put you in touch with them to help make the case for it?

At the moment, it would be safer to make libTorch support optional in the updater script, and we can/should test the library both with and without it in Actions. Any method to install libTorch that is sufficiently stable (i.e. we don't have to worry about tests failing because of issues with its installation) would be okay.

zwpku:
> Hi @zwpku any patching of the backend should be done via the `update-colvars-code.sh` script, which is also used to produced a modified worktree for each MD engine before we submit pull requests (for NAMD and LAMMPS only: VMD has no such organization).
 
> However, including libtorch as a dependency of those packages would be a decision done by their maintainers, and it's not unlikely that they may say no. Can you share contact information (privately if you prefer) so that we can put you in touch with them to help make the case for it?

> 
> At the moment, it would be safer to make libTorch support optional in the updater script, and we can/should test the library both with and without it in Actions. Any method to install libTorch that is sufficiently stable (i.e. we don't have to worry about tests failing because of issues with its installation) would be okay.

Hi @giacomofiorin, thanks!

This branch also patches MD engines using `update-colvars-code.sh`, although some changes are made to this script so that after patching one can provide extra command line arguments related to libTorch when compiling the patched MD engine (for this one needs to download libTorch separately and provide its location).  

These arguments are indeed optional. Without them, the compilation goes through and one gets an MD engine without libTorch support (i.e. the torchann class is disabled).

jhenin:
@zwpku Any chance that libtorch could be linked dynamically at run time? That would let us support it with minimally invasive changes to the MD engine build process.

zwpku:
@jhenin  It's not very clear to me, but I can give it a try. To do this, it seems that there are two issues that need to be resolved.   

   - Current implementation relies on some declarations (types and classes) from LibTorch's header files. I guess I'll have to find a way to hide them. 
   - LibTorch provides multiple libarary files. I'll need to figure out which one to load at runtime.

giacomofiorin:
Hi, unrelated to this work, I just moved the containers used in the CI workflow to GitHub (#603).  I have opted not to rebase this branch since there are conflicts that precede it. @zwpku Please comment if there is something that looks unclear in resolving them. IMO we only need to regenerate the dependency files for NAMD and LAMMPS (GNU make version) and address the very simple conflict in `gromacs/cmake/gmxManageColvars.cmake`.

I wanted however to bring up the opportunity to use GitHub's "Packages" feature to distribute pre-built versions of a MD engine with an updated Colvars _and_ linked to an updated libTorch. For potential users, the download/installation step would be almost trivial:
https://github.com/Colvars/colvars/blob/7b12838a143171faea007d364cd01127c4650847/.github/workflows/test-library.yml#L242
and the OCI/ORAS registry format is supposedly compatible across containers backends (i.e. Docker, Podman, Singularity/Apptainer, ...).

This could possibly address the second of @zwpku's concerns about managing the dynamic linking to multiple library files?

zwpku:
@giacomofiorin thanks. I roughly understood what you mentioned. I will resolve the conflicts and see how to distribute pre-built MD engines.

giacomofiorin:
> @giacomofiorin thanks. I roughly understood what you mentioned. I will resolve the conflicts and see how to distribute pre-built MD engines.

Great! About the containers, don't worry about looking into that in detail, I'm happy to help. The main concern is having compilation without libTorch proceed normally as the default option. Since this is already the case according to your previous comment, it is up to you when to move forward with review.

zwpku:
> Great! About the containers, don't worry about looking into that in detail, I'm happy to help. The main concern is having compilation without libTorch proceed normally as the default option. Since this is already the case according to your previous comment, it is up to you when to move forward with review.

By the way, this PR contains history of my local commits, which are probably not interesting to others. Is it better that I create another PR with cleaner history?

giacomofiorin:
Hi @zwpku I looked yesterday and it wasn't possible to fast forward. This means that the branch can't be rebased on top of `master` easily. We try to rebase to the extent that it is feasible to make it easier to track down bugs through bisection later. I'll give it a second look after you consider the branch ready for review.

It would not be a problem to keep all your commits (aside from merge commits, which would be suppressed in a rebase if it work). You can probably see that there are many rather trivial commits already in the repository :-)

If all you have left to do is to synchronize with `master` comment here so we can look again into whether a rebase is still possible.

Thank you!! :-)

zwpku:
@giacomofiorin  `git merge` works for me, but `rebase` will lead to conflicts.

src/colvarcomp.h
```diff
@@ -28,6 +28,11 @@
 #include "colvar.h"
 #include "colvar_geometricpath.h"
 
+#ifdef TORCH
+#include <torch/torch.h>
+#include <torch/script.h>
```
HanatoK, line 33: Maybe @giacomofiorin would prefer PIMPL for not directly including pytorch headers. I am fine with it anyway.

src/colvarcomp.h
```diff
@@ -1799,6 +1804,50 @@ class colvar::neuralNetwork
...
+    virtual void calc_gradients();
+    virtual void apply_force(colvarvalue const &force);
+
+    /// Redefined to handle periodicity
+    virtual cvm::real dist2(colvarvalue const &x1,
+			    colvarvalue const &x2) const;
+    /// Redefined to handle periodicity
+    virtual colvarvalue dist2_lgrad(colvarvalue const &x1,
+				    colvarvalue const &x2) const;
+    /// Redefined to handle periodicity
```
HanatoK, line 1837: Is it necessary to overload these functions to handle periodicity?

src/colvarcomp_torchann.cpp
```diff
@@ -0,0 +1,237 @@
...
+  if ((wrap_center != 0.0) && !is_enabled(f_cvc_periodic)) {
+    cvm::error("Error: wrapAround was defined in a torchANN component,"
+                " but its period has not been set.\n");
+    return;
+  }
+
+  std::string model_file ;
+  get_keyval(conf, "modelFile", model_file, std::string(""));
+  try {
+    nn = torch::jit::load(model_file);
```
HanatoK, line 37: Sometimes people train the model on GPU and save it, and I guess if we load the model as is the code will fail. Is it better to add a line:
```
nn.to(at::Device("cpu"));
```
to make the code foolproof?

src/colvarcomp_torchann.cpp
```diff
@@ -0,0 +1,237 @@
...
+    cvm::log("torch model loaded.") ;
+  } catch (const std::exception & e) {
+    cvm::error("Error: couldn't load libtorch model (see below).\n" + cvm::to_str(e.what()));
```
HanatoK, line 40: Add a `return;` after `cvm::error` to make the code compatible with VMD.

src/colvarcomp_torchann.cpp
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  input_tensor = torch::zeros({1,(long int) num_inputs}, options);
+
+  try { // test the model 
+    std::vector<torch::jit::IValue> inputs={input_tensor};
+    nn_outputs = nn.forward(inputs).toTensor()[0][m_output_index];
+    cvm::log("Evaluating model with zero tensor succeeded.");
+  } catch (const std::exception & e) {
+    cvm::error("Error: evaluating model with zero tensor failed (see below).\n" + cvm::to_str(e.what()));
```
HanatoK, line 100: Add a `return;` after `cvm::error` to make the code compatible with VMD.

src/colvarcomp_torchann.cpp
```diff
@@ -0,0 +1,237 @@
...
+  nn_outputs = nn.forward(inputs).toTensor()[0][m_output_index];
+
+  input_grad = torch::autograd::grad({nn_outputs}, {input_tensor})[0][0];
+
+  /*
+  if (use_gpu)
+    input_grad = input_grad.to(torch::kCPU);
+  */
+
+  x = nn_outputs.item<double>() ;
```
HanatoK, line 149: It would be better if `cvm::real` is used instead of `double`.

src/colvarcomp_torchann.cpp
```diff
@@ -0,0 +1,237 @@
...
+void colvar::torchANN::calc_gradients() {
+  for (size_t i_cv = 0; i_cv < cv.size(); ++i_cv) {
+    cv[i_cv]->calc_gradients();
+    if (cv[i_cv]->is_enabled(f_cvc_explicit_gradient)) {
+      const cvm::real factor_polynomial = getPolynomialFactorOfCVGradient(i_cv);
+      // get the initial index of this cvc
+      size_t l = cvc_indices[i_cv];
+      for (size_t j_elem = 0; j_elem < cv[i_cv]->value().size(); ++j_elem) {
+	// get derivative of neural network wrt its input 
+	const cvm::real factor = input_grad[l+j_elem].item<double>();
```
HanatoK, line 164: It would be better if `cvm::real` is used instead of `double`.

src/colvarcomp_torchann.cpp
```diff
@@ -0,0 +1,237 @@
...
+      }
+    } else {
+      const colvarvalue& current_cv_value = cv[i_cv]->value();
+      colvarvalue cv_force(current_cv_value);
+      cv_force.reset();
+      const cvm::real factor_polynomial = getPolynomialFactorOfCVGradient(i_cv);
+      // get the initial index of this cvc
+      size_t l = cvc_indices[i_cv];
+      for (size_t j_elem = 0; j_elem < current_cv_value.size(); ++j_elem) {
+	cv_force[j_elem] = factor_polynomial * input_grad[l+j_elem].item<double>() * force.real_value;
```
HanatoK, line 192: It would be better if `cvm::real` is used instead of `double`.

HanatoK (commented):
Looks good overall. However, I have not run the code explicitly. It would be better if there is a test case (skipped if no pytorch or `torchann`).

zwpku:
> Looks good overall. However, I have not run the code explicitly. It would be better if there is a test case (skipped if no pytorch or `torchann`).

@HanatoK Thanks a lot for the suggestions! I included a test case but it couldn't pass the checks on GitHub. It's a good idea to run the test only when Libtorch is supported. I'll see whether this can work out.

src/colvarcomp.h
zwpku: It seems to me that it is necessary. I actually overloaded them some time ago in order to resolve an issue I had when running eABF with this class.   

These functions are called when `extendedLagrangian` is on and `f_cv_homogeneous=true`:
https://github.com/Colvars/colvars/blob/6ac3a18895b94b80b06b4ec68988092f63cb2a9d/src/colvar.cpp#L2230-L2234
Without overloading, potential energy and gradients in `colvar::update_extended_Lagrangian()`  are computed by calling the functions in the base class (periodicity is not considered), e.g. for the gradient,
https://github.com/Colvars/colvars/blob/6ac3a18895b94b80b06b4ec68988092f63cb2a9d/src/colvarcomp.cpp#L666-L670

In my case with eABF,  the issue turned out to be that due to the wrong gradient the extended variable $x_{ext}$ can not be pulled to cross the boundary after $x=\xi(X_t)$ is already wrapped to the other side of the boundary.

src/colvarcomp.h
giacomofiorin: I think this is fine, too. We have a similar example with Lepton in `colvar.h`.

src/colvarcomp.h
giacomofiorin: I'm noticing here that only a single shift is performed when implementing the nearest-image convention. Given how general this type of variable can be, maybe simply copying the code from `distanceZ` would suffice? 
https://github.com/Colvars/colvars/blob/5e5c1d59021c48fcb25850e02c0f93d641b38f6c/src/colvarcomp_distances.cpp#L320-L329

I'll look into a way to just reuse functions like those in a derived class without resorting to the preprocessor macros.

zwpku: Right, it is safer to handle the general case.  For the moment I just copy the code from distanceZ class.

giacomofiorin:
@zwpku Thank you for handling merge conflicts coming from `master`. PR #643 contains more of them, but that could be merged after this PR to make it easier.

To move forward, can you provide a list of commands to install libTorch, so that the feature can be included in the automated tests? I can help integrate them with the GitHub Actions workflows and the hodgepodge of testing scripts in this repo if needed.

libTorch should ideally be installed from a precompiled package, unless compilation from source can proceed in a reasonable time (e.g. less than 30 minutes) when no cache is present.

giacomofiorin:
Hi @zwpku I'd like to move ahead with #643, which will create some merge conflicts with this branch. To avoid that, it may be better to merge this branch first, if you think it's ready for review. Do you have the PyTorch installation instructions suitable for reproducing the results of the tests that you added?

zwpku:
Hi @giacomofiorin,  I updated this branch with master and included the list of commands for installation at the beginning of this page.  On my local machine, the installation works for both NAMD 3.0b3 and gromacs-2023.2. 

It seems that older versions of NAMD (e.g. versions from year 2022) are no longer supported?

Let me know if any further information about the installation is needed.

giacomofiorin:
Thanks for the quick reply!

NAMD 2.x (i.e. the `master` branch) is effectively abandoned by the NAMD team, which will make all new releases based on the `devel` branch.

giacomofiorin:
@zwpku We have now a small conflict in the GROMACS 2023 patch due to a change in `master`.

Would you have any strong objections against supporting GROMACS >= 2024.0 and only those for this feature? (I can make that change)

zwpku:
@giacomofiorin, sure, it is fine to only support Gromacs >=2024 for this feature.

zwpku:
@HanatoK @giacomofiorin 
An issue I noticed in my tests was that ABF simulations with CV defined by this class slow down quite a bit. 
The PyTorch model I used involves a rotation/translation layer (implemented by myself in Python) and coordinates of backbone atoms. Maybe the computation can be faster when the model is simpler (e.g. functions of CVs). But I guess in general this class via calling libtorch library is slower than the neuralNetwork class.  I wanted to further look into the performance of this class but it didn't happen yet... 

I wonder whether one can implement similar CV classes using APIs from other ML packages, e.g. tensorflow or JAX. Any idea?

HanatoK:
@zwpku I am also working on a new NAMD interface that supports GPU-resident calculations, and I also use libtorch as an example. Based on my experience, it depends on how you implement the rotation/translation layer in python. My impression is that if you use the default automatic diffferentiation then it would be slow. I think the performance is better when using manual derivatives (my code is in https://github.com/HanatoK/namd_pytorchforces_example/blob/1f05b007b724502fccc23e06207f5ac865691285/build_rmsd_cv_opt.py#L143-L155). Have you tried to implement a custom `backward` or something similar?

As for JAX, I am also looking into it, but it seems lacking documentation of its C++ API. By the way, the pytorch community is moving to its new AOTInductor interface, and TorchScript is in maintenance mode, but again the documentation of AOTInductor is a problem.

zwpku:
@HanatoK  @giacomofiorin 
I got a question (not about this class but about the rmsd function) when writing the previous comment and would like to know your answer. To my understanding $RMSD(x) = |x-c(x)-R(x)x_{ref}|^2$, where c(x) and R(x) are the optimal translation and the optimal rotation, respectively.   If this is correct, then computing the derivative of RMSD wrt x would also need to differentiate both c(x) and R(x) by chain rule, since they both depend on x.  
On the other hand, in the code shared by @HanatoK in the comment above, it seems that this dependence is not taken into account (I've only read part of the code so maybe I misunderstood).  Also, in Colvars there seems to be an option in the rmsd CV class that controls whether to handle this dependence.  
It is not clear to me how to understand the code. Did I miss something, or is there some cancellation that allows you to avoid differentiating c(x) and R(x)? Thanks!

HanatoK:
@zwpku The RMSD is something special. I don't think it requires to calculate the derivative of the rotation matrix $R(x,x_\mathrm{ref})$ with respect to $x$ or $x_\mathrm{ref}$. You can see "The algorithm" section in [this paper](https://onlinelibrary.wiley.com/doi/10.1002/jcc.20110), page 1855. The $\mathcal{U}^T(q_\mathrm{max})$ is just the transpose of the rotation matrix.

For other CVs depending on the aligned positions, I think Colvars has indeed the fit gradients for them, but I am not quite sure.

giacomofiorin:
Hi @zwpku! @HanatoK is correct that the RMSD does not normally need to compute those derivatives, because they are zero as long as the atoms being included in the RMSD are the same over which the roto-translation is computed.  If the two are not the same (i.e. if `fittingGroup` is defined) then the fit gradients are non-zero and will need to be calculated.  The same is also true if rotated coordinates are used in a CV that is not the RMSD, which is the only one whose gradients are zero by definition.

Unfortunately, due to limitations in the current data structures we do not have support for computing/storing the fit gradients for vector-valued variables (#87), which includes Cartesian coordinates. But that limitation should not affect you, because you have already defined the `torchANN` function to return a scalar value.

Regarding speed, I would get this PR finished up for setting up NN functions of CVs. But if you want to use Cartesian coordinates directly, I would recommend checking out @HanatoK's preliminary work to that sense. I do acknowledge that you may want to use rotated-frame coordinates, which is not possible now but may become possible later (or is it?).

zwpku:
@HanatoK @giacomofiorin Thank both of you for the explanations. Now the case of RMSD function is clear to me. 
`torchann` defines a function of CVs. To define a function of Cartesian coordinates directly, one can combined it with `cartesian`, as in the example below: 

```
torchann {
    modelFile model.pt
    m_output_index 0
    cartesian {
         atoms { 
         atomNumbers {
             1 5 6 7 9 11 15 16 17 19
         }
    }
}
```

It is not clear whether I can define a function of fitted coodinates by simply turning on the options `rotateToReference` and `centerToReference` in atoms: 

```
  torchann {
      modelFile model.pt
      m_output_index 0
      cartesian {
          atoms { 
              centerToReference yes
              rotateToReference yes
	      atomNumbers {
	          1 5 6 7 9 11 15 16 17 19
	      }
              refPositionsFile all.pdb
           }
    }
```
I will look into the code to better understand it.

HanatoK:
@zwpku I attach a configuration of binding free-energy (BFEE) calculation here for your reference.
```
eulerTheta {                             
        atoms {                               
            indexGroup  ligand                 
            centerReference    on              
            rotateReference    on              
            centerToOrigin     on              
	    enableFitGradients on              
            fittingGroup {                    
                indexGroup  protein            
            }                                 
            refpositionsfile  ./complex_largeBox.xyz        
         }                                    
         refpositionsfile  ./complex_largeBox.xyz           
}
```
The system has two biomolecules, namely protein and ligand. We want to align the system simulated with respect to the reference positions only using the protein, and then measure the relative Euler theta angle of the ligand with respect to the reference.

If you only work with a single molecule then you might not need the `fittingGroup`. `eulerTheta` accepts `refpositionsfile` but your `torchann` don't need it.

vineetbansal:
Can someone tell me which commit of LAMMPS source code this was tested against?

I'm finding that if I run `./update_colvars_code.sh` against LAMMPS commit `a74500` (current as of 07/01/24), and compile/run the example, it causes runtime failures due to `parse_fix_arguments` (which relies on the `proxy` object being initialized because it calls `proxy->set_random_seed(rng_seed)`) being called *before* `proxy = new colvarproxy_lammps(lmp)` line (which actually does the initialization) in `fix_colvars.cpp`.

I can provide more details if necessary, and/or open a new issue on this, but was hoping to get things working against whatever commit of LAMMPS was used here first.

HanatoK:
> Can someone tell me which commit of LAMMPS source code this was tested against?
> 
> I'm finding that if I run `./update_colvars_code.sh` against LAMMPS commit `a74500` (current as of 07/01/24), and compile/run the example, it causes runtime failures due to `parse_fix_arguments` (which relies on the `proxy` object being initialized because it calls `proxy->set_random_seed(rng_seed)`) being called _before_ `proxy = new colvarproxy_lammps(lmp)` line (which actually does the initialization) in `fix_colvars.cpp`.
> 
> I can provide more details if necessary, and/or open a new issue on this, but was hoping to get things working against whatever commit of LAMMPS was used here first.
> 
> EDIT: Looks like this is not an issue with this PR, but exists in the target branch. The following doesn't work:
> 
> ```
> fix myfix all colvars mytorchann.colvars seed 12
> ```
> 
> but the following does:
> 
> ```
> fix myfix all colvars mytorchann.colvars
> fix_modify myfix seed 12
> ```
> 
> So perhaps it's just my misunderstanding about the usage. Sorry for the noise!

I can reproduce the same problem. The issue should be introduced with https://github.com/Colvars/colvars/commit/88730cb1c577977207902ee6a60570ae0ed9ccf5. Since the tests do not use the `seed` argument, the bug has never triggered. You can try to move `parse_fix_arguments(narg, arg, true);` after the `new` statement or ask @giacomofiorin to fix it.

giacomofiorin:
Thanks for catching this bug @vineetbansal ! 

To clarify, the versions of Colvars in this repository (ideally the `master` branch) are meant to work only with the `develop` branch of LAMMPS, with the intention of eventually incorporating the former into the latter.

Unfortunately we are lagging a bit behind for this, due to the significant changes brought up by #418, which introduced some bugs, one of which you graciously reported.

I'll look into resolving the conflicts in this branch after #701 is merged, but if you are in a hurry you could probably just use `git cherry-pick a2d00066` to get the fix in here.

jhenin:
Two items would help here:
1. a way to run the tests in GH actions to catch regressions
2. a way to load a dynamic libtorch, because we probably won't be able to add it as a dependency to the back-end codes.

HanatoK:
>     2. a way to load a dynamic libtorch, because we probably won't be able to add it as a dependency to the back-end codes.

There are two ways: (i) Colvars provides a mechanism to load library dynamically, just like the `LOAD` command in PLUMED (https://www.plumed.org/doc-v2.9/user-doc/html/_l_o_a_d.html); (ii) `torchann` uses a private implementation, and the private implementation does not get compiled with Colvars main source tree and loads the libtorch dynamically.

jhenin:
Of interest here, a libtorch interface is making its way into the Gromacs build system: https://gitlab.com/gromacs/gromacs/-/merge_requests/4551

jhenin:
@zwpku I have made changes to enable building with Gromacs 2024. Had you tested that?
The regtest is broken possibly due to changes in the extended Lagrangian integrator.

zwpku:
> > ```
> > 2. a way to load a dynamic libtorch, because we probably won't be able to add it as a dependency to the back-end codes.
> > ```
> 
> There are two ways: (i) Colvars provides a mechanism to load library dynamically, just like the `LOAD` command in PLUMED (https://www.plumed.org/doc-v2.9/user-doc/html/_l_o_a_d.html); (ii) `torchann` uses a private implementation, and the private implementation does not get compiled with Colvars main source tree and loads the libtorch dynamically.

giacomofiorin:
@zwpku Did you mean to close this PR?

zwpku:
Regarding dynamic libtorch, one issue to me is that, in order to be able to call functions in the loaded library,  one has to declare in the C++ source code of the library with extern "C" so that the names are not mangled.  I found many such declarations in plumed, but only a few in libtorch. So I not sure whether this is supported by libtorch.

HanatoK:
> @jhenin @HanatoK I don't know whether I understood dynamical loading correctly. One issue to me is that, in order to be able to call functions in the loaded library, one has to declare in the C++ source code of the library with extern "C" so that the names are not mangled when creating the .so file. I found many such declarations in plumed, but only a few in libtorch. So I not sure whether this is supported by libtorch.
> 
> Also, it would be good to have a compatible solution with the development of libtorch interface in Gromacs.

You don't have to declare everything in `extern C`. In my opinion, what we should do include:
1. Colvars should have the ability to load a dynamic library that includes a subclass of `colvarcomp`. That means Colvars needs to wrap `dlopen` (for example: https://theo-penavaire.medium.com/loading-of-a-c-class-from-a-shared-library-modern-c-722d6a830a2b and https://github.com/theopnv/Dynamic-Loading). This may requires the help from @giacomofiorin and @jhenin to do some fundamental changes in `add_component_type` of `colvar.cpp`.
2. The `colvarcomp_torchann` class can then be declared in a separate header out-of-tree (no need for `extern C`), and then in the implementation `colvarcomp_torchann.cpp`, there should be creator and deleter functions implemented in `extern C`. For example:
https://github.com/theopnv/Dynamic-Loading/blob/820e15b2ff8af6ae8a75a6cfb92b10e40a02f686/Libraries/Tatooine/Tatooine.cpp#L3-L31

If you have access to the NAMD Gitlab repository, you can have a look with the proposed PytorchForces implementation that also uses the dynamic loading:
https://gitlab.com/tcbgUIUC/namd/-/merge_requests/287/diffs#ae77510c784effb9160a1e44869634b6b0221c8a

zwpku:
> > @jhenin @HanatoK I don't know whether I understood dynamical loading correctly. One issue to me is that, in order to be able to call functions in the loaded library, one has to declare in the C++ source code of the library with extern "C" so that the names are not mangled when creating the .so file. I found many such declarations in plumed, but only a few in libtorch. So I not sure whether this is supported by libtorch.
> > Also, it would be good to have a compatible solution with the development of libtorch interface in Gromacs.
> 
> You don't have to declare everything in `extern C`. In my opinion, what we should do include:
> 
> 1. Colvars should have the ability to load a dynamic library that includes a subclass of `colvarcomp`. That means Colvars needs to wrap `dlopen` (for example: https://theo-penavaire.medium.com/loading-of-a-c-class-from-a-shared-library-modern-c-722d6a830a2b and https://github.com/theopnv/Dynamic-Loading). This may requires the help from @giacomofiorin and @jhenin to do some fundamental changes in `add_component_type` of `colvar.cpp`.
> 2. The `colvarcomp_torchann` class can then be declared in a separate header out-of-tree (no need for `extern C`), and then in the implementation `colvarcomp_torchann.cpp`, there should be creator and deleter functions implemented in `extern C`. For example:
>    https://github.com/theopnv/Dynamic-Loading/blob/820e15b2ff8af6ae8a75a6cfb92b10e40a02f686/Libraries/Tatooine/Tatooine.cpp#L3-L31
> 
> If you have access to the NAMD Gitlab repository, you can have a look with the proposed PytorchForces implementation that also uses the dynamic loading: https://gitlab.com/tcbgUIUC/namd/-/merge_requests/287/diffs#ae77510c784effb9160a1e44869634b6b0221c8a

Thanks! So this means that we compile colvarcomp_torchann class together with libtorch to build a dynamic library, say libcovarcomp_torchann.so. Then Colvars loads this library at runtime?

HanatoK:
> > > @jhenin @HanatoK I don't know whether I understood dynamical loading correctly. One issue to me is that, in order to be able to call functions in the loaded library, one has to declare in the C++ source code of the library with extern "C" so that the names are not mangled when creating the .so file. I found many such declarations in plumed, but only a few in libtorch. So I not sure whether this is supported by libtorch.
> > > Also, it would be good to have a compatible solution with the development of libtorch interface in Gromacs.
> > 
> > 
> > You don't have to declare everything in `extern C`. In my opinion, what we should do include:
> > 
> > 1. Colvars should have the ability to load a dynamic library that includes a subclass of `colvarcomp`. That means Colvars needs to wrap `dlopen` (for example: https://theo-penavaire.medium.com/loading-of-a-c-class-from-a-shared-library-modern-c-722d6a830a2b and https://github.com/theopnv/Dynamic-Loading). This may requires the help from @giacomofiorin and @jhenin to do some fundamental changes in `add_component_type` of `colvar.cpp`.
> > 2. The `colvarcomp_torchann` class can then be declared in a separate header out-of-tree (no need for `extern C`), and then in the implementation `colvarcomp_torchann.cpp`, there should be creator and deleter functions implemented in `extern C`. For example:
> >    https://github.com/theopnv/Dynamic-Loading/blob/820e15b2ff8af6ae8a75a6cfb92b10e40a02f686/Libraries/Tatooine/Tatooine.cpp#L3-L31
> > 
> > If you have access to the NAMD Gitlab repository, you can have a look with the proposed PytorchForces implementation that also uses the dynamic loading: https://gitlab.com/tcbgUIUC/namd/-/merge_requests/287/diffs#ae77510c784effb9160a1e44869634b6b0221c8a
> 
> @HanatoK Thanks! So this means that we compile colvarcomp_torchann class together with libtorch to build a dynamic library, say libcovarcomp_torchann.so. Then Colvars loads this library at runtime?

Yes. That is what I mean. libcovarcomp_torchann.so is linked to libtorch.so (either dynamic or static linking), and Colvars loads libcovarcomp_torchann.so dynamically. I am not sure if Colvars maintainers agree with this idea, though.

giacomofiorin:
> Yes. That is what I mean. libcovarcomp_torchann.so is linked to libtorch.so (either dynamic or static linking), and Colvars loads libcovarcomp_torchann.so dynamically. I am not sure if Colvars maintainers agree with this idea, though.

This is actually question very easy to answer (especially to others that you asked recently on this repo :wink:).

The only arrangements that we have ever followed are set by the maintainers of the packages that Colvars is distributed with. In this specific case, if GROMACS is considering to link libTorch (see above), IMO Colvars should link it in the same manner. That would be statically, dynamically, etc.

@HanatoK have you discussed with your colleagues in the NAMD team about how to package libTorch with NAMD? It would be awesome if it could be packaged statically as you do already for the NVIDIA libraries, FFTW, etc

HanatoK:
> The only arrangements that we have ever followed are set by the maintainers of the packages that Colvars is distributed with. In this specific case, if GROMACS is considering to link libTorch (see above), IMO Colvars should link it in the same manner. That would be statically, dynamically, etc.

But it seems @jhenin wants to dynamically load libtorch, not dynamically or statically link to it. I also think that dynamic loading appears to be the most general and flexible solution since it does not require the MD engine to link to libtorch.

> 
> @HanatoK have you discussed with your colleagues in the NAMD team about how to package libTorch with NAMD? It would be awesome if it could be packaged statically as you do already for the NVIDIA libraries, FFTW, etc

We are not going to package or redistribute libtorch, because we want to use libtorch with CUDA, and libtorch may be compiled with different CUDA architectures. More specifically, `libtorch_cuda.so` may be compiled with `-DTORCH_CUDA_ARCH_LIST`, which depends on the actual GPUs being used. The official libtorch binary might not be the optimal build for all users.

 I plan to distribute the PytorchForces and NAMD headers in its source code form in the NAMD binary tarball, and let the users download and compile libtorch following their official instruction, and then compile the binary plugin of PytorchForces linking with libtorch.

giacomofiorin:
@jhenin's suggestion was meant to avoid maintaining patched versions of the build recipes for each engine, which this branch contains right now. Loading would allow the use of a precompiled libTorch in the CI jobs, i.e. loading would be preferable for the specific purpose of running tests. 

As far as distribution goes, as long as the user is required to build dependencies there is no perfect approach :smile:

HanatoK:
> I did mention dynamic loading. However, linking at build time should be supported when possible. In Gromacs and LAMMPS, we have the benefit of a separate CMake file, so there we can relatively easily detect if torch is available at build time - especially if we coordinate with the Gromacs folks. In NAMD, it's more intrusive, so I don't know how easily the patch to the config script will be accepted. The build system of VMD is its own beast, and distribution is a problem, so dynamic loading makes more sense there.

Do you want to support both dynamic loading and direct linkage? That would make the code more complicated and I cannot see the benefit of using both. You would need a lot of ifdefs to do that. For dynamic loading, if GROMACS is linked to libtorch, then `libcolvarcomp_torchann.so` does not need to link to libtorch since it would find the libtorch symbols when loading in GROMACS. If NAMD is not linked to libtorch, then `libcolvarcomp_torchann.so` needs to link to libtorch if it is used in NAMD.

zwpku:
> @zwpku I have made changes to enable building with Gromacs 2024. Had you tested that? The regtest is broken possibly due to changes in the extended Lagrangian integrator.

@jhenin I tried to patch Gromacs 2024 with this branch. It seems that Gromacs 2024 is viewed as GROMACS-DEV in `update-colvars-code.sh` and the corresponding commands do not update the `gmxManageColvars.cmake` file (which include dependence on `libtorch`). I didn't continue to build the code but I am afraid that the `torchann` class will not be compiled.  Is there some issue in `update-colvars-code.sh`?

jhenin:
Thanks @zwpku , the patch was not in the right place.

jhenin:
The current failures to build gromacs in regtests are due to the lack of downstream changes. Merging master should help.

gromacs/cmake/gmxManageColvars.cmake
```diff
@@ -38,6 +38,19 @@ gmx_option_multichoice(GMX_USE_COLVARS
     INTERNAL NONE)
 mark_as_advanced(GMX_USE_COLVARS)
 
+function(gmx_set_colvars_torch)
+  find_package(Torch)
```
HubLot, line 42: Note: This line will have to be removed for GMX 2025

giacomofiorin: @HubLot I propose to stop patching any build recipe for the engines, and just test the features inside the library itself. I added support for linking against PyTorch in tests in PR #721, and tagged you as reviewer since you already reviewed a similar MR for GROMACS ;-)

HubLot: I agree!

giacomofiorin:
Hi, since we have one CI test working for the library, my only remaining concern would be avoiding routine conflicts with the build systems of the MD engines (see https://github.com/Colvars/colvars/pull/570#discussion_r1778227005).

src/colvarcomp.h
giacomofiorin: Going back to this, I think it would actually be better to isolate the `#include <torch/*.h>` into a source file. Now that the Torch interface is always included in the CI jobs, we have a 3-fold increase in build times when the compiler cache is invalidated.

zwpku: @giacomofiorin I added a separate header file `colvarcomp_torchann.h` to include the torch dependent code. Now COLVARS_TORCH only affects `colvarscomp_torchann.h`, `colvarscomp_torchann.cpp`, and `colvar.cpp`.
The library build time reduces from 23m 44s to  16m 59s.  Are these changes OK?

giacomofiorin:
The latest GROMACS test error is unrelated to Colvars: https://gitlab.com/gromacs/gromacs/-/issues/5204

giacomofiorin:
Hi there!  GROMACS 2025 runs without errors the `torchann` input from this branch, but there are differences from the reference files generated by @zwpku and updated by @jhenin.

See the outputs here:
https://github.com/Colvars/colvars/actions/runs/11747796165/artifacts/2164700677

zwpku:
> Hi there! GROMACS 2025 runs without errors the `torchann` input from this branch, but there are differences from the reference files generated by @zwpku and updated by @jhenin.
> 
> See the outputs here: https://github.com/Colvars/colvars/actions/runs/11747796165/artifacts/2164700677

@giacomofiorin thanks for the work. 
Any hint on what the reason might be? I don't have a good understanding on this regression test... Could it be that the files under gromacs/tests/library/000_torchann/AutoDiff are outdated?

giacomofiorin:
> > Hi there! GROMACS 2025 runs without errors the `torchann` input from this branch, but there are differences from the reference files generated by @zwpku and updated by @jhenin.
> > See the outputs here: https://github.com/Colvars/colvars/actions/runs/11747796165/artifacts/2164700677
> 
> @giacomofiorin thanks for the work! Any hint on what the reason might be? I don't have a good understanding on the regression test... Could it be that the files under gromacs/tests/library/000_torchann/AutoDiff are outdated?

@zwpku Yes, the reference files currently in that folder were produced came from another build, with a different version of libTorch. Would you expect this kind of difference? It is small, but it did exceed our threshold (1.0e-6 relative error).

zwpku:
> > > Hi there! GROMACS 2025 runs without errors the `torchann` input from this branch, but there are differences from the reference files generated by @zwpku and updated by @jhenin.
> > > See the outputs here: https://github.com/Colvars/colvars/actions/runs/11747796165/artifacts/2164700677
> > 
> > 
> > @giacomofiorin thanks for the work! Any hint on what the reason might be? I don't have a good understanding on the regression test... Could it be that the files under gromacs/tests/library/000_torchann/AutoDiff are outdated?
> 
> @zwpku Yes, the reference files currently in that folder were produced came from another build, with a different version of libTorch. Would you expect this kind of difference? It is small, but it did exceed our threshold (1.0e-6 relative error).

@giacomofiorin  If I see correctly, the torch model in that test is simply the identity map and the CV is a dihedral angle. So I expect there should be little difference due to different versions of libTorch.  Could it also be caused by some changes in the source code or in the config files of that test?  I can try to build and examine the test on my local machine.

giacomofiorin:
> > > > Hi there! GROMACS 2025 runs without errors the `torchann` input from this branch, but there are differences from the reference files generated by @zwpku and updated by @jhenin.
> > > > See the outputs here: https://github.com/Colvars/colvars/actions/runs/11747796165/artifacts/2164700677
> > > 
> > > 
> > > @giacomofiorin thanks for the work! Any hint on what the reason might be? I don't have a good understanding on the regression test... Could it be that the files under gromacs/tests/library/000_torchann/AutoDiff are outdated?
> > 
> > 
> > @zwpku Yes, the reference files currently in that folder were produced came from another build, with a different version of libTorch. Would you expect this kind of difference? It is small, but it did exceed our threshold (1.0e-6 relative error).
> 
> @giacomofiorin If I see correctly, the torch model in that test is simply the identity map and the CV is a dihedral angle. So I expect there should be little difference due to different versions of libTorch. Could it also be caused by some changes in the source code or in the config files of that test? I can try to build and examine the test on my local machine.

@zwpku Yes, if you could please check as soon as possible that would be very helpful! We may have a chance to convince the GROMACS people to include this in the 2025 release, but timing is very tight.

giacomofiorin:
Still getting deviations from the reference files that you just uploaded in the CI tests. It is probably due to differences in libTorch versions. The one in the container is 2.4:
https://github.com/Colvars/colvars/blob/4651a5983051986c5196107e38f1834b62fb1a61/devel-tools/containers/CentOS9-devel.def#L65-L68

Is the `.pt` file sensitive to the version? If so, I strongly recommend that this is re-generated using a matching same version of libTorch?

zwpku:
> Still getting deviations from the reference files that you just uploaded in the CI tests. It is probably due to differences in libTorch versions. The one in the container is 2.4:
> 
> https://github.com/Colvars/colvars/blob/4651a5983051986c5196107e38f1834b62fb1a61/devel-tools/containers/CentOS9-devel.def#L65-L68
> 
> Is the `.pt` file sensitive to the version? If so, I strongly recommend that this is re-generated using a matching same version of libTorch? 

@giacomofiorin I guess the original deviation you encountered was (partially) due to random seed. The seed was fixed in gromacs/tests/library/Common/test.mdp in a previous commit by @jhenin, but somehow it was reverted again, possibly due to a merge with master.  
The deviation from yesterday was due to the reference files I uploaded, which were generated on my local machine using Gromacs in single precision (I realized that Gromacs is compiled with double precision when running the CI tests). After changing to double precision and regenerating the reference files, it passes the test.

I tried libtorch 2.0.1, 2.3.0, and 2.4.1, and the results are the same. The `.pt` files generated using PyTorch 1.13.1 and 2.3.1 are different (in size). But when loaded in gromacs/colvars, the results (input/output/gradient) are the same.

zwpku:
@giacomofiorin besides,  I saw some differences (e.g. in colvar.cpp, colvarmodule.cpp, colvarmodule_refs.h) when I ran 
 `git diff master..torchann -- src`. They seem to be unrelated to `torchann` class. Should these files be updated?

giacomofiorin:
> @giacomofiorin besides, I saw some differences (e.g. in colvar.cpp, colvarmodule.cpp, colvarmodule_refs.h) when I ran `git diff master..torchann -- src`. They seem to be unrelated to `torchann` class. Should these files be updated?

Yes. I just did that.

If the tests pass (thank you for addressing the GROMACS precision issue!) we can proceed to merge this PR into `master`. Given the many conflicts accumulated in this branch, we should use a squash merge but link the PR in the commit.

giacomofiorin:
PR merged! Thanks so much @zwpku and everyone who helped getting this PR done!

giacomofiorin:
FYI Lukas Mullender from the GROMACS team raised a couple of comments on the code regarding the use of GPU models and precision:
https://gitlab.com/gromacs/gromacs/-/merge_requests/4780#note_2213094685

PR #572, opened 2023-08-29
Move Python, C_fortran, stubs to misc_interfaces/

jhenin:


jhenin:
I think this is safe to merge, as it does not affect any user-facing functionality.

giacomofiorin (approved):
Not sure that the source files that are already under a subfolder (i.e. `tests`) was as pressing, but the new location is just as good.

PR #579, opened 2023-09-06
Fix several issues preventing reading state from LAMMPS restart file

giacomofiorin:
With both NAMD and LAMMPS now allowing to first initialize Colvars after a simulation has begun (initial step > 0), it's getting harder to get everything right for moving restraints that require knowledge of the step when they were first defined (see #486). 

This PR fixes this behavior by recording that number as `firstStep` and embedding it in the state file. 

For LAMMPS specifically, this now works both with a Colvars state file provided via the `input` keyword or by the LAMMPS binary restart file.


jhenin:
Looks good. Why is the new function `delete_input_stream` necessary? What happens with `close_input_stream`?

giacomofiorin:
> Looks good. Why is the new function `delete_input_stream` necessary? What happens with `close_input_stream`?

It just rewinds the string stream to allow future reads (as when the files are read as strings embedded in a TPR file), but then the condition below remains true, and the input state string keeps being loaded every time `setup_input()` is called:
https://github.com/Colvars/colvars/blob/0a997fc274069970080793863a1e44bd8f2ea986/src/colvarmodule.cpp#L1313-L1320

The above is clearly an issue that needs to be fixed, either by adding a new function (as in this case) or a separate flag that is activated once and deactivated after the state is read. This is where a second set of eyes is most helpful ;-)

jhenin (approved):

PR #584, opened 2023-09-15
Merge GROMACS MDModules interface

giacomofiorin:
This PR integrates into the Colvars repo the MDModules-based interface for GROMACS written by @HubLot ([link](https://gitlab.com/gromacs/gromacs/-/merge_requests/3671)).

The existing CI tests are expanded to this interface, using the GROMACS `main` branch as target. Note that only the existing library tests are currently run this way. In parallel, @HubLot is adding separate unit tests for the interface source code ([link](https://gitlab.com/gromacs/gromacs/-/merge_requests/3671)), to be run as part of the GROMACS testing infrastructure using CMake/CTest. 

Besides adding CI support this PR also contains a few fixes for the existing interface code:
- ~~Integration of recently added `colvarproxy::input_stream_from_string()` into the MDModules based proxy~~ EDIT: superseded by binary checkpointing
- ~~Add a missing `colvars:` prefix to the GROMACS log messages~~ EDIT: this fix has been imported into the [GROMACS MR](https://gitlab.com/gromacs/gromacs/-/merge_requests/3671)
- ~~Because the GROMACS `std::string` serializer only supports strings up to 4096 bytes (detected because two state files were too big), even the formatted checkpointing functions feature an explicit loop over bytes.~~ EDIT: superseded by binary checkpointing
- ~~GROMACS' `keyvaluetreeserializer.cpp` lacks a template instantiation for `char`, so the `gromacs-mdmodules.patch` has been modified accordingly. It would be easier to add these changes to main GROMACS itself via a small focused MR.~~ EDIT: this has been merged into upstream GROMACS ([link](https://gitlab.com/gromacs/gromacs/-/merge_requests/3816)).
- Replace the TPR prefix with the EDR prefix (a [MR](https://gitlab.com/gromacs/gromacs/-/merge_requests/3826) is pending for GROMACS).
- To minimize conflicts at this time, I have temporarily disabled Lepton in the GROMACS patching procedure.


.github/workflows/test-backends.yml
```diff
@@ -99,40 +99,14 @@ jobs:
...
-      rpath_exe: install/bin/gmx_d
-
-  gromacs-2023:
-    name: GROMACS 2023
-    if: github.event_name == 'pull_request' || contains(github.event.head_commit.message, 'test-gromacs-2023')
-    uses: ./.github/workflows/backend-template.yml
-    with:
-      backend_name: GROMACS-2023
-      backend_repo: gromacs/gromacs
-      backend_repo_ref: release-2023
```
HubLot, line 135: these tests workflow needs to be kept.

devel-tools/compile-gromacs.sh
```diff
@@ -98,25 +103,22 @@ compile_gromacs_target() {
...
-        ${CTEST} --output-on-failure
-        retcode=$?
-        if [ -n "${GMX_INSTALL_DIR}" ] ; then
-            if hash ninja >& /dev/null ; then
-                ninja install
-            else
-                make install
-            fi
-            ret_code=$((${ret_code} || $?))
+        if [ -n "${GROMACS_BUILD_TESTS}" ] && [ -z "${GITHUB_ACTION}" ] ; then
```
HubLot, line 113: This condition will never be true since GROMACS_BUILD_TESTS is set only when GITHUB_ACTION is set.
So, on Github Actions, this will just build the tests. 
I don't really understand the purpose.

gromacs/gromacs-mdmodules/applied_forces/colvars/colvarsforceprovider.h
```diff
@@ -0,0 +1,210 @@
...
+    /*! \brief String naming variable holding the last known whole positions of the colvars atoms
+     * \note Changing this name will break backwards compability for checkpoint file writing.
+     */
+    static const std::string xOldWholeName_;
+
+    /*! \brief Content of the unformatted Colvars state file.
+     */
+    std::vector<unsigned char> colvarStateFile_;
+
+    /*! \brief String naming variable holding the content of the formatted Colvars state file.
```
HubLot, line 89: "formatted" here and below whereas "unformatted" is written just above

HubLot (commented):
overall, looks good to me.

devel-tools/compile-gromacs.sh
giacomofiorin: The purpose is to allow the CI jobs to build the tests (to catch potential errors), but not actually run them on GitHub Actions, because they are not supported by the GROMACS team for Linux. On top of that, we had numerous failures for standard GROMACS (`main` branch) on Windows and macOS.

Perhaps in the future we could run the GROMACS tests for stable release versions, but for `main` they generate too many false positives IMO.

HubLot: okay I understand.
but then, remove this condition no? Because i don't see the case where the condition is met and it's confusing imo.

giacomofiorin: It would be met when we run the tests locally, no?

HubLot: oh yes indeed but only if the env variable GROMACS_REGRESSION_TESTS is set.
okay i see.

gromacs/gromacs-mdmodules/applied_forces/colvars/colvarsforceprovider.h
giacomofiorin: Done, thank you!

giacomofiorin:
@HubLot @jhenin Since this is mostly CI and small fixes please take a second look. If everything is okay please merge into `master`.

I have merged back changes by @HubLot from the GROMACS MR but have not done the reverse yet. This PR also contains the EDR notification change, which hopefully should get into GROMACS `main` soon. (I would like to keep the changes to GROMACS files to a minimum in @HubLot's open MR).

jhenin (approved):

PR #589, opened 2023-09-20
Reset error code when invoking colvarmodule::reset()

giacomofiorin:
When `reset()` is called it is not an automatic operation but it is invoked by design, so it should be interpreted as part of the caller's way to deal with whatever error was there. Directly affects GROMACS interface, VMD CLI interface and Dashboard GUI.


HubLot:
Looks good to me. It solves my issue with the GROMACS test framework

HubLot (approved):

jhenin (approved):

PR #591, opened 2023-09-24
Allow redefining outputPrefix in NAMD

### LINKED ISSUE #561: Change of output prefix by NAMD script is ignored by Colvars
A NAMD script of the form:
```
#define colvars config
outputName name1
run 20
outputName name2
run 20
```
will write all output under `name1.colvars.*`. This cannot be bypassed by a script because the prefix is only updated by the colvarproxy_namd constructor, and `cv reset` does not delete the proxy object.

Possible fixes:
- have cvm::setup_output() call the proxy to give it a chance to do this kind of update
- provide a script interface to update manually

Only the first option would give the behavior that I expect as a user.

giacomofiorin:
Fixes #561 

Another minor fix is to skip writing output files from objects (colvars, biases) if `cvm::output_prefix()` is empty.


jhenin (approved):
Yup!

PR #573, opened 2023-08-29
Default param for set_unit_system

jhenin:
Note: this branch is based on cleanup_dirs, see PR #572

PR #594, opened 2023-09-27
Silence clang-tidy warnings

giacomofiorin:
Silencing some warnings generated during build jobs of the [GROMACS interface](https://gitlab.com/gromacs/gromacs/-/merge_requests/3671), specifically by library header files  included by files under `src/gromacs/applied_forces/colvars`.

There are certainly more that could be silenced, but those included here are the those that seemed to me more substantial than cosmetic.

@jhenin One of the warnings is about the default argument in `set_unit_system()` just introduced by you. Do you want to revert #573 or just ignore it?


jhenin:
I see the problem, following your suggestion.

PR #601, opened 2023-10-10
Add Tcl 8.6 to CentOS 7 container (NAMD 3.x now requires it)

giacomofiorin:
The updated definition file included here was already used to update the [CentOS 7 image](https://cloud.sylabs.io/library/giacomofiorin/default/colvars_development). 

Eventually it would be good to cache these images locally using the GHCR, but that takes a bit more work. Trying to limit the number of large GH caches would seem more pressing. 

@jhenin Can you please check and merge if the tests pass?

### DIFF
--- a/.github/workflows/backend-template.yml
+++ b/.github/workflows/backend-template.yml
@@ -87,7 +87,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/.apptainer
-          key: Linux-x86_64-containers-build
+          key: Linux-x86_64-containers-build-code
 
       - name: Checkout OpenMM (for Lepton library)
         uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


giacomofiorin:
@jhenin I canceled the workflow because the new key fooled the Actions cache workflow into picking the wrong one.

jhenin (approved):
Working, thank you!

giacomofiorin:
@jhenin Sorry, but there is yet another instance of an annoying "feature" of GitHub Actions (i.e. the kind that you'd spell with the letters "b", "u" and "g"...). 

The workflow for the backends was wrong (I put an `env` keyword in a place GH didn't like) and so its results were not considered :roll_eyes:

PR #608, opened 2023-10-19
Enable address sanitizer (via Clang) in CI

giacomofiorin:
Enables `-fsanitize=address` via environment variable. ~~At the moment the new test will fail unless the small memory leaks are addressed.~~ EDIT: now the test passes, merging.

Also disabling `--coverage` in the default configuration, to minimize potential interference between the two tools. (TODO: we still do not have coverage metric being computed).


giacomofiorin:
@jhenin Now there is only one run that still leaks, `read_xyz_traj`. Can you look into it?

After that we have all sanitizer tests passing (amazingly so).

PR #609, opened 2023-10-20
Unit test for incorrect input

jhenin:
This PR is meant to track down memory leaks in the special case of input errors.


giacomofiorin:
@jhenin I fixed the handling of unknown keywords that broke after you opted to delete a group when errors show up.

IMO a better approach would be initializing group objects empty and just keeping them in memory at all times via smart pointers. This way we wouldn't need to keep checking if the pointers to them are valid. But that would take a bit to implement.

For the moment, I just prevented a call to `check_keywords()` in case of an earlier error, because it would raise a false error in the RMSD constructor, which can't properly initialize its registry of valid keywords. This is not ideal.

PR #610, opened 2023-10-21
Allow using a Colvars input state file instead of binary checkpoint in GROMACS

giacomofiorin:
See discussion at: https://gitlab.com/gromacs/gromacs/-/merge_requests/3671#note_1577988430

This PR implements that requirement, but in a different manner than was mentioned before: a new keyword internal to the Colvars config file that specifies a state file to be used _unless_ a state is read in any other way, as provided by the MD engine. After the a state is read the first time, the default input state file name is deleted (so that it won't be loaded again).

The new keyword is currently not tested in CI yet. I feel that it would be confusing to expose it to engines other than GROMACS, which don't need it. Currently, it's only _documented_ for that engine. @jhenin What are your thoughts on TinkerHP? 

Note below how the two engines that provide their own restart file differ significantly:
- GROMACS needs a way to continue a simulation when a Colvars config file is changed or when a different build of GROMACS is used (both of which prevent restarting Colvars from the checkpoint file). The proposed change would be to use a default state file for Colvars~~keyword `colvars-inputstatefile` in the MDP~~, which will be superseded when a checkpoint file is read.
- LAMMPS currently allows (since #558) restarting from a Colvars input state file or the LAMMPS restart file. When both are provided, the former overrides the latter, i.e. this behavior is _opposite_ to what seems most appropriate for GROMACS.

~~While I finish the implementation of this feature for GROMACS, I'd appreciate feedback on whether the above inconsistency should be resolved or not.  Mainstream LAMMPS does not contain yet the Colvars code currently in `master`, so there is time to change.~~
EDIT: see here https://github.com/Colvars/colvars/pull/610#issuecomment-1787485033

### DIFF
--- a/doc/colvars-refman-main.tex
+++ b/doc/colvars-refman-main.tex
@@ -1016,18 +1016,17 @@ Both versions support loading Colvars state files in either format (binary or te
 
 In some cases, it is useful to modify the configuration of variables or biases between consecutive runs, for example by adding or removing a restraint.
 Some special provisions will happen in that case.
-When a state file is loaded, no information is available from in regarding the any newly added variable or bias, which will remain uninitialized until the first compute step.
-Conversely, any information that the state file contains about variables or biases that are no longer defined will be silently ignored.
-Because these checks are performed based only on the \emph{names} of variables and biases, it is your responsibility to ensure that these names have \emph{consistent definitions between runs.}
+When a state file is loaded, no information is available about any newly added variable or bias, which will thus remain uninitialized until the first compute step.
+Conversely, any information that the state file may contain about variables or biases that are no longer defined will be silently ignored.
+Please note that these checks are performed based only on the \emph{names} of variables and biases: it is your responsibility to ensure that these names have \emph{consistent definitions between runs.}
 
-The flexibility just described carries some limitations: namely, it is only supported when reading \emph{text-format} state files.
+The flexibility just described carries some limitations: namely, it is only supported when reading \emph{text-format} Colvars state files.
 Instead, restarting from binary files\cvlammpsonly{ (such as the LAMMPS restart file)}\cvgromacsonly{ (such as the GROMACS checkpoint file)} after a configuration change will trigger an error.
 It is also important to remind that when switching to a different build of \MDENGINE, the binary format may change slightly, even if the release version is the same.
 
-To work around the potential issues just described, a text-format Colvars state file should be used.
-\ifdefined\cvvmdornamd{This is the default in \MDENGINE{} unless the ``\texttt{COLVARS\_BINARY\_RESTART}'' is set to 1, and this information is only provided for troubleshooting.}\fi\cvlammpsonly{This can be achieved by providing an explicit \texttt{input} keyword when initializing the Colvars fix (see \ref{sec:colvars_mdengine_parameters}), which will instruct Colvars to use the given filename instead of the LAMMPS restart file.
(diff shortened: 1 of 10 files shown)


giacomofiorin:
Note: converting to draft because the first commit undoes the TPR -> EDR change, this is a conflict that needs to be resolved.

jhenin:
The issue is, LAMMPS lets the user specify new options for each run even after a restart, so if they mention a state file, it should have priority status - whereas Gromacs doesn't, so if a cpt file can be read with success, the state file could be obsolete from a previous run.

giacomofiorin:
Merge conflicts resolved, so I was able to add a needed change, i.e. preventing the GROMACS proxy from setting the initial step of the simulation when it doesn't know it anyway. I think that should be addressed separately, but out of scope for this PR.

Also removed the LAMMPS tag since we agreed that the apparent inconsistency is not confusing.

doc/colvars-refman-main.tex
```diff
@@ -1009,42 +1012,58 @@
...
+It is also important to remind that when switching to a different build of \MDENGINE, the binary format may change slightly, even if the release version is the same.
+
+To work around the potential issues just described, a text-format Colvars state file should be used.
+\ifdefined\cvvmdornamd{This is the default in \MDENGINE{} unless the ``\texttt{COLVARS\_BINARY\_RESTART}'' is set to 1, and this information is only provided for troubleshooting.}\fi\cvlammpsonly{This can be achieved by providing an explicit \texttt{input} keyword when initializing the Colvars fix (see \ref{sec:colvars_mdengine_parameters}), which will instruct Colvars to use the given filename instead of the LAMMPS restart file.
+  Furthermore, the \texttt{fix\_modify} scripting command allows to load a Colvars file after initialization (\ref{sec:cv_command_loadsave}).}\cvgromacsonly{In GROMACS, this step would normally entail generating a new TPR file: however, there is no provision to embed a text-format Colvars state file in a TPR file.
```
jhenin, line 1029: I would remove the sentence "In Gromacs... TPR file", which explains the rationale of the design, but is not directly useful to users. I'd go directly to telling users what to do.

doc/colvars-refman-main.tex
```diff
@@ -1009,42 +1012,58 @@
...
+  For this specific use case, the following Colvars option may be used:
+\begin{itemize}
+\item %
+  \labelkey{Colvars-global|defaultInputStateFile}
+  \key{defaultInputStateFile}{%
+    global}{%
+    Default input state file, if not provided by \MDENGINE}{%
+    UNIX filename}{%
+    Define a state file that will be loaded by default, unless \MDENGINE{} provides restarting information through the checkpoint.
+  }
```
jhenin, line 1039: This keyword is defined for all back-ends. Should it be part of the doc for all back-ends? Even if the other back-ends have other ways of specifying this, it could integrate better in some workflows.

jhenin (commented):
Great! Just two suggestions on the documentation.

doc/colvars-refman-main.tex
giacomofiorin: That's kind of related to my question to you about Tinker-HP. It's the only other engine where this keyword might be useful, I see really no use for it in LAMMPS, NAMD or VMD.

If it's really GROMACS specific right now, I would add a check to make it tied to it. We could always change it later if needed.

giacomofiorin:
@jhenin I made the doc edits and restricted the new keyword only to the GROMACS >= 2024 case.

I leave open in case you have suggestions about TinkerHP, otherwise I'm good with merging.

jhenin:
Colvars/Tinker-HP reads a `colvars.state` file based on its prefix. I don't think there are specific needs on that side.

PR #615, opened 2023-10-30
Collected ABF improvements

jhenin:
# Changes in mwABF

## New, expanded mwABF script
for selection rules, now implemented for more than 1 CV

## Output of per-replica data in mWABF
useful for error analysis

## proper handling of CZAR files in mw-eABF

# Architecture changes

## Dependency system
The ABF class now has sub-features with respect to the colvarbias class.

# TODOs
- [x] Update ABF doc for mwABF changes
- [x] Update mwABF test

[EDIT: removed experimental feature that doesn't need merging now]


jhenin:
Finally cleaned up, with a rather long and complex commit history. Probably needs a good squashing.

giacomofiorin:
> Finally cleaned up, with a rather long and complex commit history. Probably needs a good squashing.

Ideally it would be great if you could resolve the conflicts manually and rebase, but squashing sounds fine if you think it'd take too much work.

Keeping a linear history for `master` as much as possible looks even more important now to be able to backport bugfixes into previous releases. (LAMMPS already started issuing updates to stable releases a while ago, so it's not even just GROMACS).

jhenin:
I've tried rebasing this, without success. The history is too long and entangled with that of master.

jhenin:
Now brutally squashed. That's the best I could do.

jhenin:
Updated the commit message to just describe the net changes and remove all the noise from the squashed commits, some of which reverted others.

src/colvarbias_abf.h
```diff
@@ -184,6 +196,38 @@ class colvarbias_abf : public colvarbias {
...
+  /// Alternately we could overload the init_dependencies() function
+  virtual int init_dependencies();
+
+  enum features_bias_abf {
+    /// Start after generic cvb features
+    /// There is at least one ext-Lagrangian colvar -> run eABF
+    f_cvb_abf_extended = f_cvb_ntot,
+    /// Total number of features for an ABF bias
+    f_cvb_abf_ntot
+  };
```
giacomofiorin, line 211: This change looks more or less okay to me with the code as it is. However, because we can finally rely on C++11 or later, I think it would be great to use `enum class` for the dependency features, and this kind of initialization would prevent it.

On the other hand, we could stick with a simple `int` based `enum`, but if another bias had its own set of specialized features, its values would clash with the ABF ones.

Lastly, the extended Lagrangian was even introduced originally for metadynamics (before most people realize it doesn't add much benefit). So it seems be more perplexing to me to introduce sub-features just for this particular example.

How about adding instead `f_cvb_extended_Lagriangian` as a generic bias feature?

src/colvargrid.h
```diff
@@ -1017,21 +1061,99 @@ class colvar_grid_count : public colvar_grid<size_t>
     has_data = true;
   }
 
+  /// \brief Return the average number of samples in a given "radius" around current bin
+  /// Really a hypercube of length 2*radius + 1
+  inline int local_sample_count(int radius)
```
giacomofiorin, line 1066: It would be nice to progressively move bigger functions out of colvargrid.h and into colvargrid_def.h (thus leaving the former mostly for declarations). I don't feel strongly about it if there are good reasons to do otherwise, though.

src/colvargrid.h
```diff
@@ -490,6 +526,13 @@ template <class T> class colvar_grid : public colvarparse {
     data[i] = t;
   }
 
+ /// Get the value at the point with linear address i (for speed)
+  inline T get_value(size_t i)
```
giacomofiorin, line 530: Can this function be `const`?

src/colvargrid.h
```diff
@@ -1394,6 +1573,13 @@ class colvar_grid_gradient : public colvar_grid<cvm::real>
   /// Constructor from a multicol file
   colvar_grid_gradient(std::string &filename);
 
+  /// Constructor from a vector of colvars and a pointer to the count grid
+  colvar_grid_gradient(std::vector<colvar *> &colvars, colvar_grid_count *samples_in);
+
+  /// Parameters for smoothing data with low sampling
+  int full_samples;
```
giacomofiorin, line 1580: We can initialize plain old data in the header now! :-)

src/colvargrid.h
```diff
@@ -1474,6 +1660,15 @@ class colvar_grid_gradient : public colvar_grid<cvm::real>
     }
   }
 
+  /// \brief Report the offset between the lower end of the bin and the current value of variable i
+  inline cvm::real current_offset(int const i) const
```
giacomofiorin, line 1664: Maybe the function name could be a bit more descriptive, since elsewhere you have started indexing a colvargrid as a linear array with an integer?

src/colvargrid.h
```diff
@@ -1606,6 +1861,63 @@ class integrate_potential : public colvar_grid_scalar
...
   }
 
+  /// \brief Return the value at the upper edge/corner of the given bin,
+  /// linearly interpolated (averaged) between 2^nd neighboring bins
+  /// Used to interpolate to a grid shifted by a half-bin
+  /// e.g. from PMF to gradient grid
+  /// should not be used on upper edges of non-periodic grids
+  /// (that should never be necessary, since those are extra bins with respect
+  /// to underlying gradient grid)
+  inline cvm::real value_corner_interp(std::vector<int> const &ix,
```
giacomofiorin, line 1871: Do you think this could be moved to the class template? I'd like to use it as a basis for a function to interpolate intermediate values of a grid for a more continuous metadynamics potential.

namd/tests/interface/000_stringstates/namd-version.txt
```diff
@@ -1,3 +1,3 @@
-Info: NAMD 2.14b1 for Linux-x86_64-multicore
-colvars: Initializing the collective variables module, version "2020-04-09".
-colvars: Using NAMD interface, version "2020-04-07".
+Info: NAMD 3.0b4 for Linux-x86_64-multicore
+colvars: Initializing the collective variables module, version 2023-10-03.
+colvars: Using NAMD interface, version "2023-10-03".
```
giacomofiorin, line 3: Are the changes to this test anything more than cosmetic?

namd/tests/library/multiple_walker_abf/run_mw_tests.sh
```diff
@@ -11,16 +11,25 @@ if (( $# != 1 )); then
 fi
 NAMD_EXEC=$1
 
-#NAMD_EXEC=$HOME/Software/NAMD_2.10b1_Source_colvars/Linux-x86_64-g++/namd2
+# Runner for the netlrts version
+RUNNER="${HOME}/progs/namd/Linux-x86_64-g++.netlrts/charmrun ++local +p4"
+
+# Runner for the MPI version
+#RUNNER="mpirun -n 4"
```
giacomofiorin, line 18: Is there any chance that this could be run as part of the CI, using the MPI build of NAMD?

giacomofiorin (changes requested):
Looks mostly good to me. All my concerns are minor, except for the one about the `enum` values.

Disclaimer: given the size of this PR I couldn't spend a lot of time looking at the body of the new functions. I assume that all the changed code is currently being exercised in tests.

src/colvargrid.h
jhenin: Good point. I think the largest can even go to the cpp file, avoiding the need for an extra header. I'll give it a try.

src/colvargrid.h
jhenin: Certainly

namd/tests/library/multiple_walker_abf/run_mw_tests.sh
jhenin: Yes, that would be an excellent idea.

src/colvarbias_abf.h
```diff
@@ -137,11 +128,29 @@ class colvarbias_abf : public colvarbias {
...
+  // eABF/CZAR local data last shared
+  std::unique_ptr<colvar_grid_gradient> last_z_gradients;
+  std::unique_ptr<colvar_grid_count>    last_z_samples;
+  // ABF data from local replica only in shared ABF
+  std::unique_ptr<colvar_grid_gradient> local_gradients;
+  std::unique_ptr<colvar_grid_count>    local_samples;
+  std::unique_ptr<integrate_potential>  local_pmf;
+  // eABF/CZAR data collected from all replicas in shared eABF on replica 0
+  // if non-shared, aliases of regular CZAR grids, for output purposes
+  colvar_grid_gradient *global_z_gradients;
```
HanatoK, line 149: I see the lines above use smart pointers like `std::unique_ptr`, but here the raw pointers are still used. Could they be shared pointers?

src/colvarbias_abf.h
HanatoK: I second the use of `enum class`.

src/colvarbias_abf.cpp
```diff
@@ -285,63 +334,22 @@ int colvarbias_abf::init(std::string const &conf)
...
-
-  if (last_gradients) {
-    delete last_gradients;
-    last_gradients = NULL;
-  }
-
-  if (system_force) {
-    delete [] system_force;
-    system_force = NULL;
+  #define delete_if_non_null(ptr) if (ptr) { delete ptr; ptr = nullptr; }
```
HanatoK, line 337: It could be better if `nullptr` and `NULL` can be unified.

HanatoK (approved):
Looks good to me!

HanatoK:
One thing that I am not sure is the compatibility of the UI estimator and multiple-walker eABF. Maybe @fhh2626 could comment on it?

fhh2626:
> One thing that I am not sure is the compatibility of the UI estimator and multiple-walker eABF. Maybe @fhh2626 could comment on it?

Previously, the UI and CZAR estimators both yield the results locally. If this is a problem for the new update (as there will be a .all.grad file after the commit), I think the simplist way is to disable the UI estimator in MW-eABF to avoid inconsistency.

src/colvarbias_abf.h
jhenin: Initially I did this extension to add experimental features that have been removes since. I am not worried about clashes, because the extended features of each derived class would only be used within the derived class, so reusing the same values would have no consequence. But it would interfere with moving to enum class. This one feature can certainly be added to the generic set. I suppose that even features that are unique to a single derived class can be added to the generic set, and unused by most derived classes. It seems like pollution of the base class to me, though.

src/colvargrid.h
jhenin: Turns out this code is unused in the current version, it was used in an experimental feature that got pruned. But a more general linear interpolation function would be very useful, and this combined with current_offset would do the trick.

src/colvargrid.h
jhenin: Also unused now. This should become an ingredient for a new function, see below.

src/colvargrid.h
jhenin: Wouldn't the initialization list be even cleaner though? I think they should be gathered in one place.

src/colvarbias_abf.h
jhenin: Change done.

namd/tests/interface/000_stringstates/namd-version.txt
jhenin: Yes: there is a change due to a fix that removed double counting of the first timestep in repeated runs.

src/colvar.cpp
```diff
@@ -1722,8 +1724,18 @@ int colvar::calc_colvar_properties()
     // Special case of a repeated timestep (eg. multiple NAMD "run" statements)
     // revert values of the extended coordinate and velocity prior to latest integration
     if (cvm::proxy->simulation_running() && cvm::step_relative() == prev_timestep) {
-      x_ext = prev_x_ext;
-      v_ext = prev_v_ext;
+      // Detect jumps due to discrete changes in coordinates (eg. in replica exchange schemes)
+      cvm::real const jump2 = dist2(x, x_old) / (width*width);
+      if (jump2 > 0.25) {
```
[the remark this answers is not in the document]
jhenin: 0.25 is just a heuristic. It may detect the problem and create an error, but it will not fix it.

src/colvarbias_abf.cpp
```diff
@@ -55,32 +80,33 @@ int colvarbias_abf::init(std::string const &conf)
...
     cvm::log("Jacobian (geometric) forces will be included in reported free energy gradients.\n");
   }
 
-  get_keyval(conf, "fullSamples", full_samples, 200);
-  if ( full_samples <= 1 ) full_samples = 1;
-  min_samples = full_samples / 2;
-  // full_samples - min_samples >= 1 is guaranteed
+  full_samples = 200;
+  get_keyval(conf, "fullSamples", full_samples, full_samples);
+  get_keyval(conf, "minSamples", min_samples, full_samples / 2);
```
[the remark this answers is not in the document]
jhenin: Good catch!

src/colvarbias_abf.cpp
jhenin: I agree, now there is a mix but the NULL's could be changed to nullptr. @giacomofiorin , do you agree?

giacomofiorin: @jhenin Yes, totally. I just don't bring this up explicitly because it's quite straightforward to run [`clang-tidy`](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-nullptr.html).

Even better to use smart pointers, though, in which case the `NULL` could server as a reminder ;-)

src/colvargrid.h
jhenin: I'll keep only the functions I want inlined there, with the inline qualifier to make that clear.

jhenin: Some larger functions can be moved out, but that should happen in a separate PR.

src/colvarbias_abf.h
```diff
@@ -202,23 +202,15 @@ class colvarbias_abf : public colvarbias {
...
-  enum features_bias_abf {
-    /// Start after generic cvb features
-    /// There is at least one ext-Lagrangian colvar -> run eABF
-    f_cvb_abf_extended = f_cvb_ntot,
-    /// Total number of features for an ABF bias
-    f_cvb_abf_ntot
-  };
-
   /// \brief Implementation of the feature list for colvarbias_abf
   static std::vector<feature *> cvb_abf_features;
```
giacomofiorin, line 206: Sorry, I may be missing something, but is this still needed if the features are all stored in the base class?

giacomofiorin (changes requested):

src/colvarbias_abf.h
jhenin: Currently there is a specific dependency:
```
require_feature_self(f_cvb_history_dependent, f_cvb_get_total_force);
```
But I can bypass the dependency system for this specific case.

giacomofiorin: Ok, thanks, I get it now (the vector is a static member). Makes sense, I just hope this doesn't bite us back in the future with a subtle bug.

giacomofiorin (approved):

giacomofiorin:
@jhenin All good with me on this. If you agree with @fhh2626's suggestion to make UI and mwABF incompatible, can you add that change and merge?

src/colvarbias_abf.h
jhenin: Done.

jhenin:
Thank you all for the reviewing effort!

giacomofiorin:
> Thank you all for the reviewing effort!

Thank you for dealing with all the merge conflicts!

To be clear, I wasn't feeling strongly about the overridden accessor hiding the base-class member vector of features. By mentioning the "subtle bug" I was kinda hoping to exorcise it, since the bugs we get tend to be the kind that we _don't_ expect :-D

PR #616, opened 2023-10-30
Update multisim test for GMX MdModules

jhenin:
The test works now. Had to refresh the inputs, but the potential energy value is correct in each replica.
Not included in CI yet - that would require an MPI build of GROMACS.


giacomofiorin:
> The test works now. Had to refresh the inputs, but the potential energy value is correct in each replica. Not included in CI yet - that would require an MPI build of GROMACS.

LAMMPS was allowing that locally, I just enabled it in CI with #617. There are still some warnings from OpenMPI that we may be able to silence soon.

gromacs/tests/interface/010_MPI_multi-sim/run.sh
```diff
@@ -5,7 +5,9 @@
 # It's best to have Gromacs compiled in double precision
 # Reference files have been generated with Gromacs version 2020.6
```
HubLot, line 6: it should be updated to 2024-dev

jhenin:
@giacomofiorin Do you think it is worth including an MPI build of Gromacs in the CI for this test?

giacomofiorin:
> @giacomofiorin Do you think it is worth including an MPI build of Gromacs in the CI for this test?

Yes. Eventually we can look into silencing the warnings that you can see from the LAMMPS test runs in GH Actions, but the tests work for that backend already :-)

giacomofiorin:
Ok I updated the GROMACS build to MPI per @jhenin's request, and the tests succeed locally but fail on GitHub. Will investigate tomorrow.

HubLot:
Looks like the failed jobs builds a MPI version of GROMAC and the binary passed to `run_tests` is non-MPI

HubLot:
So the question is : which tests should be run agains the serial version of GROMACS and which tests should be run against the MPI version of GROMACS. 

The `compile-gromacs.sh` could be modified to take an optional 2nd argument to activate or not MPI.

One solution could be:
    - library tests : only serial
    - interface tests: serial & MPI

jhenin:
Solving a merge conflict, I had to decide whether to keep running tests over 2 OpenMP threads, as was added in the meantime. I've kept them for now.

jhenin:
Still getting "mpirun: command not found". I don't get it.

giacomofiorin:
> Still getting "mpirun: command not found". I don't get it.

The executables from the OpenMPI package are not in `$PATH` by default. The library runner script has the following lines to check whether you are running a thread-MPI build or a proper MPI one:
https://github.com/Colvars/colvars/blob/22be65e1b262a837e81514f8d043e1b56604ed50/gromacs/tests/library/run_tests.sh#L57-L67

I simplified the runner in the `interface` folder to reuse the one from the library with the exception of the test that requires MPI.

giacomofiorin:
P.S. I didn't try tweaking the `mpirun` command line options for the multisim test, but it might need `-oversubscribe`.

jhenin:
Thanks @giacomofiorin 
We do need oversubscribe apparently:
```
There are not enough slots available in the system to satisfy the slots that were requested by the application:
```
Despite the test failing, the CI workflow succeeds because the return value from the scripts is not passed on.

gromacs/tests/interface/010_MPI_multi-sim/run.sh
```diff
@@ -10,7 +10,7 @@ if [ "$1" ]
 then
     COMMAND=$1
 else
-    COMMAND="mpirun -np 16 gmx_mpi_d"
+    COMMAND="mpirun -np 16 -oversubscribe mpirun gmx_mpi_d"
```
HubLot, line 13: i think there is one extra 'mpirun'

HubLot:
```suggestion
    COMMAND="mpirun -np 16 -oversubscribe gmx_mpi_d"
```

HubLot:
i think we need the absolute path of `gmx_mpi_d` for the multisim

HubLot:
I added a GROMACS 2024 job in addition to main since the beta is released now.

HubLot (approved):

giacomofiorin:
Hey! I didn't have a chance to comment yet. The PR looks good, but I do not think that GROMACS 2024 needs to be tested, since most new code in the Colvars `master` will never become part of that GROMACS release.

Instead, the `release-2024` branch of GROMACS should be tested against the `gromacs-2024` branch of Colvars, where bugfixes will go. Do you agree?

jhenin:
Yes, sounds good to me!

Are we ok that we can dispense with testing against GROMACS 2022 though? I don't see how we can break it without breaking 2023 too.

PR #622, opened 2023-11-15
Apply @zwpku's fix to colvargid::get_grad

### LINKED ISSUE #621: issue of integrate_potential::get_grad for non-periodic cv
I was doing ABF simulations in NAMD using a non-periodic CV and restarting from a previous run. It seems that the function,  invoked by pmf->set_div(), 

https://github.com/Colvars/colvars/blob/67244e8df8385e784aef63fb29663d462fed1707/src/colvargrid.cpp#L667-L676

is unsafe, since Line 673 will invoke the function
https://github.com/Colvars/colvars/blob/67244e8df8385e784aef63fb29663d462fed1707/src/colvargrid.h#L58-L65
which will return an incorrect address `addr`  when the index `ix` is on the edge of the grid, e.g. ix=(-1,0).  This may lead to the access of wrong address in memory (maybe harmless though), or a fatal error when debug mode is on (my case). 

Would it be better to implement `get_grad` as below?

        size_t count, i;
        bool edge = gradients->wrap_edge(ix); 
        if (!edge) {
            if (gradients->samples) {
              count = gradients->samples->value(ix);
            } else {
              count = 1;
            }
            cvm::real fact = 0.0;
            if (count) fact = 1.0 / count;
            cvm::real const *grad = &(gradients->value(ix));

            for ( i = 0; i<nd; i++ ) 
              g[i] = fact * grad[i];
      } else {
            for ( i = 0; i<nd; i++ ) 
              g[i] = 0.0;
        }

jhenin:
to avoid accessing illegal memory locations.
Fixes #621

### DIFF
--- a/src/colvargrid.cpp
+++ b/src/colvargrid.cpp
@@ -666,24 +666,23 @@ void integrate_potential::update_div_neighbors(const std::vector<int> &ix0)
 
 void integrate_potential::get_grad(cvm::real * g, std::vector<int> &ix)
 {
-  size_t count, i;
-  bool edge = gradients->wrap_edge(ix); // Detect edge if non-PBC
+    size_t count, i;
+    bool edge = gradients->wrap_edge(ix);
+    if (!edge) {
+        if (gradients->samples) {
+          count = gradients->samples->value(ix);
+        } else {
+          count = 1;
+        }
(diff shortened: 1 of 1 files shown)


jhenin:
As far as I'm concerned, this can be merged.

PR #624, opened 2023-11-22
Fix stalling NAMD jobs in CI

giacomofiorin:
After confirming that the recent locks only occur when running on multiple threads, I converted the NAMD build recipe to be consistent with LAMMPS and GROMACS (from #616).

It would seem difficult to try and investigate the source of the problem given the Russian-doll combination of Charm++, the container runtime, and Microsoft's Ubuntu VM running on a physical server. 

Although this fixes the CI lockouts, there is a significant drawback in that the SMP code path is no longer exercised. Maybe the MPI and multicore versions can both be run?

### DIFF
--- a/.github/workflows/test-backends.yml
+++ b/.github/workflows/test-backends.yml
@@ -48,7 +48,7 @@ jobs:
       path_compile_script: devel-tools/compile-namd.sh
       test_lib_directory: namd/tests/library
       test_interface_directory: namd/tests/interface
-      rpath_exe: Linux-x86_64-g++.multicore/namd2
+      rpath_exe: Linux-x86_64-g++.mpi/namd2
     secrets:
       # Choice of license by UIUC prevents sharing the code, hence the secret
       private_key: ${{ secrets.PULL_NAMD_KEY }}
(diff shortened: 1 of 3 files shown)


jhenin:
Thanks, this is beginning to cost us a lot of time!

I agree that the SMP code paths need to be tested, they are very much used in practice and the potential for bugs is non-zero. The multicore version should be run too, but don't we risk just bringing back the CI stalls?

giacomofiorin:
> I agree that the SMP code paths need to be tested, they are very much used in practice and the potential for bugs is non-zero. The multicore version should be run too, but don't we risk just bringing back the CI stalls?

That may be the case, unfortunately. I had a half idea about running NAMD differently using the netlrts version but that may not work either.

However, a possible alternative would be to just give up on doing this on the Actions VMs, and run instead the OpenMP code paths in LAMMPS and GROMACS. While checking that I realized that fixes are needed soon for both to (re)enable OpenMP. So I'm merging this to restore the NAMD CI jobs, knowing that the SMP code path would be tested in the other engines.

PR #626, opened 2023-11-23
Remove duplicate symlink creation in CI

### LINKED ISSUE #625: Race condition in CMake symlink creation in CI tests
We have random CI test failures with this error:
`CMake Error: failed to create symbolic link '/home/runner/work/colvars/colvars/build_clang++7_C++17/tests/unittests/index.ndx': file already exists`
e.g. this:
https://github.com/Colvars/colvars/actions/runs/6902606975/job/18779678433?pr=616

The reason is explained
[ here:](https://discourse.cmake.org/t/race-condition-for-e-create-symlink-duplicate-calls/7827)

The custom scripts for unit tests `read_xyz_traj` and `parse_error` create duplicate symlinks.

jhenin:
Fixes #625

### DIFF
--- a/tests/unittests/CMakeLists.txt
+++ b/tests/unittests/CMakeLists.txt
@@ -1,5 +1,7 @@
 set(COLVARS_STUBS_DIR ${COLVARS_SOURCE_DIR}/misc_interfaces/stubs/)
 
+add_custom_target(link_files)
+
 foreach(CMD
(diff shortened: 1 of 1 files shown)


giacomofiorin:
Note: I prefer removing the bugfix label, since this is a workaround for a GitHub-specific problem, not an error that would affect users in any way.

PR #627, opened 2023-11-24
Gmx no cv config test

jhenin:
Will become relevant when https://gitlab.com/gromacs/gromacs/-/merge_requests/3955 is merged.


jhenin:
Is relevant now.

giacomofiorin:
Looks good in general, but in detail, is using precompiled TPR files intentional?

jhenin:
Ah yes, those should be untracked. We had a policy of tracking them long ago when the tests targeted only stable Gromacs releases, but as we've seen, that does't fly anymore and they have to be generated on the fly.

giacomofiorin:
> Ah yes, those should be untracked. We had a policy of tracking them long ago when the tests targeted only stable Gromacs releases, but as we've seen, that does't fly anymore and they have to be generated on the fly.

Agreed. If we need to track a specific code version, let's do it via the release branches, like `gromacs-202{4,5}`.

PR #628, opened 2023-11-25
Fix OpenMP support in LAMMPS and GROMACS

giacomofiorin:
This PR enables (or restores, in the case of LAMMPS) the OpenMP flag for the Colvars library in the CMake recipes of GROMACS and LAMMPS, when the same flag is activated for either engine.


jhenin (approved):
Looking good.

PR #629, opened 2023-11-25
Define programmatically the available CVC types, report them in case of errors

giacomofiorin:
Small but overdue improvement, in the form of introspection code to provide clearer error messages when no suitable function is given for a colvar. This is motivated to improve clarity in light of the fact that the availability of features is increasingly varied between engines/platforms.

@jhenin I was considering exposing the list of components and their descriptions via the scripting interface. Would that be useful for the Dashboard, and if so in what format?

### DIFF
--- a/.github/workflows/test-backends.yml
+++ b/.github/workflows/test-backends.yml
@@ -60,8 +60,7 @@ jobs:
       backend_repo_ref: devel
       path_compile_script: devel-tools/compile-namd.sh
       test_lib_directory: namd/tests/library
-      # Interface tests disabled until map variables are merged into NAMD3
-      # test_interface_directory: namd/tests/interface
+      test_interface_directory: namd/tests/interface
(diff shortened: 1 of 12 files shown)


jhenin:
Thanks @giacomofiorin . Just now I don't see a clear usage for the Dashboard to have a list of components, even though it's certainly possible to imagine one, the sky being the limit on that side. I think that would be useful in the context of "deep colvar inspection". Right now the Dashboard is agnostic to the colvar's components, for simplicity's sake, even though it does try to guess when colvars are orientation quaternions, for example. In general the colvars are visible through the scripting interface as monolithic objects, none of their internals are exposed.
For the time being I would keep this PR without the scripting feature, which can easily be added if and when the need arises.

giacomofiorin:
I merged this accidentally... The changes are too small to warrant reverting, but making a note just in case.

PR #631, opened 2023-11-28
Several CI ccache related fixes and improvements

giacomofiorin:
The most significant change is the removal of the conditions that runs the backend jobs only on pull requests. This should allow their cache output to become associated with the `master` branch, which in turn would be available to CI jobs in other branches.

Also included are fixes to ensure correct ccache behavior when building the backends, including in the rare case when there is initially no cache.


giacomofiorin:
Merging immediately because there is no effect on the code itself (only on the CI jobs' internal caching).

PR #640, opened 2023-12-05
Use 64bit integers to store prev timestep

### LINKED ISSUE #639: ABF histogram overflow shortly after 2 microseconds
Hi!

I have several independent simulations (Colvars 2023-10-03, Gromacs 2023.2, WTM-eABF with distanceZ colvar), that run fine for 2 microseconds, but eventually cause some sort of overflow in the ABF histogram. See screenshot attached. The histogram did not have a spike when starting the run. Contrary to what one might believe from the histogram, the colvars trajectory is not stuck at this position. The wrong counts lead eventually to artifacts in the free energy, presumably because CZAR works with them.

Since all my simulations have this issue shortly after 2 microseconds (1 fs timestep), and since the histogram appears to overflow at the bin corresponding to the value of the collective variable at step ~2^32/2, I am suspecting some sort of Int32 overflow, but with my limited knowledge of the code, I haven't found any possible candidate.

colvars abf count
colvars traj
Raw files are on [Google Drive](https://drive.google.com/drive/folders/1pyzQeAqt1YSrmodTBJyOhqfegtb-JBEf?usp=share_link)

jhenin:
Should fix #639

### DIFF
--- a/gromacs/gromacs-2022.x/colvarproxy_gromacs.h
+++ b/gromacs/gromacs-2022.x/colvarproxy_gromacs.h
@@ -32,7 +32,7 @@ protected:
   bool colvars_restart;
   std::string config_file;
   size_t restart_frequency_s;
-  int previous_gmx_step;
+  int64_t previous_gmx_step;
   double bias_energy;
 
   bool gmx_bNS; // Is this a neighbor-search step? Eventually will become unnecessary
--- a/gromacs/gromacs-2020.x/colvarproxy_gromacs.cpp
+++ b/gromacs/gromacs-2020.x/colvarproxy_gromacs.cpp
@@ -391,7 +391,6 @@ void colvarproxy_gromacs::update_data(const t_commrec *cr, int64_t const step, t
     if (first_timestep) {
       first_timestep = false;
     } else {
-      // Use the time step number inherited from GROMACS
       if ( step - previous_gmx_step == 1 )
         colvars->it++;
       // Other cases?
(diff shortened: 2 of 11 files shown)


giacomofiorin:
Looks good to me! 

For reference:
- NAMD was not affected by the bug because `size_t` is already 64-bits (only unsigned).
- LAMMPS was affected, although it's unclear how frequently: jobs of > 2^32 steps are very rare or the patched GROMACS versions. What about GROMACS 2024?

jhenin:
In GROMACS 2024, Colvars gets the current time step explicitly through the ForceProvider API, so not affected.

jhenin:
Does the standard specify that size_t is 64 bit? Maybe in practice only?
https://en.cppreference.com/w/c/types/size_t

giacomofiorin:
> Does the standard specify that size_t is 64 bit? Maybe in practice only? https://en.cppreference.com/w/c/types/size_t

I believe you'd have to go to a 32-bit architecture to find a narrower type for size_t.

gromacs/gromacs-2022.x/colvarproxy_gromacs.h
```diff
@@ -32,7 +32,7 @@ class colvarproxy_gromacs : public colvarproxy, public gmx::IForceProvider {
   bool colvars_restart;
   std::string config_file;
   size_t restart_frequency_s;
-  int previous_gmx_step;
+  int64_t previous_gmx_step;
```
HanatoK, line 35: Why is `int64_t` used instead of `cvm::step_number`?

giacomofiorin: I think this is harmless. If you remember, we typedef'ed `step_number` to equal `long long` before finding the guts to finally break with ancient C++ standards.

We may just as well use `int64_t` throughout now...

jhenin: @HanatoK Because that is the type of the data that we get from GROMACS.

PR #643, opened 2023-12-19
Several improvements to colvar component classes

giacomofiorin:
This PR contains several improvements to `colvar::cvc` and its derived classes, including substantial simplification of the code and some fixes where the previous implementation was incorrect (`distanceVec` metrics) or less robust (`period`).
- There is now only one type of CVC constructor, which takes no arguments: this removes the need to maintain two different types of constructor. `init(conf)` now initializes an object from the configuration string, by modifying existing defaults.
- Default values for plain old data are set right where the member is declared in the header, at long last! :-)
- (Fix) The `distanceVec` metric functions are now consistent with the documentation.
- There is no longer a need to explicitly derive the metric functions: the base class assumes a scalar-valued function, and only special cases are overridden. *A ton* of duplicated functions were removed with this.
- `period` can now be set only for those CVCs that allow it: that includes `distanceZ` as well all periodic angles (only where someone needs something else than 360°).
- `std::shared_ptr` is now used to collect the CVC objects inside the colvar level. I have not yet touched the arrays of CVC pointers inside complex CVC objects (e.g. the protein CVCs or the combination ones) because those will be modified anyway in upcoming work.

Other possible improvements that were out of scope to do:
- Because all CVCs have now a derived `init()` function, it would be theoretically possible to call that from `colvar::update_cvc_config()`, i.e. the backend for the `modifycvcs` scripting command. Currently, the base class method `cvc::init()` is still called, because not all the derived ones are idempotent yet.
- Atom groups could now be tracked much better with smart pointers, allowing them to be shared to save data/computation (this is done in #644).
- `colvarcomp.h` could be split up to reduce interdependencies.
- Feature objects could definitely use some smart-pointer love, too.

### DIFF
--- a/src/colvarcomp.cpp
+++ b/src/colvarcomp.cpp
@@ -29,20 +29,6 @@ colvar::cvc::cvc()
 }
 
 
-colvar::cvc::cvc(std::string const &conf)
-{
-  description = "uninitialized colvar component";
-  b_try_scalable = true;
-  sup_coeff = 1.0;
-  sup_np = 1;
-  period = 0.0;
-  wrap_center = 0.0;
-  width = 0.0;
-  init_dependencies();
-  colvar::cvc::init(conf);
-}
-
-
 int colvar::cvc::update_description()
 {
   if (name.size() > 0) {
@@ -171,7 +157,7 @@ cvm::atom_group *colvar::cvc::parse_group(std::string const &conf,
                                           char const *group_key,
                                           bool optional)
 {
-  int &error_code = init_code;
+  int error_code = COLVARS_OK;
 
   cvm::atom_group *group = nullptr;
   std::string group_conf;
--- a/src/colvarcomp.cpp
+++ b/src/colvarcomp.cpp
@@ -32,36 +32,33 @@ colvar::cvc::cvc()
 int colvar::cvc::update_description()
 {
   if (name.size() > 0) {
-    description = "cvc " + name;
+    description = "cvc \"" + name + "\"";
   } else {
     description = "unnamed cvc";
   }
-  if (function_type.size() > 0) {
-    description += " of type \"" + function_type + "\"";
-  } else {
-    description += " of unset type";
-  }
+  description += " of type \"" + function_type() + "\"";
   return COLVARS_OK;
+}
+
+
+std::string colvar::cvc::function_type() const
+{
+  if (function_types.empty()) {
+    return "unset";
+  }
+  return function_types.back();
 }
 
 
 int colvar::cvc::set_function_type(std::string const &type)
 {
-  function_type = type;
-  if (function_types.size() == 0) {
-    function_types.push_back(function_type);
-  } else {
-    if (function_types.back() != function_type) {
-      function_types.push_back(function_type);
-    }
-  }
+  function_types.push_back(type);
   update_description();
-
+  cvm::main()->cite_feature(function_types[0]+" colvar component");
   for (size_t i = function_types.size()-1; i > 0; i--) {
     cvm::main()->cite_feature(function_types[i]+" colvar component"+
                               " (derived from "+function_types[i-1]+")");
   }
-  cvm::main()->cite_feature(function_types[0]+" colvar component");
   return COLVARS_OK;
 }
 
@@ -498,7 +495,7 @@ void colvar::cvc::collect_gradients(std::vector<int> const &atom_ids, std::vecto
 void colvar::cvc::calc_force_invgrads()
 {
   cvm::error("Error: calculation of inverse gradients is not implemented "
-             "for colvar components of type \""+function_type+"\".\n",
+             "for colvar components of type \""+function_type()+"\".\n",
              COLVARS_NOT_IMPLEMENTED);
 }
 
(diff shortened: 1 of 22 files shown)


src/colvarcomp_gpath.cpp
```diff
@@ -317,13 +353,13 @@ colvar::gzpath::gzpath(std::string const &conf): CartesianBasedPath(conf) {
         cvm::log(std::string("Geometric path z(σ) will use the square of distance from current frame to path compute z\n"));
     }
     if (total_reference_frames < 2) {
-        cvm::error("Error: you have specified " + cvm::to_str(total_reference_frames) + " reference frames, but gzpath requires at least 2 frames.\n");
-        return;
+        error_code |= cvm::error("Error: you have specified " + cvm::to_str(total_reference_frames) + " reference frames, but gzpath requires at least 2 frames.\n", COLVARS_INPUT_ERROR);
```
HanatoK, line 356: Why is the return statement moved to the end of the function? I am a bit worried that if it doesn't return here, the following code would segfault or crash, which is OK for NAMD but not a good user experience for VMD.

src/colvarcomp_gpath.cpp
```diff
@@ -153,8 +164,25 @@ void colvar::CartesianBasedPath::computeDistanceBetweenReferenceFrames(std::vect
...
+
+colvar::gspath::gspath()
+{
     set_function_type("gspath");
+}
+
+
+int colvar::gspath::init(std::string const &conf)
+{
+    int error_code = CartesianBasedPath::init(conf);
```
HanatoK, line 184: Maybe we need to check the error code and return early too? If you think it is too complicated to return error code in derived classes, we can throw exceptions and catch them in some of the upper level functions.

HanatoK (changes requested):
Some of the original constructor code with `cvm::error` return as early as possible, because continue to execute the code following `cvm::error` will result in a crash. When moving to an `init` function this behavior should keep the same. Otherwise VMD users may experience crashes after loading the wrong data into some of the CVs, for example, a `pathFile` with the wrong format for `gpathCV`.

src/colvarcomp_gpath.cpp
giacomofiorin: Dude, are you really advocating for supporting VMD and and propose adding exceptions in the same review round? ;-)

src/colvarcomp_gpath.cpp
giacomofiorin: Ok with me to return early, I just overlooked the possibility of a crash.

giacomofiorin: Changed the instances that you pointed out into an early return with error.

src/colvarcomp_gpath.cpp
HanatoK: Sorry, I should be clearer. Exception is just one of the solutions, and if I were you I might use exceptions in any `init` functions and catch them in `colvar::update_cvc_config`. Early return is another one, which may need more lines of code for both base and derived classes. Using either of them should be enough. As for VMD, it is just an example of interactive applications. Colvars as a library should not terminate the main interactive program if there is an error.

giacomofiorin: I should be clearer, too :-) Exceptions would be convenient on their own, actually. We haven't used them in the past mostly to play it safe with VMD, which traditionally aims to support exotic and/or ancient compilers. Although compiler support has become more uniform with C++11, best practices like the [Google Style Guide](https://google.github.io/styleguide/cppguide.html#Exceptions) don't encourage vendors to support them well.

I think that adding early returns wherever it looks like the code may crash is probably the safest right now. I think that's still a minority of all error conditions.

HanatoK (approved):

giacomofiorin:
@HanatoK Thanks for revising the PCV constructors!

src/colvarcomp.cpp
```diff
@@ -198,6 +196,8 @@ cvm::atom_group *colvar::cvc::parse_group(std::string const &conf,
                                COLVARS_INPUT_ERROR);
       delete group;
       group = nullptr;
+      // Silence unused variable warning; TODO stop returning a pointer
```
jhenin, line 199: Could this function become an `atom_group` constructor? That would seem the most logical to me.

src/colvarcomp_gpath.cpp
jhenin: This is a common pattern in several places, I made similar "early return" changes in #650 and Haochuan added a bunch more in this PR.
I wouldn't mind having this in more places, that is, everywhere, and I wonder if there is a more concise / elegant / modern / systematic way of formulating this, besides exceptions. The C programmer in me would make it a macro... There, I said it, I'm old. C++11 best practices and Google Style Guide: get off my lawn.

giacomofiorin: Just to be clear, I'm old too but I would be happy to embrace anything that the maintainers of a major package (whatever their age might be) will commit to approve.

*In writing, please.*

src/colvarcomp.cpp
giacomofiorin: Yes, totally! However, I was also thinking about removing all those raw pointers and simply having the following as a member of the `cvc` base class:
```
std::map<std::string, std::shared_ptr<cvm::atom_groups>> groups_;
```
Then after that, it'll be easier for the CVCs to become lighter objects that operate on shared groups.

All that seems a bit out of scope and this PR is fat enough IMO.

jhenin (approved):

PR #648, opened 2024-01-16
Allow NumPy-format output for plot_colvars_traj.py

giacomofiorin:
As described in the title, allows converting segments of a Colvars traj into a more efficient binary format supported by the quintessential Python numerical analysis library.

### DIFF
--- a/colvartools/plot_colvars_traj.py
+++ b/colvartools/plot_colvars_traj.py
@@ -3,7 +3,7 @@
 # Invoke this script with --help for documentation.
 
 # Download link: https://github.com/Colvars/colvars/blob/master/colvartools/plot_colvars_traj.py?raw=true
-# This version was modified on: 2023-07-13
+# This version was modified on: 2024-01-16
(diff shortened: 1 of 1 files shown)


giacomofiorin:
Second commit is because I encountered a syntax error using a Python interpreter from Anaconda, even though it was 3.8, i.e. a version that supports the syntax. Weird.

HanatoK:
I would recommend using `pandas` and saving the data with `to_parquet`.

giacomofiorin:
> I would recommend using `pandas` and saving the data with `to_parquet`.

Sure, but how about in addition to NumPy rather than instead of it? Do you want to add the code to this branch? Should be pretty quick since you already added the hooks for Pandas.

HanatoK:
> > I would recommend using `pandas` and saving the data with `to_parquet`.
> 
> Sure, but how about in addition to NumPy rather than instead of it? Do you want to add the code to this branch? Should be pretty quick since you already added the hooks for Pandas.

That sounds good. I will submit another PR after this one is merged.

PR #649, opened 2024-01-19
Lambda-ABF implementation for the NAMD interface

jhenin:
As fully documented in the [Lagardère et al. paper](https://pubs.acs.org/doi/10.1021/acs.jctc.3c01249)
complements the Tinker-HP implementation, which is maintained on the [Tinker-HP repo](https://github.com/TinkerTools/tinker-hp).

TODO
- [x] Regression tests
- [x] Document current limitations in ref. manual

Update: why are there no regression tests? Because of a synchronization issue in NAMD, where the current alchemical colvar interface is plugged into a "non-urgent" output routine, which causes it to run in a variable order with respect to the Colvars calculation. This causes variations in the instantaneous trajectory, even though it cancels out of the ensemble averages, so the free energies are correct. This causes failures in regression tests, which are by design sensitive to trajectory details. The GPU-resident code path of NAMD is immune from this problem and produces reproducible trajectories: it is therefore the recommended way to use this code in NAMD.


fhh2626:
Hi @jhenin , I just read your excellent λ-ABF paper and then found this PR. I can't wait to give it a try. Is there any guide about how to write the Colvars file?
Thanks,
Haohao

jhenin:
Hi Haohao,
The NAMD implementation still needs some refinement: it comes with performance limitations, namely, you cannot use MTS and you need energy computation every time step. Other than that, it works perfectly, and the performance is not so bad at all.

Here is an example input (using HMR):
```
timestep                2.0
fullElectFrequency      1
nonbondedFreq           1

# For NAMD3
computeEnergies 1

alch                    on
alchFile                        $alchconfig
alchCol                 B                    
alchOutFreq             1000

alchOutFile             $outName.fepout
alchElecLambdaStart     0.5                  
alchVdwLambdaEnd        1.0            
alchVdwShiftCoeff       5.0         
alchdecouple            on

alchType                TI
alchlambda 0

colvars                on 

cv config "
smp off  # Necessary for shared ABF
colvar {
  name l
  extendedLagrangian on
  extendedMass 150000
  extendedLangevinDamping 1000

   lowerBoundary 0
   upperBoundary 1
   reflectingLowerBoundary
   reflectingUpperBoundary
   width 0.01
  alchLambda {
  }
  outputTotalForce
  outputAppliedforce
}

 abf {
   colvars l
   fullSamples 1000
   outputFreq  50000
   historyFreq 100000
   shared on
   sharedFreq 5000
}
"

cv configfile $CVconfig
cv configfile common/protein_tilt.colvars

cv config "
colvarsTrajFrequency 100
colvarsRestartFrequency 100000"

```

fhh2626:
> Hi Haohao, The NAMD implementation still needs some refinement: it comes with performance limitations, namely, you cannot use MTS and you need energy computation every time step. Other than that, it works perfectly, and the performance is not so bad at all.
> 
> Here is an example input (using HMR):
> 
> ```
> timestep                2.0
> fullElectFrequency      1
> nonbondedFreq           1
> 
> # For NAMD3
> computeEnergies 1
> 
> alch                    on
> alchFile                        $alchconfig
> alchCol                 B                    
> alchOutFreq             1000
> 
> alchOutFile             $outName.fepout
> alchElecLambdaStart     0.5                  
> alchVdwLambdaEnd        1.0            
> alchVdwShiftCoeff       5.0         
> alchdecouple            on
> 
> alchType                TI
> alchlambda 0
> 
> colvars                on 
> 
> cv config "
> smp off  # Necessary for shared ABF
> colvar {
>   name l
>   extendedLagrangian on
>   extendedMass 150000
>   extendedLangevinDamping 1000
> 
>    lowerBoundary 0
>    upperBoundary 1
>    reflectingLowerBoundary
>    reflectingUpperBoundary
>    width 0.01
>   alchLambda {
>   }
>   outputTotalForce
>   outputAppliedforce
> }
> 
>  abf {
>    colvars l
>    fullSamples 1000
>    outputFreq  50000
>    historyFreq 100000
>    shared on
>    sharedFreq 5000
> }
> "
> 
> cv configfile $CVconfig
> cv configfile common/protein_tilt.colvars
> 
> cv config "
> colvarsTrajFrequency 100
> colvarsRestartFrequency 100000"
> ```

Thank you very much! Just one additional question. I need to use HMR to enable a timestep of 2fs?

jhenin:
In alchemical simulations, I usually scale down the timestep, because one of the end points has a molecule in gas phase. Depending on the nature of the ligand, you may get stable simulations at larger timesteps.

fhh2626:
Thanks for your clarification!

fhh2626:
Hi, @jhenin ,
I suggest also make available the Colvars harmonic restraints as the alchemical CVs, which will let us completely give up FEP/TI in binding free energy calculations.
Haohao

jhenin:
> Hi, @jhenin , I suggest also make available the Colvars harmonic restraints as the alchemical CVs, which will let us completely give up FEP/TI in binding free energy calculations. Haohao

Thanks Haohao, I like the idea.

src/colvardeps.h
```diff
@@ -289,8 +297,10 @@ class colvardeps {
...
     /// the center
     f_cv_extended_Lagrangian,
-    /// \brief An extended variable that sets an external variable in the
-    /// back-end (eg. an alchemical coupling parameter for lambda-dynamics)
+    /// \brief A variable that constrains or follows an external parameter
+    /// in the back-end (eg. an alchemical coupling parameter for lambda-dynamics)
+    /// If extended Lagrangian, then we drive the external parameter
+    /// Otherwise we follow it
     /// Can have a single component
     f_cv_external,
```
giacomofiorin, line 305: I would be okay with this in a pinch, but ideally I'd like to distinguish it from other "external" objects that are computed in a reproducible way (centers of mass, volumetric maps). Are you using this just to circumvent the check below?
```
        cvm::error("Error: the calculated value of colvar \""+name+
                   "\":\n"+cvm::to_str(x)+"\n differs greatly from the value "
                   "last read from the state file:\n"+cvm::to_str(x_restart)+
                   "\nPossible causes are changes in configuration, "
                   "wrong state file, or how PBC wrapping is handled.\n",
                   COLVARS_INPUT_ERROR);
```

src/colvardeps.h
```diff
@@ -263,8 +265,11 @@ class colvardeps {
     f_cv_active,
     /// \brief Colvar is awake (active on its own accord) this timestep
     f_cv_awake,
-    /// \brief Gradients are calculated and temporarily stored, so
-    /// that external forces can be applied
+    /// \brief External force can be applied, either to atoms or to an
+    /// extended DOF
+    f_cv_apply_force,
```
giacomofiorin, line 270: This is a great addition

src/colvarbias.cpp
```diff
@@ -93,11 +93,18 @@ int colvarbias::init(std::string const &conf)
...
   colvar_values.resize(num_variables());
   for (i = 0; i < num_variables(); i++) {
     colvar_values[i].type(colvars[i]->value().type());
     colvar_forces[i].type(colvar_values[i].type());
     previous_colvar_forces[i].type(colvar_values[i].type());
+    if (!colvars[i]->is_enabled(f_cv_total_force_current_step)) {
+      // If any colvar does not have current-step total force, then
+      // we can't do step 0 data
+      feature_states[f_cvb_step_zero_data].available = false;
+    }
```
giacomofiorin, line 107: `f_cvb_step_zero_data` is also a feature used in the absence of total forces, e.g. adding a hill at step 0 for metadynamics. I'm not a fan of enabling it for that use case, but I think this check would break it.

src/colvarcomp.cpp
```diff
@@ -525,7 +524,7 @@ void colvar::cvc::calc_force_invgrads()
 
 void colvar::cvc::calc_Jacobian_derivative()
 {
-  cvm::error("Error: calculation of inverse gradients is not implemented "
+  cvm::error("Error: calculation of Jacobian derivatives is not implemented "
```
giacomofiorin, line 527: :+1:

src/colvardeps.cpp
```diff
@@ -92,6 +92,8 @@ void colvardeps::restore_children_deps() {
 
 void colvardeps::provide(int feature_id, bool truefalse) {
   feature_states[feature_id].available = truefalse;
+  // Make sure that we don't leave this feature enabled
+  if (!truefalse) disable(feature_id);
```
giacomofiorin, line 96: :100:

namd/src/colvarproxy_namd.h
```diff
@@ -46,6 +46,9 @@ class colvarproxy_namd : public colvarproxy, public GlobalMaster {
   /// Pointer to the NAMD simulation input object
   SimParameters *simparams;
 
+  /// Pointer to Controller object
+  Controller const *controller;
```
giacomofiorin, line 50: Nitpick: I think we should start using `unique_ptr` for these members. Doesn't have to be now of course, but do you see any issue with this in the future?

src/colvar.cpp
```diff
@@ -1238,10 +1245,16 @@ int colvar::init_dependencies() {
...
     init_feature(f_cv_homogeneous, "homogeneous", f_type_static);
 
-    // because total forces are obtained from the previous time step,
-    // we cannot (currently) have colvar values and total forces for the same timestep
     init_feature(f_cv_multiple_ts, "multiple_timestep", f_type_static);
-    exclude_feature_self(f_cv_multiple_ts, f_cv_total_force_calc);
+
+    // when total atomic forces are obtained from the previous time step,
+    // we cannot (currently) have colvar values and projected total forces for the same timestep
+    // TODO this will need refining for driven alchemical parameters
```
giacomofiorin: What do you mean by "driven alchemical parameters"?

src/colvar.cpp
```diff
@@ -2524,8 +2543,14 @@ std::string const colvar::get_state_params() const
...
-     << std::setw(cvm::cv_width)
-     << x << "\n";
+     << std::setw(cvm::cv_width);
+  if (is_enabled(f_cv_external)) {
+    // For an external colvar, x is one timestep in the future after integration
+    // write x at beginning of timestep
+    os << x_reported << "\n";
+  } else {
+    os << x << "\n";
+  }
```
giacomofiorin, line 2569: Same here as in my other comment about the addition of the new feature to `colvardeps`. "External" is a fairly generic term, which includes both quantities that are always available to Colvars (e.g. centers of mass and volmaps going through ComputeGlobal) and those like the lambda parameter that are not.  Here, this check only applies to the latter.

giacomofiorin (commented):
Looks good overall save for some minor comments.

src/colvardeps.h
jhenin: It is used in many places in colvar.cpp. Another way to describe this feature is: this colvar is not a function of atom coordinates, it contains a special CVC that is communicated to and from the back-end through a separate mechanism. I'm happy to rename it if you have a better name, but I don't see an alternative to this feature.

src/colvarbias.cpp
jhenin: Ok, I'm moving this check to `colvarbias_abf`.

namd/src/colvarproxy_namd.h
jhenin: Agreed! I think that should be its own update where we do the change in bulk.

src/colvar.cpp
jhenin: "Driven" means Colvars is responsible for integrating (driving) the dynamics of this parameter. For the current version of this, see  https://github.com/Colvars/colvars/blob/5e37f65d764c38eec892fadd753095747eeb37d1/src/colvar.cpp#L332

src/colvar.cpp
jhenin: Good point, this test really applies to *driven* parameters, which means
` is_enabled(f_cv_external) && is_enabled(f_cv_extended_Lagrangian)`
Updating.

PR #658, opened 2024-02-11
Allow empty string as cvc config

### LINKED ISSUE #394: Cannot create object with empty config block on single line
E.g. a CVC without parameters (yes that is possible now) cannot be defined as:
`functionType {}`
because the parser tries to interpret "}" as a keyword.
So the block has to be written
```
functionType {
}
```
Which is not great, but not terrible. I suspect it could be fixed with minor changes to `key_lookup`, but I'm reluctant to touch that function.

jhenin:
Current error message was unclear (no valid cvc)
Missing parameters will be caught individually.
Useful for alchLambda cvc, which needs no config.

Fixes #394


giacomofiorin (approved):
Good change and (more importantly) very easy to review :-)

Even if there were are any side effects not caught by current tests, #643 will change things much further anyway.

PR #660, opened 2024-02-16
Dashboard button to load colvars traj

jhenin:
This exposes the functionality of the script load_colvar_traj in the Dashboard.

Comes with an update and fixes to the script itself, which I am keeping to allow for non-interactive use.


jhenin:
No interaction with any compiled code, merging.

PR #666, opened 2024-02-27
Protein-specific coordinates based on index groups

jhenin:
The goal is to make the alpha_angles CVC (and dihedralPC) available on all back-ends.

TODO:
- [x] allow for series of groups to describe non-contiguous alpha segments
- [x] extend to dihedralPC
- [x] update documentation
- [x] regression & unit tests
- [ ] VMD Dashboard template (non blocking)


giacomofiorin:
A general comment: besides making those functions more portable there would be major advantages in making *all* CVCs rely on atom groups.

I think it would be great to just use the three groups `groupCA`, `groupN` and `groupO` consistently in all backends, and just have VMD and NAMD auto-fill them to preserve backward compatibility.

While working on #644 I realized that when that is done, CVCs that are based on centers of mass could easily be used any positions, including individual atoms. This would much simplify the data structures of the `alpha` and `dihedPC`.

jhenin:
About making the groups groupCA, groupN and groupO generic: I realized that for the alpha CVC, we need to let the user specify several individual ranges of residues to track different alpha-helical segments. This is not compatible with having a single generic name for the index groups.

I don't get the comment about CVCs based on COMs being able to use individual atoms - isn't that already the case?

doc/colvars-refman-main.tex
```diff
@@ -2840,15 +2844,35 @@
...
+\cvnamebasedonly{
 \item %
   \labelkey{colvar|alpha|residueRange}
   \key
     {residueRange}{%
       \texttt{alpha}}{%
       Potential $\alpha$-helical residues}{%
     ``$<$Initial residue number$>$-$<$Final residue number$>$''}{%
+    Alternately, in some back-ends like NAMD and VMD, the atoms may be defined using their PSF segment name (segname)
+    and a range of residue numbers.
```
giacomofiorin, line 2875: Seems a little odd to continue the previous text here

giacomofiorin (approved):
Documentation could be improved, but I think the people who would use a template input would far outnumber those who would write it from scratch.

I would merge it in the interest of time but allocate objects using smart pointers at some point.

doc/colvars-refman-main.tex
jhenin: Agreed, changed.

PR #668, opened 2024-03-08
Restore Lepton support in GROMACS main branch

giacomofiorin:
Adding back to GROMACS the Lepton build recipe (in improved form based on changes on the Colvars recipe). Also backporting the same Lepton recipe to the 2023 *patched* release to simplify things (this PR should be easy to cherry-pick into our `gromacs-2023` branch once merged).

I am not sure about making Lepton available in the `gromacs-2024` branch, which tracks an *unpatched* version of GROMACS. I found that adding Lepton to it would be more confusing.

Once merged, I'd like to start again a patched version of GROMACS `main` in https://github.com/Colvars/gromacs. Should it be named just `main-colvars` for consistency with the older patched releases, or would further distinction be needed?

### DIFF
--- a/gromacs/CMakeLists.txt.diff
+++ b/gromacs/CMakeLists.txt.diff
@@ -1,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 70b0369bed..cccfe2aa1b 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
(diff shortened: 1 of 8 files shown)


jhenin:
I agree with everything in principle. As a reminder, the GROMACS devs are reluctant to import Lepton because it duplicates the functionality of muparser. The suggested fix was to replace muparser with Lepton in its current use in GROMACS, to remove muparser from their tree. I don't know who can put in that effort though.

PR #670, opened 2024-03-11
Backup files consistently with GROMACS standard

### LINKED ISSUE #669: Backup files consistently with backends
`colvarproxy::backup_files()` uses suffixes consistent with NAMD (`.old`, `.BAK`) also in GROMACS. This is a bit confusing and easy to forget when cleaning up a folder from backup files.

If possible, this function should be overridden in the `GROMACS` proxy.

giacomofiorin:
This PR restores the overridden version of `backup_file()`, which existed in the patched releases but not in the MDModules version.

The scheme is simple:
1. If the filename ends with `.colvars.state`, keep a single backup copy ending with `.old` (consistent with current behavior), as this is the typical behavior for checkpoints. The extension differs from the CPT because the code that handles it is different and there will be some edge cases.
2. For all other files, call the relevant GROMACS function to handle multiple backup copies.

Fixes #669

### DIFF
--- a/gromacs/gromacs-mdmodules/applied_forces/colvars/colvarproxygromacs.cpp
+++ b/gromacs/gromacs-mdmodules/applied_forces/colvars/colvarproxygromacs.cpp
@@ -42,6 +42,8 @@
 #include "colvarproxygromacs.h"
 #include "colvarproxy_gromacs_version.h"
 
+#include "gromacs/utility/futil.h"
+
 #include <sstream>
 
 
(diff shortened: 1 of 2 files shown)


gromacs/gromacs-mdmodules/applied_forces/colvars/colvarproxygromacs.cpp
```diff
@@ -162,6 +164,25 @@ void ColvarProxyGromacs::error(std::string const& message)
 }
 
 
+int ColvarProxyGromacs::backup_file(char const* filename)
+{
+    std::string const filename_str(filename);
+    auto const        state_suffix_pos = filename_str.rfind(std::string(".colvars.state"));
+    if (state_suffix_pos != std::string::npos)
```
HubLot, line 171: The C++ syntax is sooo counter-intuitive x)

HubLot (approved):

HubLot:
Good catch. The behavior will be more consistent with GROMACS spirit

gromacs/gromacs-mdmodules/applied_forces/colvars/colvarproxygromacs.cpp
HubLot: (this is just a general comment which don't imply a requested change)

PR #672, opened 2024-03-12
Check that coordinate file exists before trying to read it

### LINKED ISSUE #667: Better error message when trying to load coordinates from missing files
When loading a file that isn't XYZ the first branch of the conditional below fails, falling back to the PDB reader:
https://github.com/Colvars/colvars/blob/8eb35e75c43d6f7e16d8e634e827639097d16368/src/colvarmodule.cpp#L2159-L2171
In GROMACS the resulting error message can be confusing.

It may be clearer to check first if the file is present and readable before discussing its format.

giacomofiorin:
Uses `std::filesystem::exists` from the C++17 standard library, which is supported in GROMACS, to print a more informative error message. VMD will most likely not support it in the near future, but at least doesn't raise a "not implemented" error.

Fixes #667.


giacomofiorin:
Closing, see comments in #667

PR #673, opened 2024-03-13
Clarify purpose of functions that only handle PDB files anyway

### LINKED ISSUE #667: Better error message when trying to load coordinates from missing files

giacomofiorin:
Coordinate loading functions are renamed to reflect that they are specific to the PDB format.

### DIFF
--- a/namd/src/colvarproxy_namd.C
+++ b/namd/src/colvarproxy_namd.C
@@ -907,11 +907,11 @@ e_pdb_field pdb_field_str2enum(std::string const &pdb_field_str)
 }
 
 
-int colvarproxy_namd::load_coords(char const *pdb_filename,
-                                  std::vector<cvm::atom_pos> &pos,
-                                  const std::vector<int> &indices,
-                                  std::string const &pdb_field_str,
-                                  double const pdb_field_value)
+int colvarproxy_namd::load_coords_pdb(char const *pdb_filename,
+                                      std::vector<cvm::atom_pos> &pos,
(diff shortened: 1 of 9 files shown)


jhenin:
A very good change! I'm unsure that it solves all the problems that were raised in #667 though: the case of a missing file, grompp vs. mdrun?

giacomofiorin:
> A very good change! I'm unsure that it solves all the problems that were raised in #667 though: the case of a missing file, grompp vs. mdrun?

It wouldn't. However, it would no longer print the generic message "loading atomic coordinates from a file is currently not implemented", but rather one specific to the PDB format.

Then if the XYZ file is missing/unreadable, we get the same error message as with any other file.

jhenin:
Then I propose merging this, but not closing #667 just yet.

giacomofiorin:
I edited the description to reflect your suggestion, but I'm not sure I understand why.

If we can't use `std::filesystem::exists()` and you propose instead to try open a file to catch a generic read error, that would not be different from the current behavior for a XYZ file in GROMACS. But for a PDB, we'd want the "not implemented" error to come out regardless of whether the file is unreadable.

PR #674, opened 2024-03-14
Stop initializing cvs after atom group init fail

jhenin:
Done only to CVs where it is critical because followed by memory allocation assuming that the atom groups were parsed successfully. This fixes segfaults upon trying to load non-existent atomFiles.


giacomofiorin:
Note for future reference: not backporting this bugfix to release branches, because it only applies to the code after #643

PR #675, opened 2024-03-14
Fix initialization of histogram output files and move it to the right place

giacomofiorin:
Gets rid of annoying messages like this:
```
colvars:   Histogram hist_t will be written to file ".hist_xi.dat"
```
while also avoiding doing this initialization inside the `update()` function


giacomofiorin:
Needs fixing, convert do draft

jhenin:
As soon as it passes regtests, I'm sold :-)

Issue #671, opened 2024-03-12
Support patch number for code version


giacomofiorin:
PR #670 is an example of a small fix that should go in a future patch release of GROMACS.

We haven't had this need with LAMMPS yet, mostly because the Colvars interface currently packaged with it predates the time when LAMMPS started issuing release updates. However, we may encounter it there as well.

There should be a patch version counter, which is always zero in the `master` branch of Colvars, but gets incremented when bugfixes are backported into a Colvars version that is packaged with a specific engine.

(Not adding a milestone here for simplicity, but this should ideally be taken care of in early April, in time for [GROMACS 2024.2](https://gromacs.bioexcel.eu/t/roadmap-for-2024-patch-releases/8388/3)).

PR #681, opened 2024-04-01
Track patch version 

giacomofiorin:
Adds a patch version (which is zero in the `main` branch) to identify versions of the Colvars code where only bugfixes are added, for inclusion in patch releases of other packages. 

Solves #671

### DIFF
--- a/src/colvarmodule.cpp
+++ b/src/colvarmodule.cpp
@@ -100,8 +100,10 @@ colvarmodule::colvarmodule(colvarproxy *proxy_in)
   version_int = proxy->get_version_from_string(COLVARS_VERSION);
 
   cvm::log(cvm::line_marker);
-  cvm::log("Initializing the collective variables module, version "+
-           version()+".\n");
+  cvm::log(
+      "Initializing the collective variables module, version " + version() +
+      (patch_version_number() ? (" (patch " + cvm::to_str(patch_version_number()) + ")") : "") +
+      ".\n");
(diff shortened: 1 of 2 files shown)


jhenin:
Looks good to me in general. How will version_int be set and used?

giacomofiorin:
> Looks good to me in general. How will version_int be set and used?

I think that the simplest approach would be setting it to the same number as the patch version of the MD engine. E.g. for the code that will go into the upcoming GROMACS 2024.2 patch release, we can set it directly to 2?

As for when/how to use it, the only meaningful use case that I see now is printing it at startup, to clarify which bugs may affect (or not) a certain run. As for tracking features (big and small) we probably have enough resolution already with the current version string, do we not?

jhenin:
I see. Wouldn't it be cleaner to have all of these gathered in colvars_version.h, instead of touching a file like colvarmodule.h?

giacomofiorin:
> I see. Wouldn't it be cleaner to have all of these gathered in colvars_version.h, instead of touching a file like colvarmodule.h?

Yup, that would be cleaner indeed. However, it would also interfere with how the functions in `devel-tools/version_functions.sh` operate to automatically set the version strings. Given that the new integer flag would never be changed in the `master` branch, it didn't seem worth the extra effort to me.

Of course, I'm totally okay if you feel strongly about this and decide to go ahead and change `version_functions.sh` to play nicely with the patch version.

PR #682, opened 2024-04-04
Implement ABMD as a built-in biasing method

jhenin:
Replaces the Tcl script. Generalizes to all supported MD engines.


doc/Makefile
giacomofiorin, line 1: The changes here are undoing ommit a37b43a3

src/colvarbias_abmd.cpp
```diff
@@ -0,0 +1,136 @@
...
+  cvm::real const sign = decreasing ? -1. : 1.;
+  cvm::real const diff = (val - ref_val) * sign;
+
+  if ( diff > 0. ) {
+    colvar_forces[0] = 0.;
+    bias_energy = 0.;
+    if ( (ref_val-stopping_val) * sign <= 0. ) ref_val = val;
+  } else {
+    colvar_forces[0] = - sign * k * diff;
+    bias_energy = 0.5 * k * diff * diff;;
```
giacomofiorin, line 74: The convention here is that `k` is always in the units of (energy)/(colvar unit)^2, regardless of the value of `width`. This makes sense since this is one-dimensional by design, but I'm adding a quick note to clarify that this is different from the `harmonic` bias.

src/colvarbias_abmd.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+  return COLVARS_OK;
+}
+
+
+std::ostream & colvarbias_abmd::write_traj_label(std::ostream &os)
+{
+  size_t const this_cv_width = (variables(0)->value()).output_width(cvm::cv_width);
+  os << " ref_"
+      << cvm::wrap_string(variables(0)->name, this_cv_width-3);
```
giacomofiorin, line 123: It should be `this_cv_width-4` to reflect the length of the string "ref_" (TODO: maybe we can formalize this into a utility function).

giacomofiorin (changes requested):
Looks okay to me overall, but there are a couple inconsistencies noted below.

PR #686, opened 2024-04-10
Fix restarting mwABF simulations

### LINKED ISSUE #685: Multiple-walker shared ABF does not restart correctly
The new scheme that tracks per-replica data does not treat local vs. global data correctly, leading to the wrong state upon restart.

jhenin:
- save replica-local ABF data to restart file
- simplify CZAR data sharing by not taking deltas (useless) and summing complete data from each replica instead

Fixes #685

### DIFF
--- a/src/colvarbias_abf.cpp
+++ b/src/colvarbias_abf.cpp
@@ -244,15 +244,15 @@ int colvarbias_abf::init(std::string const &conf)
   // This used to be only if "shared" was defined,
   // but now we allow calling share externally (e.g. from Tcl).
   if (b_CZAR_estimator) {
-    last_z_samples.reset(new colvar_grid_count(colvars));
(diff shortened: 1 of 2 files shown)


giacomofiorin (approved):
Thanks for your patience. It took me a while to remember that the data from other replicas is not persistent, but gets overwritten onto the same location as you loop over the replicas. Looks good to me overall.

Nitpick: perhaps `history_last_step` could be initialized to -1 already in the header, before the call to `init()`? (I assume that the latter is needed to reset the bias in its initial state when it's redefined?)

jhenin:
About `history_last_step`: I can move the initialization to the class declaration, but it would look strange as it would be the only variable initialized there. I'm in favor of doing that systematically, but that would count as refactoring and would perhaps deserve a PR of its own. Or was there a specific reason to not have it in `init()`? I think it would be functionally equivalent because `init()` is called only once for each instance.

Issue #683, opened 2024-04-09
cv save changes prefix for future output


jhenin:
From the name and the documentation, I expect "cv save" to save files to the given prefix when called, with no further side effects. In practice, it changes the file names of all future output.

The documentation could be updated, but I'd rather change the behavior to what is expected, maybe providing a separate method of changing the output prefix persistently.

giacomofiorin:
That sounds good to me. However, without extensive refactoring that in practice means that biases will need to set `output_prefix` to the argument of `cv save`, then set it back to the "persistent" value.

Currently, how biases set `output_prefix` is highly non-uniform: ABF does this in `update()`, metadynamics in `setup_output()` and all other biases in the base class `init()` function.  Before making any changes, these should be standardized, otherwise the likelihood of new bugs would be significant...

jhenin:
If `cv save` only writes a state file, do biases have to be made aware of the new, temporary prefix in this case?

I agree with the idea of making that behavior more uniform anyway, ideally by migrating it to the base class.

giacomofiorin:
> If `cv save` only writes a state file, do biases have to be made aware of the new, temporary prefix in this case?

I believe not, save for possible edge cases. However, the documented behavior of `cv save` is to also save other output files, which do need a prefix. So what you are describing should be a separate command?
 
> I agree with the idea of making that behavior more uniform anyway, ideally by migrating it to the base class.

Sounds good!

jhenin:
I can't imagine a real usage case for this. The best interface IMO would be:
- `cv save <prefix>` saves only the global state file
- `cv bias <name>save <prefix>` (already exists)
- `cv bias <name> writeoutput <prefix>` to write the output files of the given bias only

These can be combined easily to get the current functionality, but  they are more fine-grained and easier to document.
I don't see a use case for persistently changing the output prefix, but that could be a separate command if there is a specific need.

giacomofiorin:
> I don't see a use case for persistently changing the output prefix, but that could be a separate command if there is a specific need.

NAMD changes the output prefix permanently through the `outputName` Tcl command...

jhenin:
Then I'll add a command to change it as well.

jhenin:
`cvm_output_prefix` in `colvarmodule` looks like a remnant of a previous design, and conflicts with `output_prefix_str` in `colvarproxy_io`. Ok to remove it?

PR #687, opened 2024-04-12
Create abf::setup_output() for consistency

jhenin:
Partial response to #683


jhenin:
@giacomofiorin Changes I have in progress are pinpointing a consistency issue: in NAMD, every time a simulation "run" is started, the output prefixes are updated from the NAMD parameters, which enables users to change them using NAMD scripts. However, this overrides any changes using the "cv" API. A possible way to deal with this is what I started doing: keeping track of any use of the cv API and then overriding changes in NAMD. Maybe a cleaner option would be the opposite: memorize the NAMD parameter values and only apply changes if they occur, not reset blindly to the NAMD values.

PR #691, opened 2024-05-13
Check that CPU platform matches Tcl precompiled builds

giacomofiorin:
Fixes a current CI failure, due to GitHub Actions switching to ARM-based VMs for their macOS runners.

Their choice is definitely in line with how the underlying hardware for Macs has evolved. However, this means that we can no longer reuse [UIUC's precompiled libraries](https://www.ks.uiuc.edu/Research/namd/libraries/). Because building Tcl and Tk for macOS is outside the scope of this repository, the library is now built without Tcl on macOS.

### DIFF
--- a/devel-tools/build_test_library.cmake
+++ b/devel-tools/build_test_library.cmake
@@ -25,31 +25,34 @@ if(NOT DEFINED COLVARS_TCL)
   set(COLVARS_TCL ON)
 endif()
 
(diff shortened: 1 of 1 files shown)


jhenin (approved):
I was confused at first because, unlike the branch name implies, this does not disable Tcl on ARM. But defaulting to the system library is working just fine.

PR #692, opened 2024-05-14
Add Colvars logo

jhenin:
Right now, just the source files. We might want to include it in the doc itself eventually.


giacomofiorin:
Since I had already updated the GitHub account profile to use our new logo, I'm of course happy including it in the repo as well :-)

Can you however, delete the PDF and PNG files, or at least just the latter? We probably will want to generate those (at different resolutions) as part of building the doc.

If you are not planning on embedding the logo the doc at this time, and only opened this PR just to track the SVG file itself, they we might as well do just that.

PR #694, opened 2024-05-28
More robust condition to decide when biases run on thread 0

giacomofiorin:
See https://github.com/Colvars/colvars/pull/638#pullrequestreview-2078011927

In the interest of time I opted for making the conditional simpler, since there is little to gain by distributing the biases over threads.

After merging, this commit should be cherry-picked into `namd-3.0` and `gromacs-2024`.


HanatoK (approved):
Looks good to me!

Issue #476, opened 2022-05-04
Allowing custom grid definition for each bias instance


giacomofiorin:
See #468 and links therein.

The following is a list of classes that use grids, and which could be optionally be allowed to define custom grids:
- [x] `colvarbias_abf` (#733 )
- [x] `colvarbias_histogram`
- [x] `colvarbias_meta` (#733)
- [x] `colvarbias_opes` (#733)
- [x] `colvarbias_ti` (#733)
- [x] `colvarbias_reweightaMD` (#733)

(check marks indicate implementation in an open PR)

giacomofiorin:
Adopt from #434 the keyword `grid` as the standard keyword for biases.

PR #710, opened 2024-08-30
Ability to define custom grid in ABF biases

jhenin:
This creates a new keyword, `gridParameters`, which mimics `histogramGrid` within a histogram bias.
In the process, unify both keywords under the new name and deprecate `histogramGrid`.

See Issue #476 

TODO:
- [x] tests
- [x] doc  
- [x] refine doc to make it an independent section


HanatoK:
It is good to see that the grid parameters can be separated from the `colvar{}` definitions. Do you think that it would be useful if the CZAR histogram could have a different set of grid parameters?

jhenin:
No, I can't think of a use case for having CZAR on a different grid.

giacomofiorin:
@jhenin  I am reviewing this, and as you know I like the approach (see #476), but I am also remembering now an unwanted consequence of encouraging the users to use custom definitions of the grid as their go-to option.

Metadynamics exhibits artifacts when the system routinely samples states that are too close to the grid boundaries, as you know. https://colvars.github.io/master/colvars-refman-namd.html#sec:colvarbias_meta_boundaries
Neither allowing the simulation to get stuck because of these artifacts or killing performance keeping a large number of hills is desirable. I remember speaking with several frustrated users of the Colvars metadynamics implementation at the time.

The current code and documentation are now encouraging users to just use the mathematical boundaries of the CVC functions (see #310), i.e. not define grid boundaries themselves when default values are available.

If we merge this PR (as I think we will very soon), we would also need to improve the code introduced in #310 and let it check that each user-provided boundary matches the default given by the CVCs. Would you be okay with waiting to contribute to the MD engine distributions until these changes are also merged?

jhenin:
This is now superseded by #733

giacomofiorin:
@jhenin You closed it a few months ago because the commits are in #733. Okay to delete this branch?

PR #715, opened 2024-09-16
Add missing rotation in orientation component

giacomofiorin:
Fixes issue identified by @HanatoK (see https://github.com/Colvars/colvars/pull/713#issuecomment-2346452299)

The rotation of applied forces from the rotated frame back to the simulated system's frame has been missing for a very long time in the `colvar::rotation` class. Specifically, the class predates the introduction in Colvars of the rotated frame feature, and both predate the first public release of the code in the NAMD distribution.

Great catch!!


HanatoK:
Could you take a look at the fix in #713? That should include the inverse of rotation in both the `orientation` and `cartesian`, and also add the missing force on the fitting group.

giacomofiorin:
> Could you take a look at the fix in #713? That should include the inverse of rotation in both the `orientation` and `cartesian`, and also add the missing force on the fitting group.

I would rather prioritize fixing this bugfix first, and get the reference data for the `orientation-fitgroup` feature combination updated before dealing with #713.

Please feel free to change anything in this PR, or even close and open a different one, as long as the CI jobs can be run.

src/colvarcomp_rotations.cpp
```diff
@@ -133,14 +133,25 @@ void colvar::orientation::calc_gradients()
 void colvar::orientation::apply_force(colvarvalue const &force)
 {
   cvm::quaternion const &FQ = force.quaternion_value;
+  auto const rot_inv = atoms->rot.inverse().matrix();
```
HanatoK, line 136: I think it is better to move this line into the `if (atoms->is_enabled(f_ag_rotate)) {}` case.

HanatoK (approved):

Issue #716, opened 2024-09-16
Compilation error in GROMACS main test code


giacomofiorin:
See e.g. https://github.com/Colvars/colvars/actions/runs/10703315046/job/29673605133

HubLot:
weird... Looks like it comes from [here](https://gitlab.com/gromacs/gromacs/-/merge_requests/4502).
But, it compiles correctly on their Gitlab CI

PR #718, opened 2024-09-17
Import fixes and changes from GROMACS repo

jhenin:
Limited to the gromacs/applied_forces directory, not the library

Should help with #716


jhenin:
Fixes #716

PR #720, opened 2024-09-24
Integrate multiple-replicas MPI communicator into library, support GROMACS multisim

giacomofiorin:
Moves member data and functions from the LAMMPS interface to the library, with the purpose of being reused in GROMACS.

The preprocessor macro `COLVARS_MPI` controls whether `MPI_Comm` is defined: if it isn't it is defined a `void *` via typedef by default. Both LAMMPS and GROMACS set that macro in different ways:
- LAMMPS always sets it, because it has a MPI stubs "library" to link against
- GROMACS defines it only when not using thread-MPI
- NAMD ignores it altogether, and uses derived-class functions


HanatoK:
Should I rebase https://github.com/Colvars/colvars/pull/709 to use `check_replicas_enabled()` on this PR?

giacomofiorin:
> Should I rebase #709 to use `check_replicas_enabled()` on this PR?

No, #709 is ready for review. Better to rebase this onto `master` later.

giacomofiorin:
@jhenin @HubLot I synchronized the contents of this branch with the GROMACS MR [4607](https://gitlab.com/gromacs/gromacs/-/merge_requests/4607).

Should we leave this open until the other one is merged?

HubLot:
Yes, we can leave it open in cases of small changes.

HubLot (approved):

giacomofiorin:
NOTE: compared to the original draft, the corresponding MR in the GROMACS repo has been merged. This PR now takes that into account and also removes support for GROMACS versions prior to the addition of the Colvars MDModule in GROMACS 2024.

@jhenin The test I implemented is very simple, can you perhaps change its input so that it runs a shared ABF simulation with no biasing forces and you compare the resulting PMF to the sum of four independent runs.

Note also that the GROMACS main contains a slightly different version of the replica reporting message (0-based index instead of 1-based number). All tests should be run for the latest Colvars `master`, but just making sure.

giacomofiorin:
Also @jhenin I removed the GROMACS <= 2023 files altogether. If you wish to preserve them in a sub-folder, please replace the commit that deletes them with one that moves them during rebase

giacomofiorin:
(Merging according to DMs)

PR #721, opened 2024-09-28
Find PyTorch from CMake for standalone library builds

giacomofiorin:
This PR adds support for linking against PyTorch in the CMake recipes. 

As an accompanying change (in this repo, but not technically part of this PR), I also updated the "CentOS9-devel" container to host a version of PyTorch (CPU platform) installed from PyPI.

With this change, it should be possible to test PyTorch support using only the standalone library and the tests in the folder `tests/input_files`.  I also propose removing all PyTorch-related patches to the NAMD, LAMMPS and GROMACS build systems to avoid conflicts with their own repos.

The target branch of this PR is `torchann`, since that's where the changes are most useful right now. Please review to make sure that this is an acceptable version of PyTorch for testing, and propose changes to the code or retargeting against `master` if applicable.


cmake/CMakeLists.txt
```diff
@@ -92,6 +92,20 @@ if(COLVARS_OPENMP)
...
+    OUTPUT_VARIABLE SITE_PACKAGES_PATH
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+  find_package(Torch PATHS file(JOIN ${SITE_PACKAGES_PATH} "torch/share/cmake/Torch"))
+  if(Torch_FOUND)
+    message(STATUS "Will link against Torch: ${TORCH_LIBRARY}")
+    target_include_directories(colvars PRIVATE ${TORCH_INCLUDE_PATH})
+    target_link_libraries(colvars PRIVATE ${TORCH_LIBRARY})
+  endif()
+endif()
```
HubLot, line 108: How those lines interact with the other search for Torch below in the CmakeLists.txt? (see lines 140-152)

zwpku: I tried to build covlars as a library: 

- Torch_FOUND is true, libcolvars.a can be created, but torchann class is not compiled, because -DTORCH is not set. 

-  If I change the value of COLVARS_TORCH (line 138) to ON, then torchann class will be compiled and libcolvars.a can be created. However,  there are several errors "undefined reference to..." in the step "Linking CXX executable run_colvars_test".

giacomofiorin: I integrated the two, but I am also finding linkage errors that shouldn't be there (undefined symbol when running `ld` while also including `libcolvars.so`, which does define that symbol).

Unless somebody has other ideas, I would keep an eye out for what the MD packages will do in their own test suites. We will have to use their build systems for distribution anyway...

Setting this back as draft.

HubLot: looks like it's not very straightforward :/

According to #570, it looks like only the libTorch (C++) is needed. Does the Python version is mandatory? Otherwise it could simplify the cmake recipe

cmake/CMakeLists.txt
```diff
@@ -118,29 +117,37 @@ if(COLVARS_TCL)
...
-    message(FATAL_ERROR "Torch package not found.")
+  message(STATUS "Trying to find the Torch library from its Python package")
+  find_package(Python3)
+  if(Python3_FOUND)
+    execute_process(
+      COMMAND ${Python3_EXECUTABLE} -c "import sys; print(next(p for p in sys.path if 'site-packages' in p))"
+      OUTPUT_VARIABLE SITE_PACKAGES_PATH
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    find_package(Torch PATHS file(JOIN ${SITE_PACKAGES_PATH} "torch/share/cmake/Torch"))
```
HanatoK, line 134: A simple `find_package(Torch)` works for me. This doesn't. I usually add `-DCMAKE_PREFIX_PATH=/home/hanatok/HDD/software/libtorch_2.2.2` (I downloaded and unzipped the libtorch library to the path) to compile a library with libtorch.

cmake/CMakeLists.txt
```diff
@@ -118,29 +117,37 @@ if(COLVARS_TCL)
...
-    set_property(TARGET colvars PROPERTY CXX_STANDARD 17)
-    target_compile_definitions(colvars PRIVATE -DTORCH)
-    # target_compile_options(colvars PRIVATE ${TORCH_CXX_FLAGS})
-    target_compile_options(colvars PRIVATE ${CMAKE_CXX_FLAGS} ${TORCH_CXX_FLAGS})
-    target_include_directories(colvars PRIVATE ${TORCH_INCLUDE_DIRS})
-    target_link_libraries(colvars PRIVATE "${TORCH_LIBRARIES}")
-  else()
-    message(FATAL_ERROR "Torch package not found.")
+  message(STATUS "Trying to find the Torch library from its Python package")
+  find_package(Python3)
```
HanatoK, line 127: Why do we need python3?

cmake/CMakeLists.txt
giacomofiorin: > A simple `find_package(Torch)` works for me. This doesn't. 

There was probably some trickery involved in packaging libTorch for PyPI that I can't figure out.

> I usually add `-DCMAKE_PREFIX_PATH=/home/hanatok/HDD/software/libtorch_2.2.2` (I downloaded and unzipped the libtorch library to the path) to compile a library with libtorch.

The "pip" option was the recommended one at: https://pytorch.org/get-started/locally/

Does the "LibTorch" package include Python modules as well? Better if it did, because in the future I would like to also use your awesome scripts to test gradients. If it doesn't we can treat it in the same way as Charm++ I guess and put it under /opt as an optional dependency.

In any case, I would like a standard scriptable procedure to bundle libTorch in the containers. Manually bundling is the sort of thing that "works on my computer" but not great for cloud CI jobs :frowning_face:

HanatoK: > The "pip" option was the recommended one at: https://pytorch.org/get-started/locally/

No. When I select "C++/Java" in Language, the "Package" selection turns to "LibTorch" automatically, so I don't think it recommends "pip". If you select "pip" then you download the python package instead of the C++ package. What we need is the C++ package "libtorch".

HanatoK: > In any case, I would like a standard scriptable procedure to bundle libTorch in the containers. Manually bundling is the sort of thing that "works on my computer" but not great for cloud CI jobs ☹️

If you select "LibTorch", "C++/Java" and then "CPU", the website will give you the download link of libtorch. Is it possible to just use the link with wget or curl to download the libtorch for CI?

giacomofiorin: Right, so by this you are confirming that "libTorch" only has the underlying C++ library with no Python bindings. (Or are you using that same library in scripts?)

HanatoK: Yes. I believe #570 only needs the libtorch and does not need the python bindings.

giacomofiorin: I updated the container to use the C++ only libTorch. However, using just `-DCMAKE_PREFIX_PATH=/opt/libtorch` *did not* work. I still had to manually feed the prefix to `find_package()`, so I put the detection of it in the CMake script used for tests.

This PR should be working now, at least minus the stuff that was still broken in `master` on the last merge of `torchann`.

HubLot: You can also fill `Torch_DIR` cmake variable with the path to the cmake configfile directory (`-DTorch_DIR=/opt/libtorch/share/cmake/Torch`) instead of `CMAKE_PREFIX_PATH`

cmake/CMakeLists.txt
```diff
@@ -118,29 +113,29 @@ if(COLVARS_TCL)
...
-  target_link_libraries(colvars PRIVATE ${TCL_LIBRARY})
+  target_link_libraries(colvars PUBLIC ${TCL_LIBRARY})
 endif()
 
 option(COLVARS_TORCH "Link against the Torch library" OFF)
 
 if(COLVARS_TORCH)
-  find_package(Torch)
-  if (Torch_FOUND)
-    set_property(TARGET colvars PROPERTY CXX_STANDARD 17)
```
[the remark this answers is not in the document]
giacomofiorin: Ok, but are you sure that that wouldn't mean a downgrade when we are testing C++20 or later?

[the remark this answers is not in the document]
giacomofiorin: Meaning, if you know how to set the _minimum_ required version rather than hard-coding 17 that would be perfect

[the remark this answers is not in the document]
HanatoK: Is it possible to do a check to see if the CMAKE_CXX_STANDARD is less than 17, and increase it to 17 if so?
```cmake
  if (CMAKE_CXX_STANDARD LESS 17)
    message(STATUS "Increase the CMAKE_CXX_STANDARD to 17")
    set_property(TARGET colvars PROPERTY CXX_STANDARD 17)
  else()
    message(STATUS "Keep the CMAKE_CXX_STANDARD as ${CMAKE_CXX_STANDARD}")
  endif()
```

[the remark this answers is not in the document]
giacomofiorin: Yeah, that works, can you commit it?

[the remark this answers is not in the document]
zwpku: It looks good. In my case, cmake can find libtorch with either `-DCMAKE_PREFIX_PATH` or `-DTorch_DIR`.

zwpku (commented):
It looks good. In my case, cmake can find libtorch with either `-DCMAKE_PREFIX_PATH` or `-DTorch_DIR`.

PR #724, opened 2024-10-10
Fix gradients and metric functions of distanceDir

### LINKED ISSUE #714: distance_dir::apply_force seems incorrect
While testing https://github.com/Colvars/colvars/pull/713 I find that the forces in `distance_dir::apply_force` seems incorrect (I compare it to pytorch). The numerically correct results seems to be
```c++
if (!group1->noforce)
    group1->apply_force(-1.0 * force_tang / (0.5 * dist_v.norm()));

  if (!group2->noforce)
    group2->apply_force(       force_tang / (0.5 * dist_v.norm()));
```
I know the forces are calculated by some sorts of projection. I don't know why the modified code above gives me the correct numerical results, but the missing of `dist_v.norm()` in the function looks a bit strange. I need some time to figure it out.

giacomofiorin:
Fixes #714


HanatoK:
If this one is merged before #713, I will need to update the test file introduced in the PR.

giacomofiorin:
> If this one is merged before #713, I will need to update the test file introduced in #713.

Entirely up to you. I already looked at the code in #713 and it looks pretty clear, so I think we can merge either way. Let me write a review there.

src/colvarvalue.cpp
```diff
@@ -600,25 +668,30 @@ colvarvalue colvarvalue::dist2_grad(colvarvalue const &x2) const
...
-      cvm::rvector const &v1 = this->rvector_value;
-      cvm::rvector const &v2 = x2.rvector_value;
-      cvm::real const cos_t = v1 * v2;
-      return colvarvalue(2.0 * (cos_t * v1 - v2), colvarvalue::type_unit3vectorderiv);
-    }
+  case colvarvalue::type_unit3vector: {
+    cvm::rvector const &v1 = this->rvector_value;
+    cvm::rvector const &v2 = x2.rvector_value;
+    cvm::real const cos_t = v1 * v2;
+    return colvarvalue(2.0 * std::acos(cos_t) * -1.0 / cvm::sqrt(1.0 - cos_t * cos_t) * v2,
```
HanatoK, line 675: Actually I am still not quite sure about the meaning of `dist2_grad`. For a call like `x1.dist2_grad(x2)`, does it calculating the partial derivative of distance with respect to $\mathbf{x}_1$ or $\mathbf{x}_2$?

HanatoK (approved):
Looks good to me!

src/colvarvalue.cpp
giacomofiorin: The former (the scalar version is 2 * (x1 - x2), which for x2 == 0 has the same sign as x1).

Admittedly this could be clarified by adding `colvarvalue::dist2_lgrad()` and `colvarvalue::dist2_rgrad()`. Or rather removing the latter and naming `dist2_grad` to `gradient_distance_from()`? Either way not for this PR.

src/colvarcomp_distances.cpp
```diff
@@ -384,32 +384,30 @@ void colvar::distance_dir::apply_force(colvarvalue const &force)
...
+  }
 }
 
 
-cvm::real colvar::distance_dir::dist2(colvarvalue const &x1,
-                                      colvarvalue const &x2) const
+cvm::real colvar::distance_dir::dist2(colvarvalue const &x1, colvarvalue const &x2) const
 {
-  return (x1.rvector_value - x2.rvector_value).norm2();
+  return x1.dist2(x2);
```
HanatoK, line 398: So now the square of distance of two `distanceDir` is calculated as the angle between them. Should we clarify that in the documentation?

After merging this PR and #713 , I may still need to verify if `polarPhi` and `distanceDir` with a custom function could yield the same trajectories.

src/colvarvalue.cpp
HanatoK: I don't think a code change is necessary. You could add some comments before `dist2_grad` to help people figure out whether it is $\partial d^2 / \partial x_2$ or $\partial d^2 / \partial x_1$.

src/colvarcomp_distances.cpp
giacomofiorin: > So now the square of distance of two `distanceDir` is calculated as the angle between them. Should we clarify that in the documentation?

Yes. I meant to do that, but forgot. I would say, though, that defining the distance as such is the same as taking the standard Euclidean distance (L^2 norm) but making it comply with the unit-vector constraint.

> After merging this PR and #713 , I may still need to verify if `polarPhi` and `distanceDir` with a custom function could yield the same trajectories.

Yeah, that makes sense.

giacomofiorin:
@HanatoK I merged this for now before #713, where Jérôme might need to take a look. Will port into the `namd-3.0` branch as well.

src/colvarvalue.cpp
giacomofiorin: A comment was already near the declaration (`colvarvalue.h`), I just added another one in the definition (`colvarvalue.cpp`) as a commit straight to `master.

PR #728, opened 2024-10-12
Fix undefined behavior when getting the current working directory from std::filesystem

### LINKED ISSUE #726: Inconsistent GETCWD in src/colvarbias_meta.cpp can cause underfined behavior
In src/colvarbias_meta.cpp, there's `GETCWD(BUF, SIZE)` macro used to wrap `getcwd` function to put current working dir into `BUF`. However, when `std::filesystem` is available, a different implementation is used, which does not change `BUF` but instead returns a temporary C-string: https://github.com/Colvars/colvars/blob/0e0ce447630e95f2cd0322b1699e7f60b526ef9f/src/colvarbias_meta.cpp#L25C1-L30C7

This is not ok with [how `GETCWD` is later used](https://github.com/Colvars/colvars/blob/master/src/colvarbias_meta.cpp#L1743-L1754):

```cpp
    char *pwd = new char[3001];
    if (GETCWD(pwd, 3000) == nullptr) {
      if (pwd != nullptr) { //
        delete[] pwd;
      }
      return cvm::error("Error: cannot get the path of the current working directory.\n",
                        COLVARS_BUG_ERROR);
    }

    replica_list_file =
      (std::string(pwd)+std::string(PATHSEP)+
       this->name+"."+replica_id+".files.txt");
```

The conditional will never be false, but `pwd` will not be initialized; thus so `std::string(pwd)` later is UB.

giacomofiorin:
Adds two functions `colvarproxy_io::get_current_work_dir()` and `colvarproxy_io::join_paths()` to replace the code currently used in `colvarbias_meta.cpp`. It's possible that these may be removed again sometime in the future.

Fixes #726

### DIFF
--- a/src/colvarproxy_io.h
+++ b/src/colvarproxy_io.h
@@ -38,6 +38,12 @@ public:
   // Returns error code
   virtual int set_frame(long int);
 
+  /// Get the current working directory of this process
+  virtual std::string get_current_work_dir() const;
+
+  /// Join two paths using the operating system's path separation
+  virtual std::string join_paths(std::string const &path1, std::string const &path2) const;
+
   /// \brief Rename the given file, before overwriting it
   virtual int backup_file(char const *filename);
 
--- a/src/colvarbias_meta.cpp
+++ b/src/colvarbias_meta.cpp
@@ -11,24 +11,6 @@
 #include <iomanip>
 #include <algorithm>
 
-// Define function to get the absolute path of a replica file
-#if defined(_WIN32) && !defined(__CYGWIN__)
-#include <direct.h>
-#define GETCWD(BUF, SIZE) ::_getcwd(BUF, SIZE)
-#define PATHSEP "\\"
-#else
-#include <unistd.h>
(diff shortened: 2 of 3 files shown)


HubLot (approved):

src/colvarproxy_io.h
```diff
@@ -38,6 +38,12 @@ class colvarproxy_io {
   // Returns error code
   virtual int set_frame(long int);
 
+  /// Get the current working directory of this process
+  virtual std::string get_current_work_dir() const;
```
al42and, line 42: Does it need to be virtual? I think it can even be `static`.

al42and (commented):
A few nitpicks

giacomofiorin:
Thanks @HubLot and @al42and!

I removed the `virtual` for the new functions, since it is indeed unlikely that we need to add a special case before we are able to assume `std::filesystem` everywhere. (We do, however, need to do file I/O differently for NAMD, which has its own class to replace `std::ofstream`).

As for `static`, they could definitely be but that class doesn't have any other statics, so I preferred to keep it consistent with other member functions.

PR #735, opened 2024-10-21
Enable and test mwABF in GROMACS

jhenin:
Now that Colvars has access to the multisim communicator in GROMACS, mwABF is possible in principle, however it had not been tested with the current code base.

I disabled an init-time check for mutliple replicas because the info becomes available after the proxy construction in GMX. The lack of mutliple-replica setup will then be caught when the first communication attempt occurs.

TODO:

- [x] extend doc, currently written for NAMD only
- [x] Check LAMMPS support


giacomofiorin:
> I disabled an init-time check for mutliple replicas because the info becomes available after the proxy construction in GMX. The lack of mutliple-replica setup will then be caught when the first communication attempt occurs.

In #720 I opted for updating only the IForceProvider-derived class, to not touch the preprocessing part (where there is no notion of multiple replicas). Given time constraints, I preferred not to modify too much the constructor of the GROMACS proxy. But I realize now that @HubLot already isolated the module's initialization within an a `if (doParsing)` block.

Maybe we can make that block into a function and call it after setting the multisim communicator?  @jhenin @HubLot What do you think?

jhenin:
@giacomofiorin The key weakness here is the lack of deferred initialization: the proxy constructor parses the input, so it needs pretty much all the information.
I agree with your suggestion in principle, but that doesn't have a natural equivalent in the other back-ends, or rather it has to be implemented differently depending on the back-end.

giacomofiorin:
> @giacomofiorin The key weakness here is the lack of deferred initialization: the proxy constructor parses the input, so it needs pretty much all the information. I agree with your suggestion in principle, but that doesn't have a natural equivalent in the other back-ends, or rather it has to be implemented differently depending on the back-end.

Actually, maybe doing things consistently with LAMMPS may be better indeed:
1. create the proxy object
2. call `proxy->add_config()` when config is available
3. finish initializing the proxy object
4. call `proxy->parse_module_conflg()`

jhenin:
Basic LAMMPS support works, but so far I have not found a way to 

1. set different Colvars output prefixes per replica
2. set different starting configs / velocities per replica

giacomofiorin (approved):
Looks good to me. 

If the lack of deferred initialization is still a big issue for GROMACS, can you try to convince the folks in Stockholm to allow a small change to the MDModule? This would not count as a new feature IMO.

I can take care of the LAMMPS issue separately.

PR #749, opened 2024-11-15
Add references to new Colvars paper

jhenin:
This PR is here to add citations to the new minireview paper as necessary.

Now only two features do not have a specified reference.


jhenin:
Works for me in this current state.

giacomofiorin:
Should we list Lepton and Tcl support under the 2024 reference?

jhenin:
> Should we list Lepton and Tcl support under the 2024 reference?

In the absence of a dedicated reference, I would say so.

PR #753, opened 2024-11-29
Save restraint energy derivative in trajectory

jhenin:
to enable easy post-processing for TI, including equilibrium detection and variance analysis.


jhenin:
This is made redundant by https://github.com/BranniganLab/safep/pull/56

PR #759, opened 2024-12-16
min_image fix; Saves long runs from crashes;

PolyachenkoYA:
This addresses [this](https://matsci.org/t/fix-rigid-atoms-have-moved-too-far-apart/53926/14) recent issue solution


jhenin:
Merging directly since the change has been made downstream.

PR #776, opened 2025-03-07
Avoid error in acos and asin with fast-math

jhenin:
by returning hardcoded values on boundaries -1 and 1

Compared with previous fix #768 this is more radical:
- triggers when fast-math is enabled regardless of architecture (following remarks by @HanatoK https://github.com/Colvars/colvars/pull/768#issuecomment-2672473081)
- does not call the library functions when the argument is "equal" to -1 or 1, but returns hard-coded values.

### DIFF
--- a/src/colvarmodule.h
+++ b/src/colvarmodule.h
@@ -18,7 +18,7 @@
 #define COLVARS_DEBUG false
 #endif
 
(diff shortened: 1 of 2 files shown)


HanatoK:
Looks good.

PR #795, opened 2025-04-25
Account for NAMD's GlobalMasterFrequency

### LINKED ISSUE #793: extended Lagrangian is not compatible with NAMD's globalMaster MTS
Hi all,

I recently found that extended-Lagrangian algorithms (e.g., eABF) are not compatible with NAMD's new globalMaster MTS. When running an eABF simulation with:
```
globalMasterFrequency 2
globalMasterScaleByFrequency off
globalMasterStaleForces on
```
NAMD outputs:
```
colvars:Detected discrete jump in colvar value from -72.463 to -67.3934.
colvars :Reinitializing extended coordinate to colvar value.
Synchronizing (emptying the buffer of)trajectory file "output/eabf.colvars.traj"
colvars :The restart output state file will be "output/eabf.restart.colvars.state".colvars :colvars:Initializing extended coordinate to colvar value.
colvars:Detected discrete jump in colvar value from -49.7613 to -52.3231.
colvars:Reinitializing extended coordinate to colvar value.
colvars :Initializing extended coordinate to colvar value.
colvars :Reinitializing extended coordinate to last value.Initializing extended coordinate to colvar value.
```
The values in the `.czar.pmf` file remain zero. In contrast, plain ABF works well with this new feature (all input files attached:

[MTS.tar.gz](https://github.com/user-attachments/files/19861342/MTS.tar.gz)

).

I suspect this issue arises because NAMD stores the last-step biasing force and performs an additional integration step for the real system, while not for the extended degrees of freedom. Consequently, Colvars detects discrepancies and attempts corrections every two steps. I recommend modifying Colvars to recognize the `globalMasterStaleForces on` setting, store the previous biasing force, and integrate the extended DOF independently of NAMD's communication.

Best,
Haohao

jhenin:
Should fix #793 

More issues may exist with time-dependent biases that assume that they are called every timestep. A review of all time-dependent biases is needed.


jhenin:
There are 3 options for the extended Lagrangian:
- enforce compatible mts by complaining on startup if the settings are inconsistent
- enforce compatible mts by automatically setting the global factor as each colvar's factor
- do not use compatible factors but repeat the integration step somehow.

For now I've implemented the first option as the most robust and transparent, so users have to be aware of the stakes.

The same mechanism may in principle be used for all colvars and biases.

jhenin:
My favored solution for all biased is to generalize what I have for colvars: that they all require to have their own timestep factor matching that of the proxy. This is the only way to fall back to an already supported situation, so it's the safest.
Later, we can implement automatic matching of those parameters if we want to provide the convenience, but that can wait.

giacomofiorin:
@jhenin If you look at ABF specifically I can look into the other time-dependent biases.

giacomofiorin:
Note: I'll delete the `namd-globalmasterfrequency` branch, which is now superseded by this PR

giacomofiorin:
@jhenin Currently, the default value of `timeStepFactor` is 1. How about we use instead the value from the proxy?

giacomofiorin:
Also, looking more closely I don't think it's just the time-dependent biases, but also everything that is done at a predetermined frequency, like trajectory I/O or restarting.

giacomofiorin:
@jhenin Looking more at this, I keep finding places where "skipping a few steps" could potentially be a problem, and adding error checks for those seems like a lot of work.

Ideally, I think that NAMD should call Colvars *at every step*, regardless of whether it's sending updated coordinates. However, I don't see an easy way to add such a call in NAMD outside of the existing GlobalMaster/CUDAGlobalMaster infrastructure (@HanatoK please correct me if relevant).

If that's the case, how about we ensure that Colvars does its internal bookkeeping by just calling it *repeatedly* until it catches up to the current NAMD step?

jhenin:
I would certainly mention it in the doc, and add checks for the most important parameters (trajectory output and restart frequency). Anything else is likely to be commensurate to those, and also less frequently used.

PR #796, opened 2025-04-25
Keep modern gromacs files

jhenin:
Files under src/gromacs/applied_forces/colvars are now de facto maintained in the gromacs repo. Updating them from the Colvars repo tends to break the interface.

As far as I can see, the directory could be scrapped from the repo entirely, and the update script could deal with the library only.

### DIFF
--- a/update-colvars-code.sh
+++ b/update-colvars-code.sh
@@ -534,24 +534,24 @@ then
   if [ -d ${target_folder} ]
   then
     echo "${target} source tree seems to have already been patched."
-    echo "Updating to the current Colvars sources."
+    echo "Preserving files from the Gromacs tree."
   else
     mkdir ${target_folder}
     mkdir -p ${target_folder}/tests/refdata
+    condcopy gromacs/src/colvarproxy_gromacs_version.h "${target_folder}/colvarproxy_gromacs_version.h"
+    for src in ${source}/gromacs/src/applied_forces/colvars/*.* ; do
+      tgt=$(basename ${src})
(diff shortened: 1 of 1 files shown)


jhenin:
Incidentally, this fixes the Gromacs regression test.

giacomofiorin (approved):

jhenin:
Answering this: https://github.com/Colvars/colvars/pull/788#issuecomment-2821490924
> Is it possible to use a specific version of GROMACS for the CI testing in Colvars, just like NAMD?

Because the Gromacs code is generally less stable and the interface was recent, we decided to keep the main branch to detect early if Gromacs breaks anything. Since we have basic Colvars unit tests and the interface is slightly more mature, we could consider switching to a fixed release, but I think this PR is a good compromise: if the Gromacs devs take into account their changes in their own repo, that should be enough for the code to compile in our CI and undergo the more in-depth regression tests.

giacomofiorin:
> Files under src/gromacs/applied_forces/colvars are now de facto maintained in the gromacs repo. Updating them from the Colvars repo tends to break the interface.
> 
> As far as I can see, the directory could be scrapped from the repo entirely, and the update script could deal with the library only.

I agree with the change to the installer script.

For tracking or not the GROMACS files, I think it depends on the specific case. As of right now, we are not planning immediate changes to the GROMACS interface but we may those soon, and that would entail tracking those files somehow in the Colvars repo. What do you think about using diff files?

Meanwhile I'll merge this.

PR #797, opened 2025-04-29
Add torchANN documentation

### LINKED ISSUE #791: Missing torchANN documentation
While dealing with the numerous merge conflicts connected #570, there was a lot of noise and we somehow lost the relevant documentation for the `torchANN` feature :-(

@zwpku Do you still have the original files? I can make something up based on the example input, but it would be good to have the documentation that reflects your intended use.

giacomofiorin:
Add some baseline content for the `torchANN` feature based on the source code.

Fixes #791

### DIFF
--- a/doc/colvars-refman-main.tex
+++ b/doc/colvars-refman-main.tex
@@ -3799,6 +3799,49 @@ An example of configuration using \texttt{NeuralNetwork} is shown below:
 \end{cvexampleinput}
 
 
+\cvsubsec{Torch-based collective variables}{sec:cvc_torchann}
+
+The \texttt{torchANN} keyword defines a component based on an artificial neural network (ANN) model previously trained using PyTorch. Similarly to the \texttt{linearCombination} and \texttt{neuralNetwork} components, the ANN model is expressed as a function of collective variable components. The model is loaded from a PyTorch-formatted file, defined by the keyword \texttt{modelFile}\cvgromacsonly{; the contents of this file are embedded in the TPR}.
+
+This component requires that \MDENGINE{} is linked against the libTorch library; see \ref{sec:compilation_notes} for more details.
+
+\begin{cvcoptions}
+
+\item %
+  \labelkey{colvar|torchANN|modelFile}
+  \key
+    {modelFile}{%
+    \texttt{torchANN}}{%
+    Name of the file containing the model (PyTorch format)}{%
+    filename}{%
+    This file contains the PyTorch model, as saved for example by the \texttt{torch.save()} function.
+  }
+
+\item %
+  \labelkey{colvar|torchANN|m_output_index}
+  \keydef
+    {m\_output\_index}{%
+    \texttt{torchANN}}{%
+    Index of the component to use}{%
+    0-based integer}{%
+    0}{%
+    When the output of the ANN has multiple components, this option selects the component to be used as the collective variable.
+  }
+
+\item %
+  \labelkey{colvar|torchANN|doubleInputTensor}
+  \keydef
+    {doubleInputTensor}{%
+    \texttt{torchANN}}{%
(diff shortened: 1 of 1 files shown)


giacomofiorin:
@HanatoK @jhenin I am inclined to merge this in right away, since a placeholder doc section is better than none.

What do you think about the keyword `m_output_index`? When comparing with `linearCombination` and `neuralNetwork` I realized that the same field is being used with a different keyword (they use `output_component`). Should we add `output_component` to `torchANN` as well as an equivalent option?

HanatoK:
> @HanatoK @jhenin I am inclined to merge this in right away, since a placeholder doc section is better than none.
> 
> What do you think about the keyword `m_output_index`? When comparing with `linearCombination` and `neuralNetwork` I realized that the same field is being used with a different keyword (they use `output_component`). Should we add `output_component` to `torchANN` as well as an equivalent option?

I am personally OK with the changes, but I am not the main user of `torchANN` so it is better to ask if @zwpku agrees with the changes.

zwpku:
@giacomofiorin  

Sorry for being inactive. 

I agree the changes. Maybe it is even better to use the same keyword in torchann, i.e. replacing  
```
  get_keyval(conf, "m_output_index", m_output_index, 0);
```
by 
```
  get_keyval(conf, "output_component", m_output_index, 0);
```

Besides, in order to load models in c/c++ code, models have to be saved using [torch.jit.save()](https://pytorch.org/docs/stable/generated/torch.jit.save.html). 
So, in the following sentence in colvars-refman-main.tex, 

https://github.com/Colvars/colvars/blob/80d68ba7fe3c3b81443db040a181cd50ce65b8df/doc/colvars-refman-main.tex#L3817

it is more accurate to say 

>  as saved for example by the \texttt{torch.jit.save()} function. 

Do we want to add slightly more details (regarding the usage of this class) in the document, such as a very simple example? I can do this.

giacomofiorin:
> I agree the changes. Maybe it is even better to use the same keyword in torchann, i.e. replacing
> 
> ```
>   get_keyval(conf, "m_output_index", m_output_index, 0);
> ```
> 
> by
> 
> ```
>   get_keyval(conf, "output_component", m_output_index, 0);
> ```

Thanks! Since we already have included the code in the GROMACS 2025 release, we will need to support both keywords. But we can deprecate the former immediately, since it was never documented anyway. 


> 
> Besides, in order to load models in c/c++ code, models have to be saved using [torch.jit.save()](https://pytorch.org/docs/stable/generated/torch.jit.save.html). So, in the following sentence in colvars-refman-main.tex,
> 
> https://github.com/Colvars/colvars/blob/80d68ba7fe3c3b81443db040a181cd50ce65b8df/doc/colvars-refman-main.tex#L3817
> 
> it is more accurate to say
> 
> > as saved for example by the \texttt{torch.jit.save()} function.
> 
> Do we want to add slightly more details (regarding the usage of this class) in the document, such as a very simple example? I can do this.

giacomofiorin:
> Do we want to add slightly more details (regarding the usage of this class) in the document, such as a very simple example? I can do this.

Yes, please feel free to use a PR to add any doc text or examples that you deem appropriate. In doing so, please add your name to the list of authors of the doc in the main file `doc/colvars-refman.tex`.

giacomofiorin:
> Besides, in order to load models in c/c++ code, models have to be saved using [torch.jit.save()](https://pytorch.org/docs/stable/generated/torch.jit.save.html). So, in the following sentence in colvars-refman-main.tex,
> https://github.com/Colvars/colvars/blob/80d68ba7fe3c3b81443db040a181cd50ce65b8df/doc/colvars-refman-main.tex#L3817
> 
> it is more accurate to say
> > as saved for example by the \texttt{torch.jit.save()} function.

Thanks for the correction! I included it in #798

PR #798, opened 2025-04-29
Rename m_output_index to output_component in torchANN

giacomofiorin:
Make the option more consistent with the `neuralNetwork` component.

Also fix the example of the PyTorch code in the doc


giacomofiorin:
@zwpku I'm going to merge this since you already [approved](https://github.com/Colvars/colvars/pull/797#issuecomment-2840054243) and I need to update the PR for the LAMMPS repo (you should be tagged in that I believe).

For any further changes to the doc coming from you I think it's easier to add them, since they are guaranteed not to break LAMMPS (and we host the HTML doc anyway)

PR #799, opened 2025-04-30
Weighted Poisson integration

jhenin:
This is making substantial changes to the Poisson integration code, and this is creating many conflicts with recent changes in the grid code.

Incidentally, colvargrid.cpp is way too long, largely because of the complex `integrate_potential` class. I'm thinking it would be better to split it off into its own file first in master, and then squash and merge this PR so that there is less interference with the colvar_grid classes.

What do you think @giacomofiorin ?


giacomofiorin:
> This is making substantial changes to the Poisson integration code, and this is creating many conflicts with recent changes in the grid code.
> 
> Incidentally, colvargrid.cpp is way too long, largely because of the complex `integrate_potential` class. I'm thinking it would be better to split it off into its own file first in master, and then squash and merge this PR so that there is less interference with the colvar_grid classes.
> 
> What do you think @giacomofiorin ?

I'm okay with that; `colvargrid.cpp` is probably even larger than it looks with all the template instantiations.

jhenin:
I recommend that this be merged now, it's just shuffling files around and will limit future conflicts.

PR #803, opened 2025-05-12
Separate NAMD GlobalMaster and colvarproxy_namd classes

giacomofiorin:
Currently, `colvarproxy_namd` (also known as `GlobalMasterColvars`) is a derived class of `GlobalMaster`, consistent with what other NAMD features based on global forces do. This inheritance worked okay as long as we were using `GlobalMaster` as the API for communicating with NAMD. However, this also presented significant issues in #783, which added support for `CudaGlobalMasterClient`. The most significant was the duplication of functionality that is native to `colvarproxy_namd`, but could not be used without enabling the traditional `GlobalMaster` class as well.

This PR removes the inheritance between `GlobalMaster` and `colvarproxy_namd`, adding a `friend` declaration between the former and the latter. `GlobalMasterColvars` is now an actual class, is a thin wrapper around the second.

### DIFF
--- a/namd/src/GlobalMaster.h
+++ b/namd/src/GlobalMaster.h
@@ -8,7 +8,7 @@
    be done on atoms or groups of atoms on several nodes.  It is given
    the positions of atoms and groups, and provides a list of requested
    atoms, forces, and groups in return.
-   
+
    I'm not going to do groups for now, because they were done badly
    originally.  A better solution is necessary.  Hint: multiple
    groups are required.
@@ -71,13 +71,15 @@ class GlobalMaster {
   bool requestedTotalForces() { return totalForceRequested; }
 
   /* sets changedAtoms and changedForces to false again */
-  void clearChanged(); 
+  void clearChanged();
   virtual ~GlobalMaster(); // necessary for abstract classes '-P
 
   void check() const; // dies if there are problems with the rep invariants
 
   void setLattice(const Lattice *lat) { lattice = lat; }
-  
+
+  friend class colvarproxy_namd; // Temporary until refactoring is complete
+
  protected:
   GlobalMaster();
 
@@ -118,13 +120,13 @@ class GlobalMaster {
   IntList::const_iterator getGridObjIndexEnd();
   BigRealList::const_iterator getGridObjValueBegin();
   BigRealList::const_iterator getGridObjValueEnd();
-  
+
   /* these give you all the global forces being applied by masters */
   /* again, here we only need one end iterator */
   AtomIDList::const_iterator getLastAtomsForcedBegin();
   AtomIDList::const_iterator getLastAtomsForcedEnd();
   ForceList::const_iterator getLastForcesBegin();
-  
+
   /* These return the pointers to the lists of requested atom IDs
      and total forces on these atoms */
   AtomIDList::const_iterator getForceIdBegin();
@@ -133,7 +135,7 @@ class GlobalMaster {
 
   bool totalForceRequested;
   void requestTotalForce(bool yesno = true) { totalForceRequested = yesno; }
-  
+
   /* This helpful function returns an array with the masses of each of
      the groups whose positions we have.  */
   BigRealList::const_iterator getGroupMassBegin();
@@ -169,7 +171,7 @@ class GlobalMaster {
   AtomIDList::iterator lastAtomsForcedBegin;
   ForceList::iterator lastForcesBegin;
   AtomIDList::iterator lastAtomsForcedEnd;
-  
+
   /* These store all the total forces returned from the simulation */
   AtomIDList::iterator forceIdBegin;
   AtomIDList::iterator forceIdEnd;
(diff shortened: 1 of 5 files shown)


namd/src/GlobalMaster.h
```diff
@@ -0,0 +1,205 @@
...
+
+  /* sets changedAtoms and changedForces to false again */
+  void clearChanged();
+  virtual ~GlobalMaster(); // necessary for abstract classes '-P
+
+  void check() const; // dies if there are problems with the rep invariants
+
+  void setLattice(const Lattice *lat) { lattice = lat; }
+
+  friend class colvarproxy_namd; // Temporary until refactoring is complete
```
HanatoK, line 81: I think you can re-implement the functions required by `colvarproxy_namd` in `GlobalMasterColvars` as public, so that you can bypass the protected functions here.

namd/src/GlobalMasterColvars.h
```diff
@@ -0,0 +1,26 @@
+#ifndef GLOBALMASTERCOLVARS_H
+#define GLOBALMASTERCOLVARS_H
+
+#include <memory>
+
+#include "colvarmodule.h"
+#include "colvarproxy.h"
+#include "colvarproxy_namd.h"
```
HanatoK, line 8: In this file, only `GlobalMaster.h` needs to be included. You can forward declare `colvarproxy_namd`, and move the constructor to somewhere else, like a new `GlobalMasterColvars.C` or `colvarproxy_namd.C`.

namd/src/GlobalMaster.h
giacomofiorin: Totally agree. This is just a way to make it work short-term

HanatoK:
I guess the following functions can be shared between `colvarproxy_impl` in `colvarproxy_cudaglobalmaster.C` and `colvarproxy_namd`:
```
load_atoms_pdb
load_coords_pdb
update_target_temperature
log
error
check_atom_id
update_atom_properties
output_stream
flush_output_stream
flush_output_streams
close_output_stream
close_output_streams
backup_file
set_unit_system
```

giacomofiorin:
> I guess the following functions can be shared between `colvarproxy_impl` in `colvarproxy_cudaglobalmaster.C` and `colvarproxy_namd`:
> 
> ```
> ...
> ```

Exactly! Plus much of the code in the initialization of individual atoms, etc.

namd/src/GlobalMasterColvars.h
giacomofiorin: Indeed. I like the latter option better, since it's one fewer new file for a package that doesn't use CMake. There is a very similar init function in `colvarproxy_vmd.C`

namd/src/GlobalMaster.h
```diff
@@ -0,0 +1,205 @@
+/**
```
jhenin, line 1: Do we need to track GlobalMaster.h?

namd/src/GlobalMasterColvars.h
```diff
@@ -0,0 +1,26 @@
...
+
+
+class GlobalMasterColvars : public GlobalMaster {
+public:
+
+  GlobalMasterColvars() : cp(new colvarproxy_namd(this)) {}
+
+  void calculate() override {
+    cp->calculate();
```
jhenin, line 17: Maybe use more explicit identifiers than `cp`, and most importantly, `gm_`. This makes the code more hermetic.

For `cp`, why not `proxy`, and for `gm_`, maybe just `globalmaster`? The more obvious, the better.

jhenin:
I'm excited to see the code deduplication that will result from this!

giacomofiorin:
Current state is that the two classes are now separated, and things seem to be working. 

Next I'll be taking a look at consolidating with the `CUDAGlobalMaster`-based equivalent (see https://github.com/Colvars/colvars/pull/803#issuecomment-2874036919) to see how much would that entail, i.e. see whether if can fit in this PR or another one.

@HanatoK I'm considering to keep the files under the folder `namd/cudaglobalmaster` and not move them to `namd/src`, so that this repo can track their history correctly. Any thoughts?

jhenin:
If this is a complete functioning change set, I would recommend merging now, and doing the next step separately. The next PR will be more legible if it is limited in scope.

giacomofiorin:
> If this is a complete functioning change set, I would recommend merging now, and doing the next step separately. The next PR will be more legible if it is limited in scope.

Sounds good: lifting the draft status.

> Also, please move the files early, for the sake of long-term cleanliness of the directory tree. File history becomes less relevant over time, the directory tree remains.

I'm okay with that, but it would have to be done separately from this PR. Moving the files now would break the CMake recipe that currently builds the plugin.

HanatoK (approved):
Looks good to me!

namd/src/colvarproxy_namd.h
```diff
@@ -15,29 +15,29 @@
...
 #endif
 
-#include "Vector.h"
-#include "ResizeArray.h"
-#include "NamdTypes.h"
-#include "SimParameters.h"
-#include "Lattice.h"
-#include "GlobalMaster.h"
-#include "Random.h"
-#include "ConfigList.h"
```
jhenin: +1

namd/src/.clang-format
```diff
@@ -0,0 +1,14 @@
+---
```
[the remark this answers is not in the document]
giacomofiorin: Yeah, it's helping format new code consistently (also, it help me sort alphabetically that long list of includes)

 Not touching the old code yet.

PR #805, opened 2025-05-23
Auto-update Dashboard from repo

jhenin:
Maintaining a local copy of the cv_dashboard directory in the user's home directory.

- [x] check for new versions and download updates
- [x] switch to using the local version if available

This is now at the "works for me" stage. It needs to be tested:

- [ ] on MacOS
- [x] on Windows
- [ ] on systems without a git install

When offline, the automated, blocking network requests become a problem. So, new todo :

- [ ] Fail gracefully when offline (non-blocking network request)
- [ ] Option to disable (or make update manual only - prompt when update available)


jhenin:
This will probably not be integrated in this form. Closing.

PR #818, opened 2025-07-09
Account for NAMD's GlobalMasterFrequency in all biases

### LINKED ISSUE #793: extended Lagrangian is not compatible with NAMD's globalMaster MTS

giacomofiorin:
Finish the work of #795. Fixes #793 

Changes to general code:
- Initialize `timeStepFactor` from the proxy by default (users may override it in the configuration)
- Constrain timeStepFactor to be the same between colvars and their biases

Supported biases:
- [x] ABF (done in #795)
- [x] metadynamics
- [x] OPES
- [x] reweightAMD
- [x] moving restraints
- [x] ALB
- [ ] others?

### DIFF
--- a/src/colvarbias_alb.cpp
+++ b/src/colvarbias_alb.cpp
@@ -39,10 +39,11 @@ colvarbias_alb::colvarbias_alb(char const *key)
 int colvarbias_alb::init(std::string const &conf)
 {
   colvarproxy *proxy = cvm::main()->proxy;
-  int err = colvarbias::init(conf);
-  if (err != COLVARS_OK) {
-    return err;
+  int error_code = colvarbias::init(conf);
+  if (error_code != COLVARS_OK) {
+    return error_code;
   }
+
   cvm::main()->cite_feature("ALB colvar bias implementation");
 
   enable(f_cvb_scalar_variables);
@@ -80,21 +81,34 @@ int colvarbias_alb::init(std::string const &conf)
     }
   } else {
     colvar_centers.clear();
-    cvm::error("Error: must define the initial centers of adaptive linear bias .\n");
+    error_code |= cvm::error("Error: must define the initial centers of adaptive linear bias.\n",
+                             COLVARS_INPUT_ERROR);
   }
 
-  if (colvar_centers.size() != num_variables())
-    cvm::error("Error: number of centers does not match "
-                      "that of collective variables.\n");
+  if (colvar_centers.size() != num_variables()) {
+    error_code |=
+        cvm::error("Error: number of centers does not match that of collective variables.\n");
+  }
 
-  if (!get_keyval(conf, "UpdateFrequency", update_freq, 0))
-    cvm::error("Error: must set updateFrequency for adaptive linear bias.\n");
+  if (!get_keyval(conf, "updateFrequency", update_freq, 0)) {
+    error_code |= cvm::error("Error: must set updateFrequency for adaptive linear bias.\n",
+                             COLVARS_INPUT_ERROR);
+  }
+
+  if (update_freq % time_step_factor != 0) {
+    error_code |= cvm::error("updateFrequency (currently " + cvm::to_str(update_freq) +
+                                 ") must be a multiple of timeStepFactor (" +
+                                 cvm::to_str(time_step_factor) + ").\n",
+                             COLVARS_INPUT_ERROR);
+  }
 
   //we split the time between updating and equilibrating
   update_freq /= 2;
 
-  if (update_freq <= 1)
-    cvm::error("Error: must set updateFrequency to greater than 2.\n");
+  if (update_freq <= 1) {
+    error_code |=
+        cvm::error("Error: must set updateFrequency to greater than 2.\n", COLVARS_INPUT_ERROR);
+  }
 
   enable(f_cvb_history_dependent);
 
@@ -136,7 +150,7 @@ int colvarbias_alb::init(std::string const &conf)
   if (cvm::debug())
     cvm::log(" bias.\n");
 
-  return COLVARS_OK;
+  return error_code;
 }
 
 


src/colvarbias.cpp
```diff
@@ -143,6 +146,37 @@ int colvarbias::init(std::string const &conf)
...
+  if (time_step_factor % cvm::proxy->time_step_factor() != 0) {
+    error_code |=
+        cvm::error("timeStepFactor for this bias (currently " + cvm::to_str(time_step_factor) +
+                       ") must be a multiple of the global Colvars timestep multiplier (" +
+                       cvm::to_str(cvm::proxy->time_step_factor()) + ").\n",
+                   COLVARS_INPUT_ERROR);
+  }
+
+  for (auto *cv : colvars) {
+    if (cv->get_time_step_factor() != time_step_factor) {
```
jhenin, line 167: I suppose the bias' factor could also be a multiple of of the colvar's, right?

jhenin: Metadynamics or OPES with many kernels can become costly, so MTS would make sense - and then you can have a cheap harmonic restraint that you want at every step.

jhenin: Good point about ABF: for ABF, you can do MTS only if you get forces from the same timestep. The NAMD-style deferred forces scheme will break it, I need to add that condition.

jhenin: Then maybe metadynamics in higher dimension, when you don't want to use grids anymore :-)

HanatoK: Frankly speaking, I seldom use the MTS for free-energy calculations. Making the bias time step factor to support a multiple of CV time step factors could be useful if the cost of computing bias is high. OPES doesn't use grids, but it does merge Gaussian kernels and use a neighbor list to reduce the number of hills. For metadynamics, I think @jhenin is right that someone may use metadynamics for more than 4 CVs and avoid the grids, and then use reweighting to estimate the free-energy difference of two states. In such case, the cost of metadynamics could be expensive, and therefore one would resort to MTS.

src/colvarbias_opes.cpp
```diff
@@ -271,7 +284,11 @@ int colvarbias_opes::init(const std::string& conf) {
     key_lookup(conf, "grid", &grid_conf);
     m_reweight_grid.reset(new colvar_grid_scalar(m_pmf_cvs, nullptr, false, grid_conf));
     m_pmf_grid.reset(new colvar_grid_scalar(m_pmf_cvs, m_reweight_grid));
-    get_keyval(conf, "pmfHistoryFrequency", m_pmf_hist_freq, 0);
+    get_keyval(conf, "pmfHistoryFrequency", m_pmf_hist_freq, output_freq);
+    if ((m_pmf_hist_freq % output_freq) != 0) {
+      error_code |= cvm::error("Error: pmfHistoryFrequency must be a multiple of outputFreq.\n",
+                               COLVARS_INPUT_ERROR);
+    }
```
giacomofiorin, line 291: @HanatoK Here, I wasn't completely sure between checking against `output_freq` or `time_step_factor`. We could go either way.

src/colvarbias_opes.cpp
```diff
@@ -287,6 +304,12 @@ int colvarbias_opes::init(const std::string& conf) {
   m_traj_line.neff = (1 + m_sum_weights) * (1 + m_sum_weights) / (1 + m_sum_weights2);
   m_traj_line.nker = m_kernels.size();
   get_keyval(conf, "printTrajectoryFrequency", m_traj_output_frequency, cvm::cv_traj_freq);
+  if (m_traj_output_frequency % time_step_factor != 0) {
+    error_code |= cvm::error(
+        "printTrajectoryFrequency (currently " + cvm::to_str(m_traj_output_frequency) +
+            ") must be a multiple of timeStepFactor (" + cvm::to_str(time_step_factor) + ").\n",
+        COLVARS_INPUT_ERROR);
+  }
```
giacomofiorin, line 312: @HanatoK Same here, I'm not sure what the best divisor would be.

giacomofiorin:
> One case I can think of that would fall under "others" is scripted forces. Those should use the getstepabsolute and getsteprelative functions to make sure they know at what timestep they are being called.

I agree! But to confirm, the "scaled" step counter of GlobalMaster is not being exposed to Tcl. So the problems would happen when scripted forces do their own step counting, which would probably be wrong anyway in the `run 0` case?

HanatoK (approved):

giacomofiorin:
@jhenin Besides the Tcl scripted forces case, do you think we're good here?

PR #820, opened 2025-07-10
Loop over replicas when computing well-tempered metadynamics scaling factor

### LINKED ISSUE #819: Combine Gaussians for multiple-walkers metadynamics before applying well-tempered scaling
Multiple-walkers metadynamics (Raiteri et al, 2005) and well-tempered metadynamics (Barducci et al, 2008) are two separate variants that are occasionally combined together. The Barducci paper didn't explicitly state which biasing potential (partial or total) is scaled the bias factor. Colvars scales the partial biasing potential of each replica, whereas PLUMED scales the total one, as mentioned in #815.

Independently from that, however, the `write_pmf()` function in Colvars uses the *total* biasing potential to extract the free-energy landscape:
https://github.com/Colvars/colvars/blob/e06f5e6949d9bba85e54cca4534f7eb86a176f26/src/colvarbias_meta.cpp#L1949-L1952

Because fixing this inconsistency would break backward compatibility, it would be better to adopt the same convention that PLUMED uses.

giacomofiorin:
Fixes #819

### DIFF
--- a/src/colvarbias_meta.cpp
+++ b/src/colvarbias_meta.cpp
@@ -537,6 +537,7 @@ int colvarbias_meta::update_grid_params()
 
 int colvarbias_meta::update_bias()
 {
+  int error_code = COLVARS_OK;
   colvarproxy *proxy = cvm::main()->proxy;
   // add a new hill if the required time interval has passed
   if (((cvm::step_absolute() % new_hill_freq) == 0) &&
@@ -548,7 +549,8 @@ int colvarbias_meta::update_bias()
                ": adding a new hill at step "+cvm::to_str(cvm::step_absolute())+".\n");
     }
 
-    cvm::real hills_scale=1.0;
+    // Scaling factor, optionally used by EB-metadynamics or well-tempered metadynamics
+    cvm::real hills_scale = 1.0;
 
     if (ebmeta) {
       hills_scale *= 1.0/target_dist->value(target_dist->get_colvars_index());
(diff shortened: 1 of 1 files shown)


src/colvarbias_meta.cpp
```diff
@@ -561,36 +563,40 @@ int colvarbias_meta::update_bias()
...
-            calc_hills(hills.begin(),
-                       hills.end(),
-                       hills_energy_sum_here,
-                       &colvar_values);
+            if (!keep_hills) {
+              calc_hills(replicas[ir]->hills_off_grid.begin(), replicas[ir]->hills_off_grid.end(),
+                         hills_energy_sum_here, &colvar_values);
+            } else {
+              // TODO: Is it better to compute the energy from all historic hills
+              //       when keepHills is on?
```
giacomofiorin, line 582: @HanatoK I didn't understand this comment before, and forgot to ask you about it

src/colvarbias_meta.cpp
```diff
@@ -561,36 +563,40 @@ int colvarbias_meta::update_bias()
...
+              calc_hills(replicas[ir]->hills.begin(), replicas[ir]->hills.end(),
+                         hills_energy_sum_here, &colvar_values);
+            }
+            // cvm::log("WARNING: computing bias factor for off-grid hills. Hills energy: " +
+            // cvm::to_str(hills_energy_sum_here) + "\n");
           }
-          // cvm::log("WARNING: computing bias factor for off-grid hills. Hills energy: " + cvm::to_str(hills_energy_sum_here) + "\n");
+        } else {
+          calc_hills(replicas[ir]->hills.begin(), replicas[ir]->hills.end(), hills_energy_sum_here,
```
HanatoK, line 590: Now I feel the old code summing the hills starting from `new_hills_begin` is incorrect, and this fixes it. Is that right?

HanatoK (changes requested):
In my opinion, the logic of computing bias energy should be the same as `colvarbias_meta::calc_energy` without adding the new hills:
```c++
  bool index_ok = false;
  std::vector<int> curr_bin;

  if (use_grids) {

    curr_bin = values ?
      hills_energy->get_colvars_index(*values) :
      hills_energy->get_colvars_index();

    index_ok = hills_energy->index_ok(curr_bin);

  }

  if ( index_ok ) {
    // index is within the grid: get the energy from there
    for (ir = 0; ir < replicas.size(); ir++) {

      bias_energy += replicas[ir]->hills_energy->value(curr_bin);
      if (cvm::debug()) {
        cvm::log("Metadynamics bias \""+this->name+"\""+
                 ((comm != single_replica) ? ", replica \""+replica_id+"\"" : "")+
                 ": current coordinates on the grid: "+
                 cvm::to_str(curr_bin)+".\n");
        cvm::log("Grid energy = "+cvm::to_str(bias_energy)+".\n");
      }
    }
  } else {
    // off the grid: compute analytically only the hills at the grid's edges
    for (ir = 0; ir < replicas.size(); ir++) {
      calc_hills(replicas[ir]->hills_off_grid.begin(),
                 replicas[ir]->hills_off_grid.end(),
                 bias_energy,
                 values);
    }
  }
```

src/colvarbias_meta.cpp
giacomofiorin: I'm not sure, sorry. This branch corresponds to the grids-off, multiple-walkers, well-tempered use case.

With this change, the branch is more consistent with the other options. As for whether it was previously buggy, I think we should first check if anybody ever ran this case before investigating further.

src/colvarbias_meta.cpp
giacomofiorin: If you're doing that, then you would effectively disable `use_grids` without the user being able to check.

To be honest, at this point trajectories have become so long that the usefulness of `keepHills` is put into question.

src/colvarbias_meta.cpp
```diff
@@ -561,36 +563,40 @@ int colvarbias_meta::update_bias()
...
-            // TODO: Should I sum the off-grid hills from other replicas?
-            calc_hills(hills_off_grid.begin(),
-                       hills_off_grid.end(),
-                       hills_energy_sum_here,
-                       &colvar_values);
+
+      for (size_t ir = 0; ir < replicas.size(); ir++) {
+        if (use_grids) {
+          std::vector<int> curr_bin = hills_energy->get_colvars_index();
+          const bool index_ok = hills_energy->index_ok(curr_bin);
```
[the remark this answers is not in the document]
giacomofiorin: I moved the initialization of `index_ok` outside of the loop. As far as moving the check, I'm not sure whether it's worth it (meaning, I don't have enough bandwidth to do this, but wouldn't oppose you if you did).

src/colvarbias_meta.cpp
HanatoK: No. What I suggested should only affect off-grid positions. If you use both `use_grids` and `keepHills`, then to calculate the energy of a position out of the grid, you can sum all hills, and to calculate the energy of a position in the grid, you can just use the grid value. This shouldn't disable `use_grids`.

giacomofiorin: Ok, thanks. The "off-grid hills", which are kept regardless, should also include hills whose center is *inside* the grid boundaries, but they are within sufficient range that they would affect configurations just outside the grid. At least that was the original logic, assuming that it didn't break along the way.

Looping over all hills would also include hills that are well within the boundaries.

PR #825, opened 2025-08-06
Moving harmonic walls - Refactor moving restraints

OmriShf:
**Feature suggestion: Moving Harmonic Walls**

This feature enables the user to change the location of biases of the "harmonic wall" form on the colvar during the simulation, either smoothly or in defined stages, similarly to the existing "Changing force constant" feature. 

Two keywords have been added:
**TargetUpperWalls** - requires UpperWalls to exist, as initial locations for the wall. Signifies the final location of the upper wall for smooth transitions or the location of the wall during the last stage for a change in stages. Linear interpolation is done for the in-between steps/stages.  cannot be at any step lower than LowerWall for the same colvar. 
**TargetLowerWalls** - The same thing, for lower walls.  

I implemented the feature directly into the existing "colvarbias_restraint_harmonic_walls" class, and it might be inelegant or unfitting for the general structure of the library. Please let me know! I did as much testing as I could, and it seems to work as expected in all cases, including simultaneous change of the string const and the wall locations.   

a specific concern:
* The smooth version currently outputs the location of the walls to the log after every step; it is useful but might be overly verbose. I thought about adding an additional print freq user parameter. 

**JH update**

This has grown to a welcome overhaul of the moving restraint code. Remaining TODOs:
- [x] update regression tests for LAMMPS and Gromacs
- [ ] add regression tests for new feature

TODO for another PR:
- [ ] allow and test combining changing force constant and moving walls / centers.


jhenin:
This addition made it clear that the moving restraint code was redundant and inconsistent. I took the chance to factorize all the moving restraint logic. In the process, the restraint free energy calculation becomes general too. It also becomes possible (in principle) to apply perturbations to centers (or walls) and force constants at the same time.
Because there are many combinations of parameters, this needs testing for completeness.

jhenin:
Moved to main repo as #851

PR #831, opened 2025-08-20
Gradient tests

jhenin:
Adds automated tests for gradient accuracy in standalone functional tests, as discussed in #828 

In the process, fix the calculation of the coordnum gradients with pair lists (the error decreased with the tolerance, making it less noticeable).


giacomofiorin (approved):
All looks great! This is indeed exactly what we needed.

Further improvements could be adding trajectories that probe a broader set of configurations, but this can be done separately.

PR #835, opened 2025-08-21
Initialize quaternions to zero, rotations to unity

jhenin:
to support the unlikely case of combinations of quaternion cvcs.

Rotation objects are now initialized to unity, which is more consistent.

### DIFF
--- a/src/colvartypes.h
+++ b/src/colvartypes.h
@@ -980,12 +980,10 @@ public:
     reset();
   }
 
-  /// \brief Set the q0 component to 1 and the others to 0 (quaternion
-  /// representing no rotation)
+  /// \brief Set all components to zero (null quaternion)
   inline void reset()
   {
-    q0 = 1.0;
-    q1 = q2 = q3 = 0.0;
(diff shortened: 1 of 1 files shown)


src/colvartypes.h
```diff
@@ -1341,7 +1339,7 @@ class colvarmodule::rotation
   bool b_debug_gradients;
 
   /// \brief The rotation itself (implemented as a quaternion)
-  cvm::quaternion q;
+  cvm::quaternion q{1.0, 0.0, 0.0, 0.0};
```
HanatoK, line 1342: If I remember correctly the initialization in class declaration requires C++11, but we use C++11 anyway now so it would be OK.

HanatoK (approved):

src/colvartypes.h
jhenin: Yes, I'm enjoying the fact that we finally set this as a minimum! Hopefully VMD will move forward and we can adopt an even more modern standard :-)

PR #843, opened 2025-08-27
Minor changes to maintenance script to parse GitHub metadata

giacomofiorin:
Some fixes and updates to the script to allow listing the individual commits from PRs after their branches are rebased onto master.

Note: this requires that for each PR the original branch is available in the repo, because the GH CLI returns the *original* commit hashes, not the rebased ones.  It would be useful if we kept around the branches for bugfix PRs until the step of cherry-picking into release branches is automated.

### DIFF
--- a/devel-tools/print_pr_list.py
+++ b/devel-tools/print_pr_list.py
@@ -27,9 +27,9 @@ def affects_backend(labels, backend=None):
     return False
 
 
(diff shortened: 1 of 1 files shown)


giacomofiorin:
Merging because the test failures are addressed separately

Issue #809, opened 2025-06-10
Comparable Colvars version


HanatoK:
Currently version of Colvars is set by the `colvars_version.h`:

https://github.com/Colvars/colvars/blob/ec89df12686346d921788a5832231b18f1589143/src/colvars_version.h#L1-L3

It is a bit difficult to compare different Colvars version strings and compile external C++ code conditionally based on different versions of Colvars. I am a bit worried that by introducing https://github.com/Colvars/colvars/pull/788, any 3rd C++ code (for example, those code in https://www.ks.uiuc.edu/Training/Tutorials/#namd:~:text=Customizing%20CVs%20in%20NAMD%3A) relying on the Colvars internal APIs won't work, so it would be better to provide a way to conditionally compile C++ code based on the Colvars version.

I propose the following solution:
```c++
#ifndef COLVARS_VERSION

#include <cstdint>

#define COLVARS_VERSION_YEAR 2025
#define COLVARS_VERSION_MONTH 05
#define COLVARS_VERSION_DATE  12

#define COLVARS_VERSION_STR_HELPER(x) #x
#define COLVARS_VERSION_STR(x) COLVARS_VERSION_STR_HELPER(x)

#define COLVARS_VERSION \
        COLVARS_VERSION_STR(COLVARS_VERSION_YEAR) "-"\
        COLVARS_VERSION_STR(COLVARS_VERSION_MONTH) "-"\
        COLVARS_VERSION_STR(COLVARS_VERSION_DATE)

#define COLVARS_VERSION_UINT64(a,b,c) \
        ((UINT64_C(a) << 16) + (UINT64_C(b) << 8) + UINT64_C(c))
#define COLVARS_VERSION_UINT64_CURRENT \
        COLVARS_VERSION_UINT64(COLVARS_VERSION_YEAR, \
                               COLVARS_VERSION_MONTH, \
                               COLVARS_VERSION_DATE)

#endif
```

But it looks like that will break other shell scripts.

jhenin:
I agree. If we make the changes you mention, the main scripts that will break are ours - if we adapt them, I doubt we'll break any other people's workflows, and if that happens, they will be developers who can just copy the logic from our updated version scripts.

By the way: nice job with the tutorial! I hadn't seen it. I think we should link to it from the Colvars documentation, because it's not very backend-specific (especially since you can build all codes with Tcl support :-)

HanatoK:
[...]

PR #847, opened 2025-08-28
Define version numbers outside of headers

giacomofiorin:
We have three version numbers that are changing over time:
- `COLVARS_VERSION` (changes most frequently) 
- `COLVARPROXY_VERSION` (changes less frequently, and is also less used; still there mostly for inertia)
- `colvarmodule::patch_version_int` (we started adopting it more recently, for bugfix releases)

Because these macros/variables are defined in widely used headers, they often invalidate the compiler cache, making CI tests unnecessarily slow. This PR attempts to fix that by moving them outside those headers, and including them only where needed.

Additionally, this PR stops tracking GROMACS MDModule files to simplify things, and using diff files instead (happy to undo part of all of this if there is a good reason to).

@HanatoK  Since the changes are only inside `colvars_version.h` I think they should not interfere with what you had in mind in #809?


HanatoK (approved):
Looks good to me!

PR #858, opened 2025-09-09
Upgrade compilers

giacomofiorin:
Drop unsupported compilers and CentOS 7 container, and add new ones. Now the CentOS 9 is the only Linux platform being tested beside Microsoft/Ubuntu. One of the test cases with GCC 11 is downgraded from C++20 to C++17, to allow more testing there. The cache for the x86_64 Linux job is intentionally invalidated to allow dropping the old compilers' data immediately.

Also fix the Clang static analyzer CI job, which wasn't triggering an error.

Side note: I have also uploaded a CentOS 10 container, but after checking what it comes with it is not sufficiently different to warrant being used yet.

### DIFF
--- a/.github/workflows/test-library.yml
+++ b/.github/workflows/test-library.yml
@@ -236,7 +236,7 @@ jobs:
 
 
   address-sanitizer:
-    name: Clang address sanitizer
+    name: Linux x86_64 (Clang address sanitizer)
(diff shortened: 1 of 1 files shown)


jhenin:
It is a false positive, the static analyzer cannot see that the data is initialized by get_grad. But the cost of initializing explicitly should be low.

PR #859, opened 2025-09-09
Add clang-tidy CI workflow

giacomofiorin:
In draft for now: even with the default settings, the current code does not pass the test.


giacomofiorin:
@HanatoK Continuing the conversation here, one of the advantages of using a container is to use a recent version of the Clang tools. The `ubuntu-latest` GitHub instance runs 18, but CentOS 9 comes with version 20.

giacomofiorin:
With the downgrade from Clang 20 to 18 and removal of the Lepton sources, we are down to two issues, one with `colvarproxy_io::io_available()` and one with the bitwise OR of values of `Parse_Mode` type.

I have a good idea about how to fix the former (requires some small refactoring) but I'm puzzled by the latter: I though that having added the `clang::flag_enum` annotation would have taken care of it, but was wrong. Any ideas?

devel-tools/build_test_library.cmake
```diff
@@ -142,12 +142,16 @@ if(NOT result EQUAL 0)
...
-  --parallel
-  RESULT_VARIABLE result
+option(BUILD_TARGETS "Build library and related tools" OFF)
+
+if(BUILD_TARGETS)
+  execute_process(
+    COMMAND ${CMAKE_COMMAND}
+    --build ${BUILD_DIR}
+    --parallel
+    RESULT_VARIABLE result
```
HanatoK, line 152: @giacomofiorin It broke the basic check.

PR #877, opened 2025-11-06
Trap floating-point exceptions

jhenin:
This detects a couple of divide-by-zero situations that escaped us so far!


jhenin:
Thank you @HanatoK

giacomofiorin (approved):
Looks great! I hope the added CI test will tell us about new occurrences.

PR #884, opened 2025-11-17
fix: use correct ref coords for RMSD Jacobian

### LINKED ISSUE #871: Possibly incorrect Jacobian in RMSD using atomPermutation
In my opinion, the code in https://github.com/Colvars/colvars/blob/2e8764999f18e463d31021e9ae4e32430b5f504f/src/colvarcomp_distances.cpp#L1075-L1082 should select the reference positions with the best permutation. I am not sure, though.

jhenin:
This generalizes correctly the standard RMSD implementation.
Numerical effects are very small.

Fixes #871 reported by @HanatoK


jhenin:
Addressed by #881

PR #886, opened 2025-11-18
Refactor: replace static colvarmodule with class member

jhenin:
In a first stage, just trying to keep functionality intact, removing all static member data from class `colvarmodule`.

First approach: make a colvarmodule pointer a member of colvarparse, because the big classes inherit from it.

Now a number of lightweight classes (in colvartypes) or sub-classes don't have access to the pointer yet. Alternatives are to extend them with a pointer, or remove the functionality that needs it (mostly `log` and `error` calls).

EDIT: because this is so far-reaching, I am turning it into a partial implementation that can be merged in finite time, leaving some of the classes to be modified later.

Namely:
- **colvargrids_** - make broad use of cvm functions, passing pointers to the module between them is very heavy
- **colvarproxy_\*** base classes. They all implement partial functionality for colvarproxy. I am not sure anymore that inheritance is the best paradigm, or we would need a common base class colvarproxy_base at the root of a diamond inheritance.

**colvarbias** classes have been handles thanks to virtual inheritance: only the colvarbias base class constructor sets the cvmodule pointer, as passed on by the final classes. Intermediate classes don't have to worry about it.

**Final status**: the remaining uses of the static pointer can be tracked by looking for calls to `cvm::main()`. There are new functions `log_static` and `error_static` that make use of the static pointer only if available, and have fallbacks if not. All of these can be phased out in later PRs.

### DIFF
--- a/src/colvar_gpu_calc.cpp
+++ b/src/colvar_gpu_calc.cpp
@@ -71,7 +71,7 @@ int colvarmodule_gpu_calc::cvc_calc_total_force(
   if (cvmodule->debug()) {
     if (g.graph_exec_initialized) {
       // Debug graph
-      const std::string filename = cvm::output_prefix() + "_calc_total_force.dot";
+      const std::string filename = cvmodule->output_prefix() + "_calc_total_force.dot";
       cvmodule->log("Writing calc_total_force graph to " + filename);
       g.dump_graph(filename.c_str());
     }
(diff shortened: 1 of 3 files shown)


giacomofiorin:
My main concern is, all Colvars functionality relies on there being a single instance of the module.  Removing static pointers would also require adding a different kind of check that there is only one `colvarmodule` object.

jhenin:
In practice, the main application I've had in mind so far would be associating one colvarmodule instance with one molecule in VMD, although that would need more work because 1) VMD has a single Tcl interpreter, and 2) so far we have single proxy object - so we'd have to choose between duplicating the proxy object as well, or let a single proxy switch between molecules. Given the current architecture, I think one proxy per molecule would be simpler.

Thanks to Giacomo's suggestion with `Tcl_SetAssocData`, I think this is workable: the interpreter could register not just one pointer, but say, one proxy pointer for each loaded molecule. If such a pointer were missing, the script interface would know to allocate a new instance for that molecule.

You may have more ideas for using multiple instances: I'm interested.

jhenin:
To answer  @giacomofiorin 's remark about relying on there being a single instance: my goal here is that several instances could exist and that all calls are addressed to the right objects. One point to be worked out is whether they attach to the same or different proxy objects. For now, I am working under the assumption of a proxy object talking to a single colvarmodule instance.

Issue #892, opened 2025-11-22
cv configfile overwrite centers from NAMD colvarsInput


Trayder:
I was restarting a SMD run in NAMD and found that when specifying `colvarsInput` and then later `cv configfile`  in the NAMD script, the `cv configfile` overwrites the centers from the input state (it still keeps the step number). In the logfile it reports loading the restart before the config.

colvarsInput and colvarsConfig behave correctly when provided in either order.

Obviously mixing approaches is a bit weird, but the manual only mentions colvarsInput in "3.6.1 Restarting in NAMD." and strongly implies the approaches are interchangeable in "3.2 NAMD parameters". Further confusion comes from the log file reporting that the input state has been loaded and later listing the expected initial step.

jhenin:
Here is a proposal for a revised section 3.6.1:

> When the initial configuration is provided using `colvarsConfig`, `colvarsInput` can be used to give
> the name of a state file to be read immediately at startup. Once the Colvars module is initialized,
> a state file may be loaded dynamically at any time with the Tcl command `cv load`. Both versions
> support loading Colvars state files in either format (binary or text).

giacomofiorin:
I think we should remind again which options have effect before or after Tcl scripting starts. It is a NAMD issue, but one that when I describe it to colleagues I get this reaction:
🤯

jhenin:
Yes! This works very well once you understand it, but it's a tall step in the learning curve.

giacomofiorin:
Hi @Trayder we made the doc changes, which are merged into the `master` branch (see #893)  The updated doc should be online in a few minutes https://colvars.github.io/master/colvars-refman-namd.html

Does that look like it takes care of this issue?

PR #893, opened 2025-11-22
Warn when discarding state of unknown object

jhenin:
May help with issue #892

### DIFF
--- a/doc/colvars-refman-main.tex
+++ b/doc/colvars-refman-main.tex
@@ -447,7 +447,8 @@ Because these are static parameters, it is typically more convenient to use the 
     Input state file for the collective variables}{%
     UNIX filename}{%
     Keyword used to specify the input state file's name (\ref{sec:colvars_input}).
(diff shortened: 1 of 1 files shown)


src/colvarmodule.cpp
```diff
@@ -1822,9 +1823,15 @@ std::istream & colvarmodule::read_objects_state(std::istream &is)
...
 
     if (is.tellg() == pos) {
-      // This block has not been read by any object: discard it and move on
-      // to the next one
-      is >> colvarparse::read_block(word, NULL);
+      // This block has not been read by any object: discard it , print warning and move on
+      std::string conf;
+      is >> colvarparse::read_block(word, &conf);
+      std::string name;
+      if (parse_restart.key_lookup(conf, "name", &name)) {
```
giacomofiorin: `key_lookup()` modifies internal data: are you okay if we declare `parse_restart` right in this block?

PR #901, opened 2025-12-16
fix: initialize variable, appease Coverity

jhenin:


jhenin:
Just checking that tests pass.

PR #902, opened 2025-12-19
Automatically generate NAMD makefile

giacomofiorin:
This PR improves automates the step of adding Colvars source files to the file `Makefile.namd`, which is included by the NAMD main Makefile to define the Colvars library object files.

Unfortunately, updating `Make.depends` requires having a software environment already set up (including CUDA), so it is not suitable for automation at this time.

### DIFF
--- a/devel-tools/compile-namd.sh
+++ b/devel-tools/compile-namd.sh
@@ -77,6 +77,27 @@ EOF
         cmd+=(--charm-arch multicore-linux-${charm_arch_string})
     fi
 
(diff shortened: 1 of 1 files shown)


giacomofiorin:
@HanatoK My idea was to add another call to the `generate-namd-makefile.sh` script for the object files build in the CUDAGM plugin (see https://gitlab.com/tcbgUIUC/namd/-/merge_requests/484). Checking with you before going further with that.

PR #904, opened 2026-01-08
Fix: Implement numerically safe switching function

### LINKED ISSUE #903: The switching function implementation in coordnum is not numerically robust.
**Describe the bug**
The switching function implementation in coordnum is not numerically robust.
The code in https://github.com/Colvars/colvars/blob/c29f553ed83ba93077f3a348b2a2c80bbdda1863/src/colvarcomp_coordnums.cpp#L61 does not account for the case where `l2 == 1.0`, which should then be calculated with L'Hôpital's rule. The PLUMED version is more numerically robust:
https://github.com/plumed/plumed2/blob/7e8302f0d21c52b3659f1c4b7dd09ae19ebedb0f/src/tools/SwitchingFunction.cpp#L265-L277.

**Back-end (i.e. package that uses Colvars)**
All.

**Colvars version**
All.

**Expected behavior**
Colvars should switch to a numerically robust implementation of the switching function.

**Output log**
No.

jhenin:
replacing function and derivative with order 1 Taylor expansions around indeterminate form 0/0 for original expression

fixes #903

### DIFF
--- a/src/colvarcomp_coordnums.cpp
+++ b/src/colvarcomp_coordnums.cpp
@@ -61,6 +61,7 @@ cvm::real colvar::coordnum::switching_function(cvm::real const &r0,
   cvm::real const h = l2 - 1.0;
   cvm::real const en2_r = (cvm::real) en2;
   cvm::real const ed2_r = (cvm::real) ed2;
+  cvm::real func_no_pairlist;
 
   if (std::abs(h) < eps_l2) {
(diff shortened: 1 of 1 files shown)


HanatoK:
I attach my experiment results with cutoff = 6.0 and second-order Taylor expansion here:

out_y
out_dy

My test code can be found in https://godbolt.org/z/fnjnKoTqd.

PR #909, opened 2026-01-21
Simplify and optimize coordination numbers

### LINKED ISSUE #908: Incorrect number of pairs and doc error in selfCoordNum
**Describe the bug**
The number of pairs in selfCoordNum is calculated from a wrong formula:
$N_{\mathrm{pairs}} = (N-1)(N-1)$
However, it should be:
$N_{\mathrm{pairs}} = N(N-1)/2$
(as correctly reported in the [doc](https://colvars.github.io/master/colvars-refman-gromacs.html#sec:cvc_selfCoordNum)).

Additionally, the doc wrongly lists `cutoff3` as an allowed keyword.

This expression is used in two places:
1) To defined a default `upperBoundary` for the `selfCoordNum` function, which would only apply if the user doesn't provide any.
2) To allocate a pairlist buffer, but this is harmless because it simply results in a bigger buffer than needed.

**Back-end (i.e. package that uses Colvars)**
All

**Colvars version**
All since 020f555f

**Expected behavior**
The number of pairs should be calculated correctly

**Output log**
Not message in the log: the number of pairs is used to define default boundaries for the CV internally, which would only apply if the user doesn't provide any.

giacomofiorin:
This PR contains several improvements:
- [x] Condense the computation of multiple variants of coordination numbers: all classes are now derived from `coordnum` with the exception of `h_bond`
- [x] Separate the pairlist update from the switching function computation
- [x] Skip computing the switching function if pairwise distance is too large for the given tolerance
- [x] Optionally, allow using an inline function to compute PBC distances instead of a virtual function

Fixes #908

### DIFF
--- a/src/colvarcomp_coordnums.cpp
+++ b/src/colvarcomp_coordnums.cpp
@@ -232,11 +232,13 @@ void inline colvar::coordnum::main_loop()
          compute_pair_coordnum<flags | ef_use_internal_pbc>(inv_r0_vec, inv_r0sq_vec, en, ed,
                                                             x1, y1, z1, x2, y2, z2,
                                                             gx1, gy1, gz1, gx2, gy2, gz2,
-                                                            tolerance, tolerance_l2_max) :
+                                                            tolerance, tolerance_l2_max,
+                                                            cvmodule) :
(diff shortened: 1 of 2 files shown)


jhenin:
@giacomofiorin Do you wish to merge this first, and then #913 into master, or merge #913 into this?

PR #912, opened 2026-02-03
colvars_grid improvements

jhenin:
- differentiate scalar grids in any dimension
- prettier grid output with less rounding error artefacts


jhenin:
Merging, since this is orthogonal to the rest of the Universe.

PR #919, opened 2026-02-12
Handle boundary conditions internally

giacomofiorin:
This PR removes all calls to MD engines' codes for computing distances using the minimum-image convention. Now, the periodic boundary conditions are copied by the respective proxy class into a single object of type `colvarmodule::system_boundary_conditions`.

This class should be trivially copyable, and it is implemented in a single header file that uses the `COLVARS_HOST_DEVICE` macro. Therefore, the same code should work in both host and device, and could assist in computing PBC-aware distances on the device without duplicating existing code.

### DIFF
--- a/src/colvarcomp.cpp
+++ b/src/colvarcomp.cpp
@@ -458,7 +458,7 @@ void colvar::cvc::read_data()
 {
   if (is_enabled(f_cvc_pbc_minimum_image)) {
     // Copy boundary conditions from the proxy
-    boundary_conditions = cvm::main()->proxy->get_system_boundaries();
+    boundary_conditions = cvmodule->proxy->get_system_boundaries();
   } else {
     // Set as non-periodic boundary conditions (default) for this CVC
     boundary_conditions.reset();
--- a/src/colvarcomp.cpp
+++ b/src/colvarcomp.cpp
@@ -456,7 +456,13 @@ int colvar::cvc::set_param(std::string const &param_name,
 
 void colvar::cvc::read_data()
 {
-  boundary_conditions = cvm::main()->proxy->get_system_boundaries();
+  if (!is_enabled(f_cvc_pbc_minimum_image)) {
+    // Copy boundary conditions from the proxy
+    boundary_conditions = cvm::main()->proxy->get_system_boundaries();
+  } else {
+    // Set as non-periodic boundary conditions (default) for this CVC
+    boundary_conditions.reset();
+  }
 
   if (is_enabled(f_cvc_explicit_atom_groups)) {
     for (auto agi = atom_groups.begin(); agi != atom_groups.end(); agi++) {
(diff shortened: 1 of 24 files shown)


src/colvarcomp.cpp
```diff
@@ -456,6 +456,14 @@ int colvar::cvc::set_param(std::string const &param_name,
 
 void colvar::cvc::read_data()
 {
+  if (!is_enabled(f_cvc_pbc_minimum_image)) {
+    // Copy boundary conditions from the proxy
+    boundary_conditions = cvm::main()->proxy->get_system_boundaries();
```
HanatoK: Maybe it is better to use `cvmodule->proxy`?

HanatoK: The if condition here should be flipped to pass the new tests in https://github.com/Colvars/colvars/pull/936.

src/colvarproxy_system.h
```diff
@@ -214,81 +185,9 @@ class colvarproxy_system {
...
-  /// Type of boundary conditions
-  Boundaries_type boundaries_type;
-
-  /// Bravais lattice vectors
-  cvm::rvector unit_cell_x, unit_cell_y, unit_cell_z;
-
-  /// Reciprocal lattice vectors
-  cvm::rvector reciprocal_cell_x, reciprocal_cell_y, reciprocal_cell_z;
+  /// Current system boundary conditions
+  cvm::system_boundary_conditions boundaries_;
```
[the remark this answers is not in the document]
HanatoK: Done.

src/colvarcomp.cpp
giacomofiorin: > Maybe it is better to use `cvmodule->proxy`?

Thanks, this is a leftover from merging the static pointer PR

giacomofiorin:
I just rebased it onto `master`, therefore it contains triclinic PBC tests for all engines (except Tinker-HP).  Up for review now.

There is a spurious deviation in a GROMACS test, exceeding the 1e-5 tolerance at one step but staying closer to the reference after that. Not sure how to handle that.

HanatoK:
> I just rebased it onto `master`, therefore it contains triclinic PBC tests for all engines (except Tinker-HP). Up for review now.
> 
> There is a spurious deviation in a GROMACS test, exceeding the 1e-5 tolerance at one step but staying closer to the reference after that. Not sure how to handle that.

Did you use the double-precision version of gmx `gmx_d` to test?

giacomofiorin:
> Did you use the double-precision version of gmx `gmx_d` to test?

This is from the CI, i.e. I think so

giacomofiorin:
The latest push is the rebase onto master (previously, it was only rebased onto the branch of #944)

HanatoK (approved):

giacomofiorin:
After the latest rebase, there are no test failures

@jhenin If you like it, merge it.

PR #929, opened 2026-04-27
Update Colvars_traj documentation

capritchard:
If I understand correctly, **Colvars_traj** creates a new instance of **Colvar_traj** for each collective variable. Right now the documentation reads like each collective variable is an instance of Colvars_traj. I think this is a typo, unless I'm not understanding the code (in which case, an explanation would be very much appreciated)


giacomofiorin (approved):
Thanks for catching this!

PR #932, opened 2026-05-11
Fix: better test for MTS eABF

### LINKED ISSUE #931: eABF not compatible with NAMD MTS (GlobalMasterFrequency)?
Hi,

I previously ran eABF successfully with `GlobalMasterFrequency 2`, but this combination now seems to fail.

when using:
```tcl
colvar {
  name r
  extendedLagrangian on
  ...
}

abf {
  colvars r
  ...
}
```

```tcl
GlobalMasterFrequency 2
globalMasterStaleForces on
```

Colvars exits with the following error:

```text
Error: ABF cannot use timeStepFactor > 1 because colvar r does not provide total force estimates for the current timestep.
```

I looked at the source code and found this check:

```cpp
if (!colvars[i]->is_enabled(f_cv_total_force_current_step)) {
  // If any colvar does not have current-step total force, then
  // we can't do step 0 data
  provide(f_cvb_step_zero_data, false);
  // And we cannot do MTS either
  if (time_step_factor > 1) {
    return cvmodule->error("Error: ABF cannot use timeStepFactor > 1 because " +
      colvars[i]->description +
      " does not provide total force estimates for the current timestep.\n");
  }
}
```

My understanding is that eABF estimates the mean force from the extended-coordinate spring force, rather than from the instantaneous projected atomic total force used by standard ABF. Therefore, I am wondering whether this current-step total-force requirement is also intended to apply to eABF.

Is `GlobalMasterFrequency > 1` currently unsupported for eABF in NAMD/Colvars, or is this an unintended consequence of the shared ABF/eABF initialization path?

Thanks,  
Haohao

jhenin:
compare time step factor to proxy instead of 1
to enable again eABF with globalMasterFrequency

Fixes #931

### DIFF
--- a/src/colvarbias_abf.cpp
+++ b/src/colvarbias_abf.cpp
@@ -126,15 +126,18 @@ int colvarbias_abf::init(std::string const &conf)
       // If any colvar does not have current-step total force, then
       // we can't do step 0 data
       provide(f_cvb_step_zero_data, false);
-      // This ABF implementation assumes that the total force available at one time step (where ABF runs)
-      // is that from the previous timestep where ABF ran
-      // So we cannot do MTS with a larger factor than the global proxy MTS factor
-      // (That would require the colvar to update its total force at a time step where it doesn't run
-      // or to remember its value from the timestep when the total force is collected)
+      // This ABF implementation assumes that the total force available at one time step (where ABF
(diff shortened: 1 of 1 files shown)


HanatoK (approved):

giacomofiorin (approved):
The fix itself looks good to me.

Are you okay if I add the proxy MTS factor to the error message before mergin?

PR #945, opened 2026-07-21
Improve documentation of spinAngle and tilt

giacomofiorin:
- Provide full definition for both
- Add a new descriptive figure
- Add references for some of the equations that these variables are based on
- Add examples


doc/colvars-refman-main.tex
jhenin: This is not part of the current change set, but I realize this sentence is not very clear. I am not sure what it's getting at, so I cannot clarify it.

giacomofiorin: Yeah, it's a bit imprecise. I'll rephrase.

giacomofiorin: See the updated version?

jhenin (approved):

PR #946, opened 2026-07-22
fix: avoid crash in VMD on MacOS

jhenin:
due to Tk bug triggering OSX crash when using file extensions with more than one dot, or not beginning with a dot.

https://core.tcl-lang.org/tk/tktview/080a28104e25e054f8f5869d9333aa57c221b896

Bug is fixed upstream, but I am adding a platform-specific hack to work around it.

Might be fixed in Tk 9.0?

Marked draft until tested on the relevant platform.


jhenin:
Tested successfully, good for merging.

PR #947, opened 2026-07-23
Check that pull requests can be cherry-picked onto branches

giacomofiorin:
Ensure that bugfixes can be easily backported onto a list of pre-defined maintenance branches.  Currently, this list includes:
- `gromacs-2026`
- `namd-3.0`
On those branches, only PRs labeled `bugfix` are tested

Also tests other development branches:
- `gromacs-2027`
- `lammps-develop`
- `namd-3.1`
because they may become maintenance ones later on.  We probably have to agree on how to update them (fast-forwarding everything from `master` on a rolling basis, or do that only when the PRs to those packages are merged).

### DIFF
--- a/.github/workflows/check-cherry-pick.yml
+++ b/.github/workflows/check-cherry-pick.yml
@@ -1,34 +1,33 @@
 name: "Check cherry-pick compatibility"
 
 # Triggered on newly opened PRs and on subsequent pushes; checks whether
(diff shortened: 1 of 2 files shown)


giacomofiorin:
Tested that the logic works, including triggering the job if the label `bugfix` is applied after opening the PR

PR #949, opened 2026-07-29
Implementation of TABF

fhh2626:
This PR implements the use of the temperature scaling factor α as a CV (https://doi.org/10.1021/acs.accounts.5c00666), achieving REMD-like sampling with smaller temperature intervals. A new version of NAMD is required (https://gitlab.com/tcbgUIUC/namd/-/merge_requests/537).

example:
```
colvar {
  name alpha
  width 0.01
  lowerBoundary 0.97
  upperBoundary 2.03
  hardLowerBoundary on
  hardUpperBoundary on
  reflectingLowerBoundary on
  reflectingUpperBoundary on

  extendedLagrangian on
  extendedMass 1500000
  extendedLangevinDamping 1000

  tabfAlpha {}
}
```

In NAMD config:
```
tabf on
tabfAlphaStart 1.0
tabfAlphaMin   0.97
tabfAlphaMax  2.03
tabfOutFreq     100
```


giacomofiorin:
As a potential feature, looks very interesting!  I'm trying to understand the TABF method from to the paper you linked (https://doi.org/10.1021/acs.accounts.5c00666).

That paper uses WTM and eABF in combination with NAMD. However, no use of Colvars is reported in the paper.  Did you use a different implementation of those methods and are you then planning to start using the ones provided by Colvars with this PR?

HanatoK:
> As a potential feature, looks very interesting! I'm trying to understand the TABF method from to the paper you linked (https://doi.org/10.1021/acs.accounts.5c00666).
> 
> That paper uses WTM and eABF in combination with NAMD. However, no use of Colvars is reported in the paper. Did you use a different implementation of those methods and are you then planning to start using the ones provided by Colvars with this PR?

I think they mentioned Colvars in the SI but not cited. The original WTM-TABF was implemented in OpenMM.

namd/src/colvarproxy_namd.C
```diff
@@ -1702,3 +1721,65 @@ int colvarproxy_namd::get_dE_dlambda(cvm::real* dE_dlambda) {
...
 }
+
+
+/// Request TABF energy computation every freq steps
+int colvarproxy_namd::request_tabf_energy_freq(int const freq) {
+  if (freq != 1) {
+    cvmodule->error("TABF alpha dynamics currently requires timeStepFactor 1.\n");
+    return COLVARS_INPUT_ERROR;
+  }
+  if (!simparams->tabfOn) {
```
HanatoK, line 1733: @fhh2626 Colvars is required to be compatible with multiple recent versions of NAMD, so if you use `simparams->tabfOn` then the code would not be compiled with a NAMD without TABF (like NAMD 3.0.x), so I suggest that you could try adding a C macro like `#define NAMD_SUPPORT_TABF` in `src/common.h` of NAMD, and then guarding the code like:
```
#if defined(NAMD_SUPPORT_TABF)
if (!simparams->tabfOn) {
...
}
#endif
```

HanatoK (changes requested):

fhh2626:
> As a potential feature, looks very interesting! I'm trying to understand the TABF method from to the paper you linked (https://doi.org/10.1021/acs.accounts.5c00666).
> 
> That paper uses WTM and eABF in combination with NAMD. However, no use of Colvars is reported in the paper. Did you use a different implementation of those methods and are you then planning to start using the ones provided by Colvars with this PR?

Dammit, @giacomofiorin, that was careless of us. We normally cite the Colvars paper in the Methods section, but this paper does not have a Methods section, and we failed to include the citation elsewhere. We sincerely apologize for the confusion. 
We used OpenMM for the WTM-TABF calculations, rather than the WTM and eABF implementations provided by Colvars. However, Colvars was indeed used for some of the other calculations, and we should have stated this explicitly. We will definitely describe our use of Colvars and the relevant implementation details much more carefully in our follow-up papers.

To answer your question directly, this PR is follow-up work to that paper and implements WTM-TABF in NAMD. The WTM-TABF implementation used in the original paper was developed in OpenMM.

namd/src/colvarproxy_namd.C
fhh2626: done

HanatoK:
@fhh2626 I suggest to enable the backend test workflows in your CI (https://github.com/fhh2626/colvars/actions) as follows:
tabf_ci
After that, we could check if your code could be built with various backends.

@giacomofiorin Is it possible to allow @fhh2626 to pull the NAMD and VMD code from https://github.com/Colvars/namd and https://github.com/Colvars/vmd for CI?

giacomofiorin:
> @giacomofiorin Is it possible to allow @fhh2626 to pull the NAMD and VMD code from https://github.com/Colvars/namd and https://github.com/Colvars/vmd for CI?

I think it would be easier to submit PRs from a branch of this repo.

PR #958, opened 2026-09-14
Fix the NAMD GM interface and update the CUDAGM build files

HanatoK:
`get_smp_mode` and `set_smp_mode` should have been correctly implemented in `colvarproxy.cpp` where the mode to be set should be from `get_available_smp_modes`.

This PR also updates the CUDAGM build dependency files.


giacomofiorin (approved):
This PR correctly resolves an inconsistency between `colvarproxy` and `colvarproxy_namd`.

giacomofiorin:
@HanatoK I added the `bugfix` label, which triggered the failed checks for cherry-pick ability.

I don't believe that you could trigger a bug in NAMD 3.0: otherwise, an adapted version of this commit should be pushed to the `namd-3.0` branch

HanatoK:
> @HanatoK I added the `bugfix` label, which triggered the failed checks for cherry-pick ability.
> 
> I don't believe that you could trigger a bug in NAMD 3.0: otherwise, an adapted version of this commit should be pushed to the `namd-3.0` branch

I don't think this should be pushed to `namd-3.0`, because this depends on 872521aa07cc653c3a5c0cd587a9aa93e507af33, which is not in `namd-3.0`.

